PR #1, opened 2020-04-02
Add common boilerplate

charlieduong94:
Just copied most of the stuff from the `jupiterone-client-nodejs` project. Willing to address opinions on where config should live in later prs, but just trying to get moving for now.

@aiwilliams will be getting travis setup in a separate pr.


package.json
```diff
@@ -0,0 +1,86 @@
+{
+  "name": "@jupiterone/integration-sdk",
+  "version": "0.0.1",
+  "description": "The SDK for developing JupiterOne integrations",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "repository": "git@github.com:JupiterOne/jupiterone-client-nodejs.git",
```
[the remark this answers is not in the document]
charlieduong94: F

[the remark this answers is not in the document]
charlieduong94: Fixed

.gitignore
```diff
@@ -0,0 +1,11 @@
+dist/
+node_modules/
+coverage/
+.vscode/
+.env
+
+.DS_Store
```
fomentia, line 7: Minor, but I really think we shouldn't include OS- or editor-specific entries in projects' gitignores. Those should be in each user's global gitignore.

.prettierignore
```diff
@@ -0,0 +1 @@
+dist
```
fomentia, line 1: Is `node_modules` automagically ignored by Prettier? That'd be sikk ngl

package.json
```diff
@@ -0,0 +1,86 @@
+{
+  "name": "@jupiterone/integration-sdk",
+  "version": "0.0.1",
```
fomentia, line 3: Another minor thing, following semver this should be `0.1.0` (I think, pls correct if wrong).

package.json
```diff
@@ -0,0 +1,86 @@
...
+    "singleQuote": true
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.{js,jsx,ts,tsx,md,html,css}": "prettier --write"
+  },
```
fomentia, line 81: I think you said in a previous PR that you would move these out at some point but I just want to have something to comment on in this file.

.prettierignore
charlieduong94: Yep, automagically ignored

package.json
charlieduong94: I think you are right, since there's nothing to release yet, mind if the change to `0.1.0` happens when there's actually an initial publish?

.gitignore
charlieduong94: ah sorry, copypasta from j1 node client

charlieduong94: will remove later so that dad can do travis things

package.json
fomentia: Sounds gucc

PR #2, opened 2020-04-03
Add documentation about the expected SDK and CLI interface

charlieduong94:


docs/development.md
```diff
@@ -0,0 +1,775 @@
...
+capable of interfacing with a given provider. Next, the `getStepStartStates`
+will run to get a list of integration phases and steps that should be executed.
+Finally, the phases in `integrationStepPhases` are then executed in sequence.
+
+### What's in the `IntegrationExecutionContext` and `IntegrationExecutionStepContext`?
+
+The `IntegrationExecutionContext` contains some basic utilities to provide some
+information about the entity data.
+
+It includes a `logger` that can be used to display additional information
```
aiwilliams, line 214: I think the `logger` would be distinct from a mechanism that allows presenting information to the user in a job recording, assuming the current job events presentation we have today. The `logger` would allow for traditional backend info/debug logging.

How about another thing, `jobLog`, with functions that constrain the class of messages? For example, when a list of resources work (like `listResources`), but details are forbidden with a `400` or `403` response (`describeResource`), we could say `jobLog.forbiddenResource({ err, _type, resource: "someResource" })` and the execution environment would `logger` this and also construct a consistent message for the UI (or collect them and provide a summary).

docs/development.md
```diff
@@ -0,0 +1,775 @@
...
+types will come later.
+
+### Auto-magic behind the framework
+
+#### Logging
+
+When running an integration via the CLI tool will automatically log out
+information about the state changes that the integration. This allows for
+developers to keep track of what phase and step an integration is performing
+without the need to explicily add logging information themselves.
```
aiwilliams, line 252: I think of this description as applying to the `jobLog` I described above.

docs/development.md
```diff
@@ -0,0 +1,775 @@
...
+The `executionContext` that is provided in the `executionHandler` step exposes a
+`jobState` utility that can be used to collect entity and relationship data via
+`addEntities` and `addRelationship` functions. The `jobState` utility will
+automatically flush the data to disk as a certain threshold of entities and
+relationships is met. The data flushed to disk are grouped in folders that based
+on the phase and step that was run. Flushed data will be replicated to a
+separate directory that groups data by an entity or relationship's `_type`.
+Additional utilities for reading in entities or relationships via the `_type`
+are provided on the `jobState` object (the `listEntitiesByType` and
+`listRelationshipsByType` functions). From our experience, integrations most
```
aiwilliams, line 265: I love that we start with just `_type` searching. I think that we should consider making these functions `iterateEntities({ _type: "some_type" }, (e) => { ... })`, with that input being `type IterateEntitiesFilter = { _type: string }` at first, and later ` type IterateEntitiesFilter = { _type?: string, _class?: string | string[], [property: string]: value: undefined | string | RegExp | number | whateverapropcanbeingraph }`. This will make the interface consistent no matter what properties we search by in the future. Same for `iterateRelationships()` of course.

docs/development.md
```diff
@@ -0,0 +1,775 @@
...
+generate and an `executionHandler` function that performs the data collection
+work. It is important to provide `types` because that data will be used to
+provide context for the backend synchronization process and determine how
+updates and deletes should be applied.
+
+Optionally, a `step` can contain a `dependsOn` list that references previous
+phases and steps. This field will be used to determine if previous work has
+failed to complete and if the synchronization process should treat the data
+retrieved in the step as a partial dataset. See the [Failure handling](#Failure
+handling) section below for more information.
```
fomentia, line 196: If I have five independent resources, why wouldn't I configure them all to run as steps? Is there ever a time that I'd want to process a resource in a new phase if it _didn't_ somehow depend on the resources processed in the preceding phases? In other words, is a `dependsOn` relationship with the previous phases implied simply by processing my resource in a new phase?

docs/development.md
```diff
@@ -0,0 +1,775 @@
...
+
+#### Data collection
+
+The `executionContext` that is provided in the `executionHandler` step exposes a
+`jobState` utility that can be used to collect entity and relationship data via
+`addEntities` and `addRelationship` functions. The `jobState` utility will
+automatically flush the data to disk as a certain threshold of entities and
+relationships is met. The data flushed to disk are grouped in folders that based
+on the phase and step that was run. Flushed data will be replicated to a
+separate directory that groups data by an entity or relationship's `_type`.
```
fomentia, line 262: Will it actually be replicated, as in copied? If we're using the filesystem, can we just use links?

docs/development.md
```diff
@@ -0,0 +1,775 @@
...
+    "id": "build-relationships-phase",
+    "name": "Collect accounts",
+    "status": "partial_success",
+    "steps": [
+      {
+        "id": "build-user-to-group-relationships",
+        "name": "Fetch Accounts",
+        "types": ["my_integration_user_to_group_relationship"],
+        "dependsOn": ["fetch-users", "fetch-groups"],
+        "status": "partial_success_from_dependency_failure"
```
fomentia, line 395: It would be nice to have a list of possible phase and step exit statuses in this section.

docs/development.md
```diff
@@ -0,0 +1,775 @@
...
+that would prefer to use a different language for building integrations, the
+`sync` command does not require that a developer use the JupiterOne integration
+framework at all. The `sync` command will just recursively walk the
+`.j1-integration/graph` directory, search for `.json` files, validate the data
+to ensure it is formatted correctly, and publish it up via the synchronization
+API.
+
+The JupiterOne data model is open source and can be used by anyone to ensure
+that data conforms to our expectations. If you build custom tooling or your own
+framework for developing integrations, let us know!
```
fomentia, line 619: Ah that's hot. That's real hot. :fire:

docs/development.md
```diff
@@ -0,0 +1,775 @@
...
+After the collection phase, the integration summary and partial datasets
+metadata will be written to disk in the `.j1-integration/summary.json` file.
+
+Here is an example of what the summary file would look like.
+
+```
+{
+  "integrationStepPhasesResult": [
+    {
+      id: "account-phase",
```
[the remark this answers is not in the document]
charlieduong94: Sure, will change

docs/development.md
```diff
@@ -0,0 +1,775 @@
...
+      name: "Collect accounts",
+      status: "success"
+    },
+    {
+      id: "users-and-groups-phase",
+      name: "Collect users and groups",
+      status: "partial_failure",
+      steps: [
+        {
+          id: "fetch-users",
```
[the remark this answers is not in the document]
charlieduong94: :+1:

docs/development.md
```diff
@@ -0,0 +1,775 @@
...
+```
+{
+  "integrationStepPhasesResult": [
+    {
+      id: "account-phase",
+      name: "Collect accounts",
+      status: "success"
+    },
+    {
+      id: "users-and-groups-phase",
```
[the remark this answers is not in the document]
charlieduong94: will also change

docs/development.md
```diff
@@ -0,0 +1,775 @@
...
+
+If you are working with an existing integration instance and would prefer to
+leverage the configuration field values from that be used, you can optionally
+supply an instance id. The CLI will leverage the values stored on the
+integration instance instead of locally defined environment variables.
+
+By default, when an `--instance` is specified, a developer will be prompted to
+input some credentials.
+
+ex: `j1-integration collect --instance <integration instance id>`
```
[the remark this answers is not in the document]
charlieduong94: Yeah

docs/development.md
charlieduong94: Seems reasonable. Will change

docs/development.md
charlieduong94: will add

docs/development.md
aiwilliams: 🤔 Great question. I'm not sure a dependency is always implied. There have been times when I wanted to group steps so they could fetch their data concurrently, but they had no relationship to each other. A later phase would fetch other unrelated things where some of those steps could have dependencies on one of the previous phase steps, but it would not be possible for the computer to know which steps were dependent.

docs/development.md
charlieduong94: My OG thought was that each step would produce a big dump of entities and relationships at the end of each step, but since there is already work being done to split up data via `_type`, there's no real reason why we couldn't just produce some smaller files and create links. Will update.

docs/development.md
fomentia: @aiwilliams In your example, you have phase 1 and 2, where some of the steps in phase 2 are dependent on the results of steps in phase 1 but some are not. Is there a reason you didn't move the independent steps in phase 2 into phase 1?

But I do hear what you're saying about having steps in a later phase depend on only some of the steps in an earlier phase such that implying a dependency on all steps in that earlier phase would be unnecessary.

docs/development.md
charlieduong94: As I was writing this, I was p much omitting the actual event logging and my intent was to have the `logger` just be for debugging since we've talked about the issue with people not really knowing what kind of events to log (my writing sucks so that wasn't clear). 

I like the idea of a `jobLog` with a limited interface though. Providing that would allow for devs to provide some additional information about what could have potentially wrong while a step is being executed. Thoughts on perhaps exposing the `logger` under a different name to clarify that information logged with it will not be displayed in the event log? Perhaps `debugLogger` or `internalLogger`? (@philidem help me out here) 

`forbiddenResource` seems like a good function to start with exposing. Any other functions that might be worth adding?

docs/development.md
charlieduong94: no u

docs/development.md
charlieduong94: Yeah, if you have five independent resources that do not rely on each other, you would probably just want to configure them as just steps. If there was a resource that did not rely on previously collected resources, you would never really want to have it be processed in a later step since wouldn't make much sense waiting to collect that data.

And you make a good point here, the `dependsOn` relationship is seemingly implied by the previous phases considering the way they define an order of operations for the integration run. However (with the current setup), I don't think that the implied relationship there should be used to determine partial datasets. Mainly because the structure of phases and steps is too limited to be able to create an implicit dependency graph that is capable of determining partial datasets without causing later phases + steps to be marked as partial. For example, a resource collected in one phase could fail to be retrieved and cause data in a later phase that could be unrelated to be marked as partial.

I do think it's important to be able to specify the prerequisites that need to be fulfilled to determine partial datasets. I'm curious if we should potentially consider going with a different structure to support building integrations with a structure that allows for integrations to be built with implicit dependencies. Another potential option there would be to maybe do away with phases entirely and consider going with a more terraform like approach with explicit dependencies and have our state machine construct a graph of work to perform and manage the execution based on that.

fomentia: @charlieduong94 Makes sense; also, sounds good. I'm in favor of copying Hashicorp wherever possible. :laughing:

docs/development.md
aiwilliams: I think the only thing I've ever wanted to expose to the job event log from an integration was permissions problems. The `invocationValidator` function just throws certain types of errors which the execution environment catches and sends to the job event log; this has allowed for communicating authentication problems. Authorization problems tend to show up in specific resources, and that cannot be an exception path because it would kill the step. So, I think the only real need is to send authorization/resource access failures.

Perhaps the `logger` could have another function, since it is a type we control, such as `logger.auth()`, and we not only log that to `console` using Bunyan, but we also send a job event? Then we just have the `logger` alone, (`IntegrationLogger`) and document functions to indicate what will be presented to a user and what will not?

We'd like to send `error()` and `auth()` I think to the job event log.

docs/development.md
aiwilliams: > Another potential option there would be to maybe do away with phases entirely and consider going with a more terraform like approach with explicit dependencies and have our state machine construct a graph of work to perform and manage the execution based on that.

Yah, I like this idea very much. It makes a lot of sense to be explicit in dependencies, and would get rid of the phase/step thing and just allow developers to write data fetching work code, with dependencies declared, and then also relationship building work that depends on all fetching work to complete.

charlieduong94:
I've updated the PR but due to positive feedback about removing phases and going only with steps, I'm going to go ahead and work out a plan for that. I'll be updating the PR to remove references to phases.

docs/development.md
charlieduong94: I think I dig it. At first I was thinking it was a bit magical but the more I think about it, the more I like having a unified interface.

docs/development.md
```diff
@@ -0,0 +1,829 @@
...
+It's best to keep a "collect and forget" mindset to avoid retaining too much
+collected data in memory. Flushed data can always be pulled back into memory
+later via the `listEntitiesByType` and `listRelationshipsByType` and reading
+data from disk is decently fast.
+
+#### Failure handling
+
+By default, the framework will only halt the execution of an integration if the
+configuration validation fails. Failures that occur during the execution of a
+step will not halt the execution of later steps that depend on it.
```
aiwilliams, line 410: I was thinking that if step B depends on step A, and step A raises an error/fails, we would not run step B because it depends on step A. If we do execute step B anyway, it will have to find out if something went wrong in step A to avoid referencing data that may not exist, right?

charlieduong94: I was thinking that a failure in a previous step does not necessarily mean that there is an absence of data to work on. For example, if I am able to fetch users but only half of the groups, I should be able to still build relationships between the users and groups where possible. And since the relationship step relies on a step that failed, it's status would be marked as `partial_success_due_to_dependency_failure` and the types that it has produced would be included in the `partialDatasets` sent to the synchronizer.

charlieduong94:
Ahh yes, I thought I had that in here with phases (when that was a thing) and steps but I guess I never committed it. Will add.

charlieduong94:
Awesome. Thanks a ton @aiwilliams and @fomentia for the awesome feedback.

PR #4, opened 2020-04-07
Implement cache directory utilities needed for jobState

charlieduong94:
This adds a `cacheDirectory` utility that will be shared by both the `jobState` and CLI. The goal of this util is to provide an interface for writing data relative to some specified directory (the cache directory). The two functions implemented are for writing data and will be used by the `jobState` as data is flushed. Additional functions to help with iterating through files will come as the `iterateEntities` and `iterateRelationship` functions are implemented.

Normally I would use library like `fs-extra` to assist with writing files (does `mkdir -p` for you) but now that node's built-in `fs` lib has much better promise support and a recursive option with `mkdir`, I decided `fs-extra` added little value. 

An additional benefit of sticking with regular `fs` is that we can use tools like [memfs](https://github.com/streamich/memfs) (kind of a successor to `memory-fs`) to execute tests against an in-memory file system which is fast, does quick clean up, and (imo) has a nice interface for setting up test scenarios.

### DIFF
--- a/src/__tests__/cacheDirectory.test.ts
+++ b/src/__tests__/cacheDirectory.test.ts
@@ -1,0 +1,176 @@
+import { promises as fs } from 'fs';
+import { v4 as uuid } from 'uuid';
+import { vol } from 'memfs';
+
+import {
+  getDefaultCacheDirectory,
+  writeJsonToPath,
+  symlink,
+} from '../cacheDirectory';
+
+jest.mock('fs'); // applies manual mock which uses memfs
+
+afterEach(() => {
+  jest.clearAllMocks();
+  vol.reset(); // clear out memory
+});
+
+describe('getDefaultCacheDirectory', () => {
+  test('should utilize the current directory for building the default cache directory', () => {
+    const mockCwdResult = `/${uuid()}`;
+    jest.spyOn(process, 'cwd').mockReturnValue(mockCwdResult);
+
+    expect(getDefaultCacheDirectory()).toEqual(
+      `${mockCwdResult}/.j1-integration`,
+    );
+  });
+});
+
+describe('writeJsonToPath', () => {
+  test('should pretty write json to the specified file', async () => {
+    const json = { test: '123' };
+
+    const directory = '/';
+    const filename = `${uuid()}.json`;
+
+    await writeJsonToPath({
+      cacheDirectory: directory,
+      path: filename,
+      data: json,
+    });
+
+    const writtenData = await fs.readFile(`${directory}/${filename}`, 'utf8');
+    expect(writtenData).toEqual(JSON.stringify(json, null, 2));
+  });
+
+  test('should recursively create directories prior to writing', async () => {
+    const json = { woah: 'json' };
+
+    const directory = '/';
+    const filename = `test/dir/that/does/not/already/exist/${uuid()}.json`;
+
+    const mkdirSpy = jest.spyOn(fs, 'mkdir');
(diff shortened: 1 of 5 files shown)


src/__tests__/cacheDirectory.test.ts
```diff
@@ -0,0 +1,176 @@
...
+import {
+  getDefaultCacheDirectory,
+  writeJsonToPath,
+  symlink,
+} from '../cacheDirectory';
+
+jest.mock('fs'); // applies manual mock which uses memfs
+
+afterEach(() => {
+  jest.clearAllMocks();
```
aiwilliams, line 14: How do you feel about using the option in a jest config: https://jestjs.io/docs/en/configuration#clearmocks-boolean I prefer it because I always want mocks cleared, and it saves from having to add this line in test files.

src/__tests__/cacheDirectory.test.ts
```diff
@@ -0,0 +1,176 @@
...
+      },
+    );
+
+    // not pretty but wanted to enforce that mkdir was being called before
+    // write file
+    //
+    // ref: https://github.com/facebook/jest/issues/4402#issuecomment-534516219
+    const mkdirOrder = mkdirSpy.mock.invocationCallOrder[0];
+    const writeFileOrder = writeFileSpy.mock.invocationCallOrder[0];
+    expect(mkdirOrder).toBeLessThan(writeFileOrder);
```
aiwilliams, line 78: Does `memfs` not raise an error if an attempt is made to write to a file path that references a directory(ies) that does not exist? Or to say it another way, a normal file system would fail to write the file if the directory doesn't exist, so when the file is written/read in the test, you can know that `mkdir` must have been called.

src/cacheDirectory.ts
```diff
@@ -0,0 +1,54 @@
...
+
+interface WriteDataToPathInput {
+  cacheDirectory?: string;
+  path: string;
+  data: object;
+}
+
+/**
+ * Helper function for writing arbirary data to a path
+ * relative to the cache directory
```
aiwilliams, line 18:
```suggestion
 * Writes data to a path relative to the cache directory.
```

FWIW, I'm not a big fan of calling out the "helper" nature of a function because, the way I see it, all functions are helpers to other functions 😄

aiwilliams:
> Normally I would use library like `fs-extra` to assist with writing files (does `mkdir -p` for you) but now that node's built-in `fs` lib has much better promise support and a recursive option with `mkdir`, I decided `fs-extra` added little value.
> 
> An additional benefit of sticking with regular `fs` is that we can use tools like [memfs](https://github.com/streamich/memfs) (kind of a successor to `memory-fs`) to execute tests against an in-memory file system which is fast, does quick clean up, and (imo) has a nice interface for setting up test scenarios.

I think this is a great idea.

src/__tests__/cacheDirectory.test.ts
charlieduong94: It does fail if the directory doesn't exist like a regular file system. I suppose the assertion here is a bit overkill.

src/__tests__/cacheDirectory.test.ts
charlieduong94: I'm cool with changing that. Will make the tweak.

src/cacheDirectory.ts
charlieduong94: Will update.

PR #5, opened 2020-04-07
Implement entity and relationship flushing portions of jobState

charlieduong94:
Going to have this PR point to the [cacheDirectory-util](cacheDirectory-util) branch for now until that is merged. This is a slightly bigger PR than the previous one, but it's mostly tests.

This PR introduces the `JobState` interface as well as very stripped down `Entity` and `Relationship` types (integrations don't need to care about much anymore). Raw data isn't on the entities yet because I'd like to add that with the `createIntegration*` converters. I'm baking in these types here for now so I can keep moving along, but these types may end up coming from the data model repo later.

I've implemented the `addEntities`, `addRelationships`, and `flush` functions required by the `JobState` interface. `iterateEntities` and `iterateRelationships` are currently stubbed out and will be implemented in the next PR.

Note: this does not handle duplicate `_key` tracking. Going to tackle that later.


src/framework/jobState/LocalStepJobState/LocalStepJobState.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  async addRelationships(newRelationships: Relationship[]) {
+    this.relationships = this.relationships.concat(newRelationships);
+
+    if (this.relationships.length >= GRAPH_OBJECT_BUFFER_THRESHOLD) {
+      await this.flushRelationshipsToDisk();
+    }
+  }
```
aiwilliams, line 44: Any reason to care about using the sum, such as `entities + relationships >= GRAPH_OBJECT_BUFFER_THRESHOLD`? Without it, the memory could grow to twice as much as that number constant indicates, I think.

src/framework/jobState/LocalStepJobState/LocalStepJobState.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    if (this.entities.length) {
+      work.push(this.flushEntitiesToDisk());
+    }
+
+    if (this.relationships.length) {
+      work.push(this.flushRelationshipsToDisk());
+    }
+
+    await Promise.all(work);
```
aiwilliams, line 73: I read the docs at https://github.com/sindresorhus/type-fest/blob/HEAD/source/promisable.d.ts, but I don't think I fully understand the value in this scenario. Would you mind helping me? 😄

I think it would be good to have `flushEntitiesToDisk()` worry about whether there is anything to flush or not, or else that function cannot be called without the caller verifying that it can be called.

```suggestion
    await Promise.all([
      this.flushEntitiesToDisk(),
      this.flushRelationshipsToDisk(),
    ]);
```

package.json
```diff
@@ -18,6 +18,14 @@
     "test": "jest --passWithNoTests",
     "prepublishOnly": "tsc --declaration"
   },
+  "dependencies": {
+    "@types/lodash": "^4.14.149",
+    "eslint-config-prettier": "^6.10.1",
+    "eslint-plugin-prettier": "^3.1.2",
```
aiwilliams, line 24: Are these meant to be `devDependencies`?

src/framework/jobState/LocalStepJobState/flushDataToDisk.ts
```diff
@@ -0,0 +1,91 @@
...
+  cacheDirectory?: string;
+  step: string;
+  collectionType: CollectionType;
+  data: Entity[] | Relationship[];
+}
+
+/**
+ * Utility for flushing collected data to disk and
+ * creating placing a symlink into an 'index' directory
+ * based on type.
```
aiwilliams, line 21:
```suggestion
 * Utility for flushing collected data to disk and
 * creating a symlink into an 'index' directory
 * based on type.
```

src/framework/jobState/LocalStepJobState/LocalStepJobState.ts
```diff
@@ -0,0 +1,97 @@
...
+import { flushDataToDisk } from './flushDataToDisk';
+
+export const GRAPH_OBJECT_BUFFER_THRESHOLD = 500; // arbitrarily selected, subject to tuning
+
+interface LocalStepJobStateInput {
+  step: string;
+  cacheDirectory?: string;
+}
+
+export class LocalStepJobState implements JobState {
```
aiwilliams, line 15: I wondered about this name and forgot in the review... What is the reason for including the word "Step"? I don't see how this is limited to step state. Also, what do you think of the name `FileSystemJobState`?

src/framework/jobState/LocalStepJobState/LocalStepJobState.ts
```diff
@@ -0,0 +1,97 @@
...
+import { Entity, Relationship } from '../../types';
+import { JobState, GraphObjectFilter, GraphObjectIteratee } from '../types';
+
+import { flushDataToDisk } from './flushDataToDisk';
+
+export const GRAPH_OBJECT_BUFFER_THRESHOLD = 500; // arbitrarily selected, subject to tuning
+
+interface LocalStepJobStateInput {
+  step: string;
+  cacheDirectory?: string;
```
aiwilliams, line 12: As I think of this as a `FileSystemJobState`, I now wonder if the name here should just be `rootDirectory`. It isn't clear to me (from the `FileSystemJobState` perspective I suppose), what would be "cached", whereas `rootDirectory` clearly communicates the location in which all content will be maintained and it isn't really a cache. Or is it?

package.json
charlieduong94: oof yes, fixing now

src/framework/jobState/LocalStepJobState/LocalStepJobState.ts
charlieduong94: I was thinking that the worse that could happen would be having 1000 entities or relationships in memory, which doesn't seem like too much to me.

src/framework/jobState/LocalStepJobState/LocalStepJobState.ts
charlieduong94: tbh, idk why I didn't just use `Promise` instead of `Promisable`. Removing.

charlieduong94: And I'm cool with moving the check into `flushEntitiesToDisk`

src/framework/jobState/LocalStepJobState/LocalStepJobState.ts
charlieduong94: I suppose `cache` is probably the wrong terminology since the data isn't really hidden (although by default data is  `.j1-integration` directory, which suggests that it's a hidden directory. I can change this for the whole project in a later PR since the terminology is use all over the place.

src/framework/jobState/LocalStepJobState/LocalStepJobState.ts
aiwilliams: Agreed, it's nothing really. I think what you have is just fine. I should have written, it was surprising to me when I read through the code that there could be more than `GRAPH_OBJECT_BUFFER_THRESHOLD` objects not flushed.

src/framework/jobState/LocalStepJobState/LocalStepJobState.ts
charlieduong94: That's a much better name than `LocalStepJobState`. @philidem would be proud.

I included `step` in the name because the intent was to create a job state object with reference to the step per execution so that the job state knows where to place data. The more I think about it though, the more it seems like having an individual job state per step is a bad idea considering lots of steps could run in parallel. With how things are implemented here, each state object has it's own set buffer of entities to manage.

I think I'm going to tweak the implementation a bit to allow for `child` job states to be created (kinda like loggers) with context about steps and where to write data, but a shared pool of entities and relationships.

charlieduong94:
> Was the documentation of the types here pulled from jupiter-types?

Yeah, I picked the fields from `jupiter-types`

src/framework/jobState/LocalStepJobState/LocalStepJobState.ts
charlieduong94: Actually... might be better to handle this in a later PR. I have a solution in mind that probs won't take long to implement, but time might be better spent getting the integration sdk in a usable state before Monday. With how small these upcoming integrations will be, I don't think memory will be an issue for a while.

aiwilliams: Waiting sounds good to me! The most important thing in my mind right now is making it work, clear interfaces/naming, and where to place responsibilities. Performance improvements should definitely come later.

> With how things are implemented here, each state object has it's own set buffer of entities to manage.

I'll admit that I didn't pick up on the idea that each step would have it's own state obj. It *does* make sense that each step has state, but I've been looking at this code so far as the collector of entities and relationships, which probably should cross steps.

This suggests to me that there is likely to be 1) step state and 2) entities/relationships collected across steps. I think it may lead to some confusion to have those concepts mixed into a single thing. Perhaps there is a `FileSystemGraphObjectCollector`, and a separate `StepState` concept having a containment hierarchy reflecting the step dependency graph. 🤔

We'll see! Hopefully those things will be transparent to the integration author and we can iterate the internals. 💪

aiwilliams: Or `FileSystemGraphObjectStore` 😄

src/framework/jobState/LocalStepJobState/LocalStepJobState.ts
aiwilliams: Ahh, that's helpful for understanding where the name came from, thank you! FWIW, I don't think of being hidden an intrinsic property of a cache. I tend to call something a cache if it's meant to keep expensive work from happening, and a buffer if it's meant to delay future work.

src/framework/jobState/FileSystemJobState/FileSystemJobState.ts
```diff
@@ -0,0 +1,92 @@
+import { Entity, Relationship } from '../../types';
+import { JobState, GraphObjectFilter, GraphObjectIteratee } from '../types';
+
+import { flushDataToDisk } from './flushDataToDisk';
+
+export const GRAPH_OBJECT_BUFFER_THRESHOLD = 500; // arbitrarily selected, subject to tuning
+
+interface LocalStepJobStateInput {
```
aiwilliams, line 8:
```suggestion
interface FileSysemJobStateInput {
```

charlieduong94: ah big oops. Changing rnrq

src/framework/jobState/FileSystemJobState/flushDataToDisk.ts
```diff
@@ -0,0 +1,91 @@
...
+ * Utility for flushing collected data to disk and
+ * creating a symlink in the 'index' directory
+ * based on the entity or relationship '_type'.
+ */
+export async function flushDataToDisk({
+  cacheDirectory,
+  step,
+  collectionType,
+  data,
+}: FlushDataToDiskInput) {
```
fomentia, line 28: If the type for the object parameter is so closely scoped that its name is `nameOfFunc + "Input"`, why don't you simply include it in the function declaration?

fomentia:
@aiwilliams The open-source data model project currently exports no types, only the schemas and helpers for validation. Were you thinking we'd move all entity-related types from `jupiter-types` into the data model?

aiwilliams:
> Were you thinking we'd move all entity-related types from `jupiter-types` into the data model?

No, I think we should be selective, limiting what we have in the data-model to those types which communicate what a graph object is - a `_key`, `_type`, `_class` as required for entities and relationships, `displayName`, `webLink` as optional for entities and relationships, and `_fromEntityKey`, `_toEntityKey` as required for relationships, and any number of other properties on entities and relationships, named as much as possible according to suggestions in the class schemas. There would be no other types. I would not find much value in having every single data-model schema type defined because I'd like to avoid pretending those types matter and keep folks from hand-coding against them. That is, we should not need/want to maintain types for `Host`, `Record`, etc. but encourage integration authors to write tests for their interesting conversions and use the generators which take in the source data and `_class: "Host"` and verify the auto-generated entity meets minimum requirements of the schema.

Note that https://github.com/JupiterOne/integration-sdk/blob/master/src/framework/types/persistedObject.ts does not include anything that is persister-only maintenance/tracking information, and I think this is most excellent for the data-model. We want always to present the most narrow interface we can, when we're dealing with a language that uses types to communicate the design of the program and the expected use of the objects.

I think we should rename `PersistedObject` to `GraphObject`. This will clarify the fact that persistence is a concern orthogonal to the data-model. Of course, the persister may have requirements and features that extend this data-model - such as `_mapping` properties that may be included for building mapped relationships, but this is not really a concern for the data-model.

PR #6, opened 2020-04-08
Implement jobstate iterate functions

charlieduong94:
This PR implements the `iterateEntities` and `iterateRelationships` functions. I added a function for recursively walking a directory to read files and used that for implementing iterate functions. It's a bit overkill right now, but was quick to implement and will be needed when we start working on the synchronization command anyway.

### DIFF
--- a/src/framework/jobState/FileSystemJobState/FileSystemJobState.ts
+++ b/src/framework/jobState/FileSystemJobState/FileSystemJobState.ts
@@ -50,12 +50,7 @@ export class FileSystemJobState implements JobState {
     filter: GraphObjectFilter,
     iteratee: GraphObjectIteratee<Entity>,
   ) {
-    const entitiesMatchingType = this.entities.filter(
-      (e) => e._type === filter._type,
-    );
-    for (const entity of entitiesMatchingType) {
-      await iteratee(entity);
-    }
+    await this.flushEntitiesToDisk();
 
     await iterateEntityTypeIndex({
       cacheDirectory: this.cacheDirectory,
@@ -68,12 +63,7 @@ export class FileSystemJobState implements JobState {
     filter: GraphObjectFilter,
     iteratee: GraphObjectIteratee<Relationship>,
   ) {
-    const relationshipsMatchingType = this.relationships.filter(
-      (r) => r._type === filter._type,
-    );
-    for (const relationship of relationshipsMatchingType) {
-      await iteratee(relationship);
-    }
+    await this.flushRelationshipsToDisk();
 
     await iterateRelationshipTypeIndex({
       cacheDirectory: this.cacheDirectory,
--- a/src/framework/jobState/FileSystemJobState/FileSystemJobState.ts
+++ b/src/framework/jobState/FileSystemJobState/FileSystemJobState.ts
@@ -3,6 +3,11 @@ import { JobState, GraphObjectFilter, GraphObjectIteratee } from '../types';
 
 import { flushDataToDisk } from './flushDataToDisk';
 
+import {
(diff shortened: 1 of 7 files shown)


src/framework/jobState/FileSystemJobState/FileSystemJobState.ts
```diff
@@ -45,16 +50,36 @@ export class FileSystemJobState implements JobState {
...
+    );
+    for (const entity of entitiesMatchingType) {
+      await iteratee(entity);
+    }
+
+    await iterateEntityTypeIndex({
+      cacheDirectory: this.cacheDirectory,
+      type: filter._type,
+      iteratee,
+    });
```
aiwilliams, line 64: Is this meant to iterate things that have not been flushed and things that have been flushed? What do you think about ensuring that all things are flushed, then just walk what is on disk?

src/framework/jobState/FileSystemJobState/indices.ts
```diff
@@ -0,0 +1,89 @@
+import { Entity, Relationship } from '../../types';
+import { GraphObjectIteratee } from '../types';
+
+import {
+  walkDirectory,
+  WalkDirectoryIterateeInput,
+} from '../../../cacheDirectory';
```
aiwilliams, line 7: Is this in `src/cacheDirectory` because you anticipate using it in code outside of `jobState/` or `FileSystemJobState/`?

src/framework/jobState/FileSystemJobState/FileSystemJobState.ts
charlieduong94: yeah, this would make the code a little more simple. Changing

src/framework/jobState/FileSystemJobState/indices.ts
charlieduong94: Yeah, hope to use it when performing `synchronization` (most of that code will likely live under it's directory).

charlieduong94:
> Note that I approve even with my suggestions/questions because I would like you to have the freedom to keep going and address changes, should we agree some are good, in future PRs.

Yeah I know. Depending on the feedback, I sometimes prefer to address it immediately otherwise I'll forget.

src/framework/jobState/FileSystemJobState/indices.ts
```diff
@@ -0,0 +1,89 @@
...
+    iteratee: async (input) => {
+      const object = parseData(input);
+      if (isObjectFlushedRelationshipData(object)) {
+        for (const relationship of object.relationships) {
+          await iteratee(relationship);
+        }
+      }
+    },
+  });
+}
```
fomentia, line 63: Definitely seems like `iterateEntityTypeIndex` and `iterateRelationshipTypeIndex` could have their implementation combined into a third function that both wrap, but that may not be worth it to you.

src/framework/jobState/FileSystemJobState/__tests__/FileSystemJobState.test.ts
```diff
@@ -145,6 +147,67 @@ describe('addRelationships', () => {
...
+
+    await jobState.iterateEntities({ _type: matchingType }, collectEntity);
+    expect(jobState.entities).toHaveLength(0);
+
+    expect(collectedEntities).toEqual(matchingEntities);
+  });
+});
+
+describe('iterateRelationships', () => {
+  test('should flush buffered relationshipos and iterate the relationship "_type" index stored on disk', async () => {
```
fomentia, line 178: Minor typeo here, "relationshipos".

PR #7, opened 2020-04-08
Implement step dependency graph building and execution

charlieduong94:
This PR implements functions for constructing a dependency graph using each step's `dependsOn` field and executing the steps in order to ensure that all dependencies have been met.

You'll notice that there's definitely some data missing from the context passed into the step's `executionHandler` right now. The interface exposed here will change when parts of the `IntegrationExecutionContext` are implemented (coming soon). 

Sorry this is a big PR, the majority of the code are tests. The function that actually orchestrates the execution of the step graph is fairly small. I detailed as much as I could in comments to detail why functions do what they do. I recommend starting with the `enqueueLeafSteps` function and following the chain there.


src/framework/execution/__tests__/dependencyGraph.test.ts
```diff
@@ -0,0 +1,571 @@
...
+      },
+    ];
+
+    const graph = buildStepDependencyGraph(steps);
+    await executeStepDependencyGraph(graph);
+
+    expect(jobStateFlushSpy).toHaveBeenCalledTimes(1);
+  });
+
+  test('should halt execution of dependant steps until current step has completed', async () => {
```
aiwilliams, line 125:
```suggestion
  test('should hold execution of dependent steps until current step has completed', async () => {
```

src/framework/execution/__tests__/dependencyGraph.test.ts
```diff
@@ -0,0 +1,571 @@
...
+
+    expect(spyA).toHaveBeenCalledTimes(1);
+    expect(spyB).toHaveBeenCalledTimes(1);
+    expect(spyC).toHaveBeenCalledTimes(1);
+
+    expect(spyA).toHaveBeenCalledBefore(spyB);
+    expect(spyB).toHaveBeenCalledBefore(spyC);
+  });
+
+  test('should steps with dependencies resulting in a status PARTIAL_SUCCESS_DUE_TO_DEPENDENCY_FAILURE should also assume the same status', async () => {
```
aiwilliams, line 251: This is a bit hard to parse. How do you feel about this?
```suggestion
  test('steps with dependencies resulting in PARTIAL_SUCCESS_DUE_TO_DEPENDENCY_FAILURE assume the same status', async () => {
```
I generally feel the `"should "` prefix isn't very helpful since every test is declaring what should be true about the subject under test.

src/framework/execution/__tests__/dependencyGraph.test.ts
```diff
@@ -0,0 +1,571 @@
...
+    expect(spyB).toHaveBeenCalledTimes(1);
+    expect(spyC).toHaveBeenCalledTimes(1);
+    expect(spyD).toHaveBeenCalledTimes(1);
+    expect(spyE).toHaveBeenCalledTimes(1);
+
+    expect(spyA).toHaveBeenCalledBefore(spyC);
+    expect(spyB).toHaveBeenCalledBefore(spyD);
+    expect(spyC).toHaveBeenCalledBefore(spyD);
+    expect(spyE).toHaveBeenCalledBefore(spyD);
+    expect(spyE).toHaveBeenCalledBefore(spyD);
```
aiwilliams, line 497: these are the same expectation.

src/framework/execution/__tests__/dependencyGraph.test.ts
charlieduong94: engrish is hard. I like your suggestion.

charlieduong94:
> Do you plan to support concurrent execution of steps at all? I think it's totally okay not to for now.

If concurrency is not explicitly set for the  `p-queue`, it defaults to unlimited concurrency. So all steps that are capable of running (leaf steps) are executed at the same time. As steps complete and other steps have their requirements fulfilled, their work will also be added to the queue and performed with unlimited concurrency. This behavior isn't obvious so I'll add a test.

src/framework/execution/__tests__/dependencyGraph.test.ts
charlieduong94: British English vs American English always gets me.

aiwilliams: Haha! I didn't even consider that when I read it.

aiwilliams:
> defaults to unlimited concurrency... This behavior isn't obvious so I'll add a test.

That would be great, as it will express that as the intention so others who fail to RTM for `p-queue` won't ask. 😆 Thanks @charlieduong94 !

src/framework/execution/__tests__/dependencyGraph.test.ts
```diff
@@ -0,0 +1,692 @@
...
+    await waitForExpect(() => {
+      spies.forEach((spy) => expect(spy).toHaveBeenCalledTimes(1));
+    });
+
+    resolvers.forEach((resolve) => resolve());
+
+    await executionPromise;
+  });
+
+  test('executes all leaf steps concurrently', async () => {
```
aiwilliams, line 220: This name is a duplicate of the test above, and even looks like it *might* be a complete copy of the previous test. 🤔

src/framework/execution/__tests__/dependencyGraph.test.ts
```diff
@@ -0,0 +1,692 @@
...
```
aiwilliams, line 258: This is that same name as well, but this test is definitely different.

src/framework/execution/__tests__/dependencyGraph.test.ts
charlieduong94: owoh, idk how I did that

PR #9, opened 2020-04-08
Implement function for loading instanceConfigFields from environment

charlieduong94:


charlieduong94:
Oh I should mention. I didn't include the `array` type that was on the managed sdk's `IntegrationInstanceConfigField`. Seemed like a rare occurance and was thinking the name `stringArray` or `stringList` type would be better (iirc usage of the `array` type was for comma delimited strings)

PR #10, opened 2020-04-09
Implement context creation and validation in executeIntegration function

charlieduong94:
Add types related to integration instances, invocation events, actions,
and context.

This PR leverages the config loading function from the previous for creating a local integration instance. That instance is used to initialize the integration execution context, which is then passed to the validator function.


charlieduong94:
Basic logging is coming next, then integrating the step execution into the `executeIntegration` function.

charlieduong94:
Actually, hold off on reviewing. Going to rework some stuff.

charlieduong94:
Hold off on reviewing pls

charlieduong94:
Okay, this is ready for review again. I initially had the invocation event and actions in this pr, but decided to just keep it simple for now since our main focus is setting up local integration dev. I plan to re-introduce it later.

src/framework/execution/__tests__/executeIntegration.test.ts
```diff
@@ -0,0 +1,18 @@
+import { IntegrationExecutionContext } from '../types';
+import { executeIntegrationLocally } from '../executeIntegration';
+import { LOCAL_INTEGRATION_INSTANCE } from '../instance';
+
+test('should execute validator function if provided in config', async () => {
+  const validate = jest.fn();
+
+  await executeIntegrationLocally({
+    invocationValidator: validate,
```
aiwilliams, line 9: I've wanted to rename this to `validateInvocation`, since it is a function that does something, not an object with properties (that is, verb vs. noun). What do you think?

src/framework/execution/executeIntegration.ts
```diff
@@ -0,0 +1,27 @@
...
+export function executeIntegrationLocally(config: IntegrationInvocationConfig) {
+  const instance = createIntegrationInstanceForLocalExecution(config);
+  const context: IntegrationExecutionContext = { instance };
+
+  return executeIntegration(context, config);
+}
+
+/**
+ * Executes an integration based and performs actions based on the
+ * event that was provided.
```
aiwilliams, line 20:
```suggestion
 * Executes an integration and performs actions based on the provided event.
```

src/framework/execution/types/instance.ts
```diff
@@ -0,0 +1,55 @@
+/**
+ * A stored user configuration for executing the integration defined by
+ * associated the `integrationDefinitionId`.
```
aiwilliams, line 3:
```suggestion
 * the associated `integrationDefinitionId`.
```

src/framework/execution/types/instance.ts
```diff
@@ -0,0 +1,55 @@
...
+   *
+   * Integrations supporting an offsite installation flow will set this value to
+   * `false` when the instance is created, indicating that the configuration is
+   * incomplete. The value will become `true` when the offsite flow has
+   * completed successfully.
+   *
+   * Integrations that do/will not support an offsite installation flow will
+   * leave this value `undefined`.
+   */
+  offsiteComplete?: boolean;
```
aiwilliams, line 54: I'm not totally sure, but I wonder if this should be in the types contained in the integration SDK. I believe this property is not seen by the integration itself, but is used by J1 to manage offsite flow. It's super convenient to store it on the instance, but isn't a concern outside of J1. What do you think of leaving it out so it isn't something integration developers have to care about?

@fomentia please correct me if I'm wrong, but I don't see any references to this in the GitHub integration.

src/framework/execution/__tests__/executeIntegration.test.ts
charlieduong94: I'm cool with that name.

src/framework/execution/types/instance.ts
charlieduong94: Ah, I probably should have checked the github integration before blindly adding this. I thought this was used for preventing an integration from doing anything when invoked, but I now see that this work is done by the scheduler. I'll remove it.

PR #11, opened 2020-04-09
Add basic integration logger

charlieduong94:
This PR mostly adopts the integration logger implementation that already existed in the managed integration sdk. 

I changed a few small things. For example, the creating a logger no longer requires an `INTEGRATION_NAME` environment variable to be supplied. I think we can fetch that from the integration definition during managed runs and provide a default value during local runs.

Note: this doesn't include any special handling of `logger.auth` or `logger.error` messages.

### DIFF
--- a/src/framework/execution/__tests__/logger.test.ts
+++ b/src/framework/execution/__tests__/logger.test.ts
@@ -1,0 +1,168 @@
+import { createIntegrationLogger } from '../logger';
+import Logger from 'bunyan';
+import { IntegrationInvocationConfig } from '../types';
+
+const invocationConfig = {} as IntegrationInvocationConfig;
+const loggerName = 'integration-logger';
+
+describe('logger.trace', () => {
+  test('includes verbose: true for downstream verbose log pruning', () => {
+    const integrationLogger = createIntegrationLogger(
+      loggerName,
+      invocationConfig,
+    );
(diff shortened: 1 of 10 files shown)


src/framework/execution/__tests__/logger.test.ts
```diff
@@ -0,0 +1,168 @@
+import { createIntegrationLogger } from '../logger';
+import Logger from 'bunyan';
+import { IntegrationInvocationConfig } from '../types';
+
+const invocationConfig = {} as IntegrationInvocationConfig;
+const loggerName = 'integration-logger';
+
+describe('logger.trace', () => {
+  test('includes verbose: true for downstream verbose log pruning', () => {
```
aiwilliams, line 9: Just wanted to point out... I'm not sure all the reasons there was a need to add an additional property, but I wonder if eventually we should change our log ingesting code to ignore `level=10` so that we don't have to add this additional property.

charlieduong94: iirc the decision for the allowing a verbose flag was to give a little more fine grain control over how logs are ignored while still giving services the ability to open the floodgates in the rare times when we wanted lower level logs to be published.

Also I think in the past our pipeline had a few edge cases where `info` logs could become excessively large and for some reason we wanted to keep those logs at that level (probably to still capture that data in CW).

PR #14, opened 2020-04-10
minor change: move @types/lodash to dev deps

charlieduong94:


charlieduong94:
Had to resolve a small conflict

PR #16, opened 2020-04-10
Integrate step start states into dependency graph execution

charlieduong94:


charlieduong94:
Note: There's a little bit of noise because I did a bit of refactoring and move some code related to step execution and result evaluation into a separate file.

src/framework/execution/__tests__/dependencyGraph.test.ts
```diff
@@ -649,4 +685,164 @@ describe('executeStepDependencyGraph', () => {
     expect(spyC).toHaveBeenCalledBefore(spyE);
     expect(spyD).toHaveBeenCalledBefore(spyE);
   });
+
+  test('does not execute step if marked ', async () => {
```
aiwilliams, line 689: Looks like an incomplete sentence.

src/framework/execution/__tests__/dependencyGraph.test.ts
```diff
@@ -649,4 +685,164 @@ describe('executeStepDependencyGraph', () => {
...
+    );
+
+    expect(spyA).toHaveBeenCalledTimes(0);
+
+    expect(result).toEqual([
+      {
+        id: 'a',
+        name: 'a',
+        types: [],
+        status: IntegrationStepResultStatus.NOT_EXECUTED,
```
aiwilliams, line 731: I wonder if we need a more specific status of `DISABLED`, to clarify the reason.

src/framework/execution/__tests__/dependencyGraph.test.ts
```diff
@@ -649,4 +685,164 @@ describe('executeStepDependencyGraph', () => {
...
+      },
+    ]);
+  });
+  test("does not execute step and it's dependencies if marked as disabled", async () => {
```
aiwilliams, line 735: 🕺

src/framework/execution/__tests__/validation.test.ts
```diff
@@ -0,0 +1,78 @@
+import { IntegrationStep, IntegrationStepStartStates } from '../types';
+
+import { validateStepStartStates } from '../validation';
+
+describe('validateStepStartStates', () => {
+  test('throws error if step is provided in start states', () => {
```
aiwilliams, line 6:
```suggestion
  test('throws error if unknown step is provided in start states', () => {
```

src/framework/execution/__tests__/validation.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      {
+        id: 'a',
+        name: 'a',
+        types: [],
+        executionHandler: jest.fn(),
+      },
+    ];
+
+    expect(() => validateStepStartStates(steps, states)).toThrow(
+      `Invalid step id "b" found in start states.`,
```
aiwilliams, line 25:
```suggestion
      `Unknown steps found in start states: "b".`,
```
Should this handle more than one, as the following does, so that all are reported in a single error?

src/framework/execution/__tests__/validation.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    );
+  });
+
+  test('throws steps are not accounted for in start states', () => {
```
aiwilliams, line 29:
```suggestion
  test('throws error when steps are not accounted for in start states', () => {
```

src/framework/execution/__tests__/validation.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      {
+        id: 'c',
+        name: 'c',
+        types: [],
+        executionHandler: jest.fn(),
+      },
+    ];
+
+    expect(() => validateStepStartStates(steps, states)).toThrow(
+      `Steps not defined in start states found: "b", "c"`,
```
aiwilliams, line 57:
```suggestion
      `Start states not found for : "b", "c"`,
```
I would prefer this because I think of the steps as being the authoritative set of things to do, with the start states as simply being input to those steps.

src/framework/execution/error.ts
```diff
@@ -16,3 +16,13 @@ export class IntegrationConfigValidationError extends Error
   implements IntegrationError {
   code = 'LocalConfigFieldTypeMismatch';
 }
+
+export class IntegrationStepStartStateInvalidStepIdError extends Error
+  implements IntegrationError {
+  code = 'InvalidStepIdSpecifiedInStartState';
```
aiwilliams, line 22:
```suggestion
  code = 'UnknownStepIdSpecifiedInStartState';
```
I prefer the term "Unknown" over "Invalid". What do you think, would it be more clear this way? Maybe "Undefined"?

src/framework/execution/__tests__/dependencyGraph.test.ts
charlieduong94: Ah yeah, `DISABLED` would be more clear match up with the `disabled` field in the start states. Will change.

src/framework/execution/__tests__/validation.test.ts
charlieduong94: Good suggestion, that would be a lot more friendly.

src/framework/execution/error.ts
charlieduong94: Going with "Unknown"

PR #17, opened 2020-04-10
Adopt converters from managed sdk

charlieduong94:
This just copies the converter utilities from the managed sdk into this repo. The other utils are coming.

### DIFF
--- a/src/framework/data/converters.ts
+++ b/src/framework/data/converters.ts
@@ -1,0 +1,122 @@
+import { createHash } from 'crypto';
+
+import camelCase from 'lodash/camelCase';
+
+const TRUE_REGEX = /^true$/i;
+const FALSE_REGEX = /^false$/i;
+const BASE10_NUMBER_REGEX = /^\d+$/;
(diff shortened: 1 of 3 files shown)


src/framework/data/converters.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+  }
+  return obj;
+}
+
+export function getTime(
+  time: Date | string | number | undefined,
+): number | undefined {
+  return time ? new Date(time).getTime() : undefined;
+}
```
aiwilliams, line 110: I've wondered how safe this function is because I've seen different date formats, including the stupid kind that include no time zone information. Perhaps we should consider changing this up or creating different forms that force integration developers to declare the time zone the provider assumes in all their poorly formatted date/time strings. 😆 Just want others to be aware.

charlieduong94: Yeah, this function isn't great. I def think we should use something like moment/luxon/date-fns and force a timezone to be provided.

PR #19, opened 2020-04-10
Adopt ip utilities from managed sdk. Added a few tests.

charlieduong94:
I didn't see any tests for these in the managed sdk, so I added a few. There is probably a third party util out there that does a better job, but I think we can investigate that later.


aiwilliams:
> There is probably a third party util out there that does a better job, but I think we can investigate that later.

I agree, there's got to be a good resource for this and it would be nice to direct integration developers to add that thing if they need to process IP address, which many integrations do not.

PR #20, opened 2020-04-10
Adopt create integration entity util

charlieduong94:
This copies over the `createIntegrationEntity` utilities and tweaks some of the types a bit to allow the  `Entity` type defined in this repo to play nicely.

### DIFF
--- a/__mocks__/fs.ts
+++ b/__mocks__/fs.ts
@@ -1,2 +1,4 @@
 import { fs } from 'memfs';
 export const promises = fs.promises;
+
+export const readdirSync = jest.requireActual('fs').readdirSync;
--- a/src/framework/types/entity.ts
+++ b/src/framework/types/entity.ts
@@ -1,15 +1,17 @@
 import { PersistedObject } from './persistedObject';
 
-export type Entity = PersistedObject &
-  EntityOverrides &
-  EntityAdditionalProperties;
+export type Entity = EntityCoreProperties & EntityAdditionalProperties;
+
+type EntityCoreProperties = PersistedObject & EntityOverrides & RawDataTracking;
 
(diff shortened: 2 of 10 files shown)


__mocks__/fs.ts
```diff
@@ -1,2 +1,4 @@
 import { fs } from 'memfs';
 export const promises = fs.promises;
+
+export const readdirSync = jest.requireActual('fs').readdirSync;
```
charlieduong94, line 4: This was needed to get schema loading from the `@juptierone/data-model` working.

src/framework/types/entity.ts
```diff
@@ -19,4 +21,49 @@ type EntityPropertyValue =
   | number[]
   | boolean
   | undefined
-  | null;
+  | null
+  | EntityRawData[];
```
charlieduong94, line 25: This is a tad yucky but was needed get the `_rawData` type to play nicely. tbh I'm considering just making `EntityPropertyValue` an `any` type.

aiwilliams: Can we make it essentially "not an object"?

fomentia: @charlieduong94 This makes sense to me tbh. Why do you say it's `yucky`?

charlieduong94: I mainly say this is yucky because it doesn't quite fulfill the constraints I want.

This allows for all additional fields (non core entity properties) to contain an `EntityRawData` array. From a typescript "loose" object perspective, it makes sense, but I don't want users to think that they can values they really can't.

charlieduong94:
Merging now. Going to figure out a fix for the type later. ATM if people are using the data model, it won't cause issues.

PR #23, opened 2020-04-11
Improve prepublish transpilation

charlieduong94:
Building now actually references the tsconfig.dist.json file.

The dist directory now follows the same structure as src.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "type-check": "tsc --noEmit",
     "test": "jest",
     "test:ci": "yarn lint && yarn type-check && yarn test",
-    "prepublishOnly": "tsc --declaration"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,7 +17,8 @@
     "type-check": "tsc --noEmit",
     "test": "jest",
     "test:ci": "yarn lint && yarn type-check && yarn test",
-    "prepublishOnly": "tsc --declaration"
+    "build": "tsc -p tsconfig.dist.json --declaration",
```
charlieduong94, line 20: Made this a separate command because I found `yarn build --watch` to be helpful when working on the cli.

aiwilliams:
Thanks a ton for explanations of changes like this @charlieduong94, and some insight into the workflow it supports.

PR #24, opened 2020-04-11
Tweak job state to utilize a shared object store.

charlieduong94:
Decided to go ahead and just do this now. 

This PR makes some changes to the job state so that a shared object store is used for collecting data. The main motivation for doing this is to reduce the amount of memory used by integrations. Prior to this change, each step had it's own object store with their own entity and relationship buffers, meaning integrations with many concurrent steps could consume a lot of memory. 

This PR is a bit big so I did my best to break up the commits.

First two commits (4e26132 and 93b4689) are just refactor commits to start moving things around (changing `FileSystemJobState` to `FileSystemGraphObjectStore`).

The last 4 commits are the ones that actually have the changes that make this work.


src/framework/storage/FileSystemGraphObjectStore/path.ts
```diff
@@ -1,17 +1,17 @@
...
-export function buildStepCollectionFilePath({
-  step,
+export function buildObjectCollectionFilePath({
+  storageDirectoryPath,
   collectionType,
   filename,
-}: BuildStepCollectionFilePathInput) {
-  return ['graph', step, collectionType, filename].join('/');
+}: BuildObjectCollectionFilePathInput) {
+  return ['graph', storageDirectoryPath, collectionType, filename].join('/');
```
aiwilliams, line 14: Ah, here's that line I wondered about. Why not `path.join()`?

src/framework/storage/FileSystemGraphObjectStore/__tests__/flushDataToDisk.test.ts
```diff
@@ -24,16 +24,16 @@ test('should group objects by "_type" and write them to separate files', async (
   const allEntities = randomizeOrder(flatten(Object.values(testEntityData)));
 
   const cacheDirectory = '/' + uuid();
-  const step = uuid();
+  const storageDirectoryPath = uuid();
 
   await flushDataToDisk({
     cacheDirectory,
-    step,
+    storageDirectoryPath,
```
aiwilliams, line 31: I think the `cacheDirectory` name is more confusing now than before. These two arguments are joined with `/graph/` between them, which makes me wonder now why it's not just a single argument of `storageDirectoryPath`.

src/framework/storage/FileSystemGraphObjectStore/__tests__/flushDataToDisk.test.ts
```diff
@@ -24,16 +24,16 @@ test('should group objects by "_type" and write them to separate files', async (
...
     collectionType: 'entities',
     data: allEntities,
   });
 
-  const entitiesDirectory = `${cacheDirectory}/graph/${step}/entities`;
+  const entitiesDirectory = `${cacheDirectory}/graph/${storageDirectoryPath}/entities`;
```
aiwilliams, line 36: I'm now thinking this should use `path.join()` 😬

src/framework/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -0,0 +1,138 @@
...
+
+  async flush() {
+    await Promise.all([
+      this.flushEntitiesToDisk(),
+      this.flushRelationshipsToDisk(),
+    ]);
+  }
+
+  async flushEntitiesToDisk() {
+    await this.lockOperation(
```
aiwilliams, line 93: I think it'd be great to get some of the commentary you added to the commit for this semaphore into this class's documentation. That was super helpful for explaining the purpose.

src/framework/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -0,0 +1,138 @@
...
+} from './indices';
+
+export const GRAPH_OBJECT_BUFFER_THRESHOLD = 500; // arbitrarily selected, subject to tuning
+
+interface FileSystemGraphObjectStoreInput {
+  cacheDirectory?: string;
+}
+
+export class FileSystemGraphObjectStore {
+  readonly cacheDirectory?: string;
```
aiwilliams, line 22: Since this is optional, I now think the name could be `storageDirectoryPathPrefix`.

src/framework/execution/jobState.ts
```diff
@@ -0,0 +1,23 @@
+import { JobState, IntegrationStep } from './types';
+
+import { FileSystemGraphObjectStore } from '../storage';
+
+export function createStepJobState(
+  step: IntegrationStep,
+  graphObjectStore: FileSystemGraphObjectStore,
```
aiwilliams, line 7: I imagine this will take a `GraphObjectStore` interface at some point, eh?

src/framework/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
charlieduong94: Will add.

src/framework/storage/FileSystemGraphObjectStore/path.ts
charlieduong94: tbh I forgot that was a function. Will change in separate PR.

src/framework/storage/FileSystemGraphObjectStore/__tests__/flushDataToDisk.test.ts
charlieduong94: Yeah, I'm thinking of removing the `cacheDirectory` option entirely since it's only there to support a future option (maybe) of allowing users to specify where they want data stored from an integration run. The more I think about it, it doesn't seem like the option is really worth supporting. Will remove in a separate pr.

PR #26, opened 2020-04-13
Add project convention file loading for later use with cli

charlieduong94:
This PR adds loading in an integration's invocation config based on an project structure convention.

Expected structure relative to where the cli is run.
```
src/instanceConfigFields.json
src/validateInvocation.ts
src/getStepStartStates.ts
src/steps/*
```
ATM this works for both typescript and javascript projects, but I only wrote tests for typescript projects now since that's our target audience this week. I'll add tests to ensure plain js projects work down the road (would be quick).

Let me know if you want changes to where files should live.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
   "repository": "git@github.com:JupiterOne/integration-sdk.git",
   "author": "JupiterOne <dev@jupiterone.io>",
   "license": "MPL-2.0",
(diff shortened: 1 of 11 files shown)


charlieduong94:
btw, there's no real validation here yet. Going to tackle that separately since it's not really stopping anyone from building integrations.

PR #27, opened 2020-04-13
Add collect command for running integration locally

charlieduong94:


charlieduong94:
@aiwilliams brought up a good point that the integration config loading isn't necessarily a cli specific thing, so I'll move that in a later PR.

charlieduong94:
Output:
cli

PR #34, opened 2020-04-14
Make loading of ts-node optional for javascript developers

charlieduong94:
So I noticed that typescript is actually a peer dependency of `ts-node`, meaning that users pulling in the sdk won't actually get all of typescript (which is good).

But this also means that because of how ts-node is loaded, plain js users won't be able to actually load configs without having both typescript and ts-node installed (which is bad). 

This PR fixes that by performing a little detection of typescript files prior to requiring ts-node. This makes ts-node no longer needs to be an explicit dependency of the project. 

I moved ts-node out of the `dependencies` and into `peerDependencies`. I also added it to `devDependencies` since we need it here for testing.


src/cli/config.ts
```diff
@@ -96,3 +98,21 @@ export function loadModuleContent<T>(relativePath: string): T | undefined {
 
   return integrationModule?.default ?? integrationModule;
 }
+
+async function isTypescriptPresent() {
+  const paths = await globby(path.join('src', '**', '*.ts'));
```
charlieduong94, line 103: This module is quite nice. I'll might replace the implementation of `walkDirectory` with this.

package.json
```diff
@@ -57,9 +57,13 @@
     "memfs": "^3.1.2",
     "prettier": "^2.0.2",
     "ts-jest": "^25.3.0",
+    "ts-node": "^8.8.2",
     "typescript": "^3.8.3",
     "wait-for-expect": "^3.0.2"
   },
+  "peerDependencies": {
+    "ts-node": "^8.8.2"
+  },
```
aiwilliams, line 66: Does this cause a warning if they don't install `ts-node`?

src/cli/__tests__/config.test.ts
```diff
@@ -12,10 +12,7 @@ import { loadProjectStructure } from './loadProjectStructure';
 jest.mock('../log');
 
 afterEach(() => {
-  // clear require cache
-  Object.keys(require.cache).forEach((modulePath) => {
-    delete require.cache[modulePath];
-  });
+  jest.resetModules();
```
aiwilliams, line 15: Nice.

charlieduong94: oh I should mention that I found out last night that jest's `require` doesn't actually implement the `require.cache`.

package.json
charlieduong94: Yeah, it will display a warning. tbh I was kinda on the fence on whether to add `ts-node` to peer deps knowning that it was a peer dep, but unfortunately there's no concept of optional peer dependencies that do not cause a warning to be logged. 

I settled on that idea that it was probably better to be explicit about peers, but I have no strong opinions on this.

aiwilliams: I feel like warnings, when not resolved, are "broken windows", inviting vandalism at worst, or confusing people who wonder why they're being warned at the very beginning of their experience.

I would be inclined to handle this by not having the peer dependency - since it is optional - and simply have the code that requires `ts-node` handle the require error, providing a nice output saying "It looks like you're building with TypeScript. Please add `ts-node` as a dependency of your project to allow the CLI to compile your code." (or something like that). We would document the need as well, of course, in the bit that says a project can be done as TS or not.

That said, it's nothing that needs to be resolved right now. Also, I'd be curious how the community does optional dependencies in general.

charlieduong94: kk, I'll remove the peer dep in one of the next prs prior to the next publish

charlieduong94: Actually, looks like I created a conflict, removing now

PR #35, opened 2020-04-14
Add template project

charlieduong94:
I just copied all of the config that this project uses into the template.

I separated the configs into separate files for the template since doing it later would have been a pain. Later down the road we'll add... templating to the template project and turn it into a generator.


template/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+#### Installing dependencies
+
+From the root of this project, run `npm install` to install dependencies. If you
+have `yarn` installed, you can install dependencies by running `yarn`.
+
+#### Loading credentials
+
+Create a `.env` file at the root of this project and add environment variables
+to match what is in `src/instanceConfigFields.json`. The `.env` file is is
```
aiwilliams, line 23:
```suggestion
to match what is in `src/instanceConfigFields.json`. The `.env` file is
```

template/README.md
```diff
@@ -0,0 +1,87 @@
...
+src/
+  /instanceConfigFields.json
+  /validateInvocation.ts
+  /getStepStartStates.ts
+  steps/
+    exampleStep.ts
+    // add additional steps here
+```
+
+Each of the files listed above contribute creating an
```
aiwilliams, line 73:
```suggestion
Each of the files listed above contribute to creating an
```

template/README.md
```diff
@@ -0,0 +1,87 @@
...
+[integration configuration](https://github.com/JupiterOne/integration-sdk/blob/master/docs/development.md#the-integration-framework).
+
+Additional files can be placed under `src` and referenced from each of the
+integration files.
+
+The template project hosted
+[here](https://github.com/JupiterOne/integration-sdk/tree/master/template)
+provides an simple example of how an integration can be setup.
```
aiwilliams, line 81:
```suggestion
provides a simple example of how an integration can be setup.
```

template/src/steps/fetchAccounts.ts
```diff
@@ -0,0 +1,38 @@
+import {
+  IntegrationStep,
+  IntegrationStepExecutionContext,
+  createIntegrationEntity,
+} from '@jupiterone/integration-sdk';
+
+const step: IntegrationStep = {
+  id: 'fetch-accounts',
+  name: 'Fetch accounts',
+  types: ['my-integration-account'],
```
aiwilliams, line 10: We always use the form `my_integration_account`. I think we'll want to encourage that for consistency. Sorry if I've been missing that in all the other codes!

template/src/validateInvocation.ts
```diff
@@ -0,0 +1,14 @@
...
+) {
+  context.logger.info(
+    {
+      instance: context.instance,
+    },
+    'Validating integration config...',
+  );
+
+  context.logger.info('Integration instance is valid!');
+}
```
aiwilliams, line 14: We may want to include some guidance on what this function needs to do when invocation is not valid. I think it would be as simple as:
```
if (valid) {
  context.logger.info('Integration instance is valid!');
} else {
  throw new Error("Failed to authenticate with provided credentials!")
}
```

template/src/steps/fetchAccounts.ts
charlieduong94: ah biig oops. I think it's been consistent in other places.

template/src/steps/fetchAccounts.ts
```diff
@@ -0,0 +1,38 @@
...
+              _type: 'my-integration-account',
+              _class: 'Account',
+            },
+          },
+        }),
+      ]);
+    } catch (err) {
+      logger.error(err, 'Unexpected error occurred');
+      throw err;
+    }
```
aiwilliams, line 34: I think we should encourage just allowing the error to be thrown (no `try/catch` needed), since the runtime environment will log the error. This can simplify steps a good deal as folks don't have to think about it unless they want to do something more than logging. This example suggests they must log errors themselves.

PR #42, opened 2020-04-15
Fix compatibility issues with windows

charlieduong94:


src/cli/config.ts
```diff
@@ -100,7 +100,14 @@ export function loadModuleContent<T>(relativePath: string): T | undefined {
...
 async function isTypescriptPresent() {
-  const paths = await globby(path.join('src', '**', '*.ts'));
+  // NOTE: this does not use path.join because globby
+  // (which uses fast-glob, which uses micromatch)
+  // requires that forward slashes are used.
+  //
+  // Refs:
+  // - https://github.com/mrmlnc/fast-glob#pattern-syntax
+  // - https://github.com/micromatch/micromatch#backslashes
+  const paths = await globby('src/**/*.ts');
```
aiwilliams, line 110: 😍 the docs on this @charlieduong94 !

charlieduong94:
Yup! I noticed the warnings. Follow up PR will suppress them.

charlieduong94:
I should note that atm we are running both travis and gh builds. Once we figure out publishing from that, we can drop travis entirely.

Also, I wasn't going to add gh builds in the PR originally, but I had found that the Travis windows builds were still in beta and quite buggy (tests pass but build would never complete).

charlieduong94:
cc and shoutout to @austinkelleher

His terraform provider pr is what inspired me to get matrix builds up for this.

PR #43, opened 2020-04-15
Suppress eslint warnings

charlieduong94:
We have these rules off in our [internal eslint config](https://bitbucket.org/lifeomic/typescript-tools/src/27c6707b50776a6d361cc9d41c1e6fd1a35e5831/config/eslint-base.json#lines-16).

The tests that have the `expect.assertions(0)` do not have assertions because they are typescript "fit" tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,9 @@
         {
           "functions": false
         }
-      ]
+      ],
+      "@typescript-eslint/explicit-function-return-type": "off",
+      "@typescript-eslint/no-explicit-any": "off"
     }
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -106,7 +106,9 @@
         {
           "functions": false
         }
-      ]
+      ],
+      "@typescript-eslint/explicit-function-return-type": "off",
+      "@typescript-eslint/no-explicit-any": "off"
```
aiwilliams, line 111: 😭 😭

PR #44, opened 2020-04-15
Add docs directory to template

charlieduong94:
Added stub development, data model, and authentication docs
to the template. The goals is to encourage developers to write
documentation as they build out integrations.

### DIFF
--- a/template/docs/authentication.md
+++ b/template/docs/authentication.md
@@ -1,0 +1,3 @@
+# Authentication
+
+Supply details here for information on how to authenticate with a provider.
--- a/template/docs/authentication.md
+++ b/template/docs/authentication.md
@@ -1,3 +1,0 @@
-# Authentication
-
-Supply details here for information on how to authenticate with a provider.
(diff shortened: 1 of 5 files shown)


charlieduong94:
Once this is approved, I'll update the https://github.com/JupiterOne/graph-snowflake and https://github.com/JupiterOne/graph-trend-micro projects to include these files.

template/docs/authentication.md
```diff
@@ -0,0 +1,3 @@
+# Authentication
```
erkangz, line 1: can we combine Authentication and Data Model into a single file? That would serve as a user-facing integration doc that we can easily include in `docs` to publish to support site.

charlieduong94: Sure. We can do that. What do you think about placing that under perhaps an `integration.md` file?

charlieduong94: or maybe `jupiterone.md`?

PR #53, opened 2020-04-20
Use .env to load instance config when using mock context

charlieduong94:
This should help out with setting up test scenarios when developing locally without having devs need to manually set environment variables or use `dotenv`.

I removed some logging from the individual `load*` functions exposed by `src/framework/config.ts` because I didn't want those messages to add to test output. IMO those messages are not super useful anyway. If we find it's worth bringing back those messages, I say we do it with more useful messaging (ex. "x file was detected, loading") and in a separate PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- `createMockExecutionContext` and `createMockStepExecutionContext` will now
+  attempt utilize the fields specified in `src/instanceConfigField.json` to
+  populate the instance `config` when testing.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,14 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- `createMockExecutionContext` and `createMockStepExecutionContext` will now
+  attempt utilize the fields specified in `src/instanceConfigField.json` to
+  populate the instance `config` when testing.
+- Messages about loading in individual parts of the config are no longer shown
+  when running `j1-integration collect`.
```
aiwilliams, line 17: 💯

PR #54, opened 2020-04-20
Document testing guidelines

aiwilliams:
Born from https://github.com/JupiterOne/graph-trend-micro/pull/6#pullrequestreview-395801553. As always, a work in progress, in the end, the best approach should win, and there will need to be flexibility, but patterns are meant to capture things learned, best practices, etc., so... Would love to revisit this regularly as we implement tests across a number of integrations.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,8 +1,7 @@
 # Integration development
 
-This project exposes both a module exposing a framework for constructing
-integrations and also a CLI to assist with collecting and publishing data with
-JupiterOne.
+This project exposes both a a framework for constructing integrations and a CLI
+to assist with collecting and publishing data with JupiterOne.
 
 For the purpose of synchronizing data, developers only need to collect the "new
 state of the world" from a provider and have the data sent to JupiterOne. Data
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,6 +1,6 @@
 # Integration development
 
-This project exposes both a a framework for constructing integrations and a CLI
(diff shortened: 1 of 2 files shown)


docs/development.md
```diff
@@ -1,8 +1,7 @@
 # Integration development
 
-This project exposes both a module exposing a framework for constructing
-integrations and also a CLI to assist with collecting and publishing data with
-JupiterOne.
+This project exposes both a a framework for constructing integrations and a CLI
```
charlieduong94, line 3:
```suggestion
This project exposes both a framework for constructing integrations and a CLI
```

docs/testing.md
```diff
@@ -1,13 +1,89 @@
...
+
+Strive to isolate unit tests according to:
+
+1. Client authentication and data fetching, including response error handling
+1. Converting source data to entity/relationship objects, ensuring properties
+   are transferred as expected
+1. Producing entities with the right relationships based on a known set of
+   source data (do not test properties)
+
+Use [Polly.js](https://netflix.github.io/pollyjs) for testing the client code.
```
austin-rausch, line 61: I would specify making this where applicable.  With the snowflake SDK it makes weird requests (pre flight exec, post flight exec and return data in difficult formats, also pretty sure some of it is over raw TCP) and the return data is hard to figure out what is what.

aiwilliams: Okay, will do!

PR #56, opened 2020-04-20
Visualize command

softwarewright:
- Adding the visualize command
- Integrating with Vis for graph visualization https://visjs.github.io/vis-network/docs/network/


src/framework/visualization/retrieveIntegrationData.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { Relationship, Entity } from '../types';
+import { readJsonFile } from '../../fileSystem';
+import { IntegrationData } from './types/IntegrationData';
+import { IntegrationMissingCollectJSON } from './error';
+
+/**
+ * Retrieve integration data from the JSON files from the cache directory
+ */
+export async function retrieveIntegrationData(
```
aiwilliams, line 12: I wonder if this behavior could belong to the `FileSystemGraphObjectStore`? It seems like the kind of storage related operation it should perform and then, should we in the future have other types of graph object storage, the visualizer won't know/care how entities/relationships are loaded.

src/framework/visualization/retrieveIntegrationData.ts
```diff
@@ -0,0 +1,45 @@
...
+  const [entityPaths, relationshipPaths] = await Promise.all([
+    globby([path.posix.join(integrationPath, 'index', 'entities')]),
+    globby([path.posix.join(integrationPath, 'index', 'relationships')]),
+  ]);
+
+  if (entityPaths.length === 0 && relationshipPaths.length === 0) {
+    throw new IntegrationMissingCollectJSON(
+      'Unable to find any entities or relationships, have you run the "j1-integration collect" command',
+    );
+  }
```
aiwilliams, line 24: It seems to me that something at a higher layer could report there is no data to display, and would also include information about where it's looking exactly. Also, knowing about the command that could generate the data isn't a concern the storage mechanism should know about, and I would imagine this visualization tool could be used with entities/relationships generated by other tools.

src/framework/visualization/generateVizTemplate.ts
```diff
@@ -0,0 +1,35 @@
+import { VizNode, VizEdge, VizOptions } from './types/viz';
+
+/**
+ * Creates the html for the viz template
+ */
+export function generateVizTemplate(
```
aiwilliams, line 6: Is this generating template, or is it generating the actual content? Perhaps `generateVizHTML` or `generateVizDocument`?

src/fileSystem.ts
```diff
@@ -32,11 +32,31 @@ export async function writeJsonToPath({
...
+}: WriteContentToPathInput) {
   const directory = getRootStorageDirectory();
   const fullPath = path.resolve(directory, relativePath);
 
   await ensurePathCanBeWrittenTo(fullPath);
-  await fs.writeFile(fullPath, JSON.stringify(data, null, 2), 'utf8');
+  await fs.writeFile(fullPath, content, 'utf8');
+}
+
+export async function readJsonFile<T>(path: string) {
```
aiwilliams, line 57: How about `readJsonFromPath`, to balance the name with `writeJsonToPath`? For both functions, I wonder about dropping the `ToPath` bit from the function name since the input object has function calls looking like `writeJson({ path: '/somewhere', content: dataToStringify })`.

src/cli/commands/visualize.ts
```diff
@@ -0,0 +1,14 @@
...
+import * as log from '../../log';
+
+import { generateVisualization } from '../../framework';
+
+export function visualize() {
+  return createCommand('visualize')
+    .description('Visualizes the collected data as a graph')
+    .action(async () => {
+      const graph = await generateVisualization();
+      log.info(`Visualize Integration SDK graph here: ${graph}`);
```
aiwilliams, line 12: How about dropping "Integration SDK" adjective and do just `Visualize graph here`?

src/cli/__tests__/cli.test.ts
```diff
@@ -47,3 +51,40 @@ describe('collect', () => {
     });
   });
 });
+
+describe('visualize', () => {
+  beforeEach(() => {
+    loadProjectStructure('typeScriptVisualizeProject');
```
aiwilliams, line 57: I wonder if we need to maintain a complete project like this to generate a couple of entities and relationships for testing the vis.js HTML generation. If the visualizer just took in a `FileSystemGraphObjectStore`, and that had the ability to iterate it's entities and relationships, then it would be pretty easy to pass in an in-memory store and see the generated HTML in a test.

src/framework/visualization/generateVizTemplate.ts
softwarewright: Good point, it does generate the HTML. I'll update that.

src/cli/commands/visualize.ts
softwarewright: Sounds good.

src/fileSystem.ts
softwarewright: I can make the first change you suggested. I understand what you are saying about dropping the `toPath` piece I would be fine with removing it. I think it does make the code a bit cleaner.

src/framework/visualization/retrieveIntegrationData.ts
softwarewright: That does make more sense looking at that file, I didn't realize it existed. I'll make that change.

src/cli/__tests__/cli.test.ts
softwarewright: In this test, I'm following the pattern of the `collect` command and creating an integration test where we see if the CLI works. I've already tested the lower level functions with mocked file systems to see if they work. If we want we could also just remove this test since there is one testing the integration between collect and visualize already. Not sure what you do think @aiwilliams.

aiwilliams: I think it's fine to keep, we'll see if it ever feels expensive to maintain! Thanks @softwarewright for considering.

src/framework/visualization/retrieveIntegrationData.ts
softwarewright: Fair point, do you think it would be good for this to happen in the visualize command file as a check when we could pass the data in or do you think it belongs somewhere else.

As for this possibly working with other tools, it's very possible. Would it be a good idea for the collect command to generate the relationships after running for the visualize command to consume? Then we can make the visualize command default to a location that the collect command would write to. What are your thoughts?

This is all that is required for creating a graph
```
    var nodes = new vis.DataSet([
        {id: 1, label: 'Node 1'},
        {id: 2, label: 'Node 2'},
        {id: 3, label: 'Node 3'},
        {id: 4, label: 'Node 4'},
        {id: 5, label: 'Node 5'}
    ]);

    // create an array with edges
    var edges = new vis.DataSet([
        {from: 1, to: 3},
        {from: 1, to: 2},
        {from: 2, to: 4},
        {from: 2, to: 5}
    ]);
```
https://visjs.github.io/vis-network/docs/network/

src/framework/visualization/retrieveIntegrationData.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Retrieve integration data from the JSON files from the cache directory
+ */
+export async function retrieveIntegrationData(
+  integrationPath: string,
+): Promise<IntegrationData> {
+  const [entityPaths, relationshipPaths] = await Promise.all([
+    globby([upath.toUnix(`${integrationPath}/index/entities/**/*.json`)]),
+    globby([upath.toUnix(`${integrationPath}/index/relationships/**/*.json`)]),
```
charlieduong94, line 17: ATM the `.j1-integration/index` directory only contains symlinks to files in the `.j1-integration/graph` directory and is grouped by `_type`, but this could potentially change in the future if we decide to add support for grouping by other properties. 

Also, one thing we are planning for with this sdk is for developers writing integrations in other languages to use the CLI to perform synchronization with our persister (see [here](https://github.com/JupiterOne/integration-sdk/blob/master/docs/development.md#keeping-the-door-open-for-developers-not-using-javascript-or-typescript)). Those developers my not even end up having an `index` directory. 

The goal was to just allow those developers to just write to the `.j1-integration/graph` with directory structure they want and still be able to perform synchronization. The loose requirements allows for developers to not worry about matching our own structure since they may end up coming up with a different structure that suites them better (they may not have the concept of steps). It also allows for people to just write json to the `.j1-integration/graph` directory by hand for the _super_ simple integrations where devs just want to use j1 as a simple config store. I'd like to still allow for visualizations to work for these developers.

src/cli/__tests__/__snapshots__/cli.test.ts.snap
```diff
@@ -0,0 +1,55 @@
...
+            var data = {
+              nodes: nodes,
+              edges: edges
+            };
+            var options = {\\"edges\\":{\\"arrows\\":{\\"to\\":{\\"enabled\\":true}}}};
+            var network = new vis.Network(container, data, options);
+          </script>
+    </body>
+    </html>"
+`;
```
charlieduong94, line 28: I generally dislike snapshot testing (unless using actual image snapshots for component testing), but this is probably one of the few cases where I think I'm cool with it since it can actually help catch bad changes.

src/framework/visualization/types/viz.ts
```diff
@@ -0,0 +1,21 @@
+export interface VizOptions {
```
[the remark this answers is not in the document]
softwarewright: Lol yes, and thank you I just realized I spelled that wrong everywhere. Great catch.

.editorconfig
```diff
@@ -0,0 +1,4 @@
+[*.ts]
```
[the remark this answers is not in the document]
softwarewright: Can do!

src/framework/visualization/generateVizHTML.ts
```diff
@@ -0,0 +1,35 @@
...
+  return `<!DOCTYPE html>
+    <html lang="en">
+    <head>
+        <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <title>Integration SDK Graph</title>
+    </head>
+    <body style="height:100vh;width:100%">
+        <div id="integration-graph" style="height:100vh;width:100%"></div>
+        <script type="text/javascript" src="https://unpkg.com/vis-network/standalone/umd/vis-network.min.js"></script>
```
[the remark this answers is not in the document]
softwarewright: That makes sense, I'll update that.

src/framework/visualization/retrieveIntegrationData.ts
softwarewright: Got it, so it is better to just glob the graph directory instead?

src/__tests__/__fixtures__/typeScriptIntegrationProject/src/steps/fetchAccounts.ts
```diff
@@ -0,0 +1,29 @@
+import {
+  createIntegrationEntity,
+  IntegrationStepExecutionContext,
+} from '../../../../../framework';
+
+export default {
+  id: 'fetch-accounts',
+  name: 'Fetch Accounts',
+  types: ['my_account'],
+  // executionHandler: () => {},
```
aiwilliams, line 10: This looks like an old comment.

src/cli/__tests__/cli.test.ts
aiwilliams: Considering @charlieduong94 comments about having the visualization tool work off any valid set of files, I'm growing convinced we can avoid running the `collect` command against a fixture integration, and instead just provide a memfs file setup to the generator.

src/framework/visualization/retrieveIntegrationData.ts
aiwilliams: Reading @charlieduong94 comments:

> [allow] for people to just write json to the .j1-integration/graph directory by hand for the super simple integrations where devs just want to use j1 as a simple config store. I'd like to still allow for visualizations to work for these developers.

Considering this goal, and having just reviewed `FileSystemGraphObjectStore` code, it seems now it would be best for the visualization code to avoid the `FileSystemGraphObjectStore`. That thing very much expects a certain file system structure, whereas this visualization tool will just walk a tree and figure out what is in the json files.

src/framework/visualization/retrieveIntegrationData.ts
aiwilliams: I wonder if the program could generate the HTML, with empty data (do not throw an error), and have the HTML include something saying there is no data, and then have the command output the location of the graph HTML file, and perhaps also a warn statement saying there was no data found (which it can know by looking at the generated HTML file). I imagine the CLI command will take something like `--data-directory`, which defaults to `.j1-integration/graph`, so it would know the path and could say it produced an empty graph.

src/__tests__/__fixtures__/typeScriptIntegrationProject/src/steps/fetchAccounts.ts
softwarewright: I didn't remove that when I was testing something, good catch.

src/framework/visualization/retrieveIntegrationData.ts
softwarewright: That makes sense, I'll look into adding a warning to the HTML file when there are no relationships and remove the exception above. I think a log to the console would be good as well telling the user that no nodes were found. I'll also add the optional --data-directory param.

src/cli/__tests__/cli.test.ts
softwarewright: That makes sense, would we still want to have at least one integration test between collect and visualize to make sure we haven't broken the visualization with our own commands? Then we can figure out how we want to move this visualize test to memfs.

src/framework/visualization/generateVisualization.ts
```diff
@@ -0,0 +1,34 @@
...
+export async function generateVisualization(): Promise<string> {
+  const integrationPath = getRootStorageDirectory();
+
+  const { entities, relationships } = await retrieveIntegrationData(
+    integrationPath,
+  );
+
+  const nodeDataSets = entities.map((entity) => ({
+    id: entity._key,
+    label: entity.name.toString(),
```
austin-rausch, line 18: I think `name` may be on most entities, but I believe will still use `displayName` in the graph viewer like you have with relationships below.

src/framework/visualization/generateVisualization.ts
```diff
@@ -0,0 +1,34 @@
...
+  }));
+  const edgeDataSets = relationships.map((relationship) => ({
+    from: relationship._fromEntityKey.toString(),
+    to: relationship._toEntityKey.toString(),
+    label: relationship.displayName.toString(),
+  }));
```
austin-rausch, line 24: Aren't these fields already typed as string? I don't think the extra safety of `.toString()` is necessary. Its also missing on the `_key` above for the `nodeDataSets`.

src/framework/visualization/retrieveIntegrationData.ts
```diff
@@ -0,0 +1,45 @@
...
+    throw new IntegrationMissingCollectJSON(
+      'Unable to find any entities or relationships, have you run the "j1-integration collect" command',
+    );
+  }
+
+  const entities: Entity[] = [];
+  const relationships: Relationship[] = [];
+
+  for (const path of entityPaths) {
+    const parsedEntities = await readJsonFromPath<{ entities: Entity[] }>(path);
```
austin-rausch, line 30: Nitpicky comment, feel free to ignore: You could have done all of these in parallel. (not important though its just a developer tool...)

src/cli/__tests__/__snapshots__/cli.test.ts.snap
charlieduong94: Oh related to this. Can you add the [--ci](https://jestjs.io/docs/en/cli.html#--ci) flag to the `test:ci` script? It will help ensure that unwanted tweaks to the snapshot are caught.

src/cli/__tests__/cli.test.ts
aiwilliams: I see what you mean, yeah, it would be good to know that our collection output can be visualized.

src/framework/visualization/retrieveIntegrationData.ts
charlieduong94: Yep!

src/framework/visualization/generateVisualization.ts
softwarewright: Talked with @charlieduong94 there was an issue with the typing and I just needed to be more specific with my types to avoid needing to cast. Good catch!

package.json
```diff
@@ -130,7 +132,8 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged",
+      "pre-push": "yarn build"
```
softwarewright, line 136: @charlieduong94 What do you think about this, I figured this can help catch things that would fail on ci before they get pushed up.

src/framework/visualization/generateVisualization.ts
softwarewright: Got it, updated it.

package.json
charlieduong94: :+1: Seems good to me

package.json
```diff
@@ -27,13 +27,14 @@
     "format": "prettier --write '**/*.{ts,js,json,css,md,yml}'",
     "type-check": "tsc --noEmit",
     "test": "jest",
-    "test:ci": "yarn lint && yarn type-check && yarn test",
+    "test:ci": "yarn lint && yarn type-check && yarn test -- --ci",
```
charlieduong94: I don't think the double dashes are necessary with yarn. (I know it is for `npm run` commands though)

softwarewright: I'm not sure why that's still there I already removed them

src/cli/__tests__/cli.test.ts
```diff
@@ -71,15 +80,20 @@ describe('visualize', () => {
...
 
 describe('collect/visualize integration', () => {
+  let htmlFileLocation;
   beforeEach(() => {
     loadProjectStructure('typeScriptIntegrationProject');
+
+    htmlFileLocation = path.join(
+      path.resolve(process.cwd(), '.j1-integration/graph'),
+      'index.html',
+    );
```
charlieduong94, line 94: `resolve` accepts more than two param so you can technically pass in  `index.html`  after `.j1-integration/graph`.

template/package.json
```diff
@@ -5,6 +5,7 @@
   "license": "MPL-2.0",
   "scripts": {
     "start": "j1-integration collect",
+    "visual": "j1-integration visualize",
```
charlieduong94, line 8: Thoughts on maybe making the script name `visualize` or perhaps even `graph`?

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "outDir": "dist",
     "target": "es2018",
-    "lib": ["es2018"],
+    "lib": ["es2018", "DOM"],
```
charlieduong94, line 5: Oh I didn't notice this before. Is `DOM` necessary for something?

package.json
charlieduong94: ahh k I see that now

tsconfig.json
softwarewright: For the vis types

template/package.json
softwarewright: I like graph, I'll change it to that.

src/cli/__tests__/cli.test.ts
softwarewright: Oh, you're right! I didn't notice that in the IntelliSense. I'll fix that now.

tsconfig.json
charlieduong94: ahh okay

PR #57, opened 2020-04-21
Generate instance config in mock context if instanceConfig option is not provided

charlieduong94:


docs/testing.md
```diff
@@ -18,24 +18,50 @@ If the `instanceConfig` option is not set, the SDK will read the
 `src/instanceConfigFields.json` and load in values from environment variables
 and the project's `.env` file if present.
 
+If unable to create load a config from environment variables, the function will
```
aiwilliams, line 21:
```suggestion
If unable to load a config from environment variables, the function will
```

charlieduong94: Ah nice catch. Fixed.

PR #58, opened 2020-04-21
add polly config optios to setupRecording

ndowmon:
Some of my integration tests include `404` responses from `user` endpoints, and users are instead recorded with abridged data (name, email, userId).

I would like to pass all of the `PollyOptions` through `setupRecording` so that developers can use options like `recordFailedRequests: true` while creating integrations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,15 @@ and this project adheres to
 
 ## [Unreleased]
 
-## 0.6.1 - 2020-04-21
-
 ### Fixed
 
 - Fixed issue with dependency graph failing to wait for step dependencies to
   complete prior to performing more work
+
+### Added
+
+- `setupRecording` takes an optional parameter `options` which allows developers
+  to pass in parameters found in polly's `PollyConfig` interface.
 
 ## 0.6.0 - 2020-04-21
 
--- a/src/testing/__tests__/recording.test.ts
+++ b/src/testing/__tests__/recording.test.ts
@@ -16,15 +16,12 @@ jest.mock('fs');
 let recording: Recording;
 let server;
 
-beforeAll(async () => {
+beforeEach(async () => {
   server = await startServer();
 });
 
-afterAll(async () => {
+afterEach(async () => {
   await server.close();
-});
-
-afterEach(() => {
   recording.stop();
   vol.reset();
 });
@@ -78,9 +75,6 @@ test('accepts recordFailedRequests PollyConfig option', async () => {
       toUnixPath(path.resolve(__dirname, '__recordings__', name)),
     ),
   ).toEqual(true);
-
-  server.close();
-  server = await startServer();
 });
 
 test('redacts cookies from responses', async () => {
--- a/src/testing/__tests__/recording.test.ts
+++ b/src/testing/__tests__/recording.test.ts
@@ -181,6 +181,28 @@ test('allows for entries to be mutated via mutateEntry function', async () => {
   expect(har.log.entries[0].request.headers).toEqual([]);
 });
 
+test('allows for overriding matchRequestBy options with deepDefault', async () => {
+  recording = setupRecording({
+    name: 'test',
+    directory: __dirname,
+    options: {
+      matchRequestsBy: {
+        order: false,
+        url: {
+          query: false,
+        },
+      },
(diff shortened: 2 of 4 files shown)


ndowmon:
@charlieduong94 I added a test and pushed a new tag (`v0.5.1`)

aiwilliams:
> @charlieduong94 I added a test and pushed a new tag (`v0.5.1`)

Travis will publish to NPM whenever a tag is created, no matter which branch the referenced commit resides on, so 0.5.1 was already published to NPM 😬 What we have to do is wait until a PR(s) is merged, then version and push to master.

src/testing/__tests__/recording.test.ts
```diff
@@ -51,6 +51,38 @@ test('create files relative to the recording directory based on the input name',
...
+  const [recordingPath] = Object.keys(vol.toJSON());
+
+  expect(
+    recordingPath.startsWith(
+      toUnixPath(path.resolve(__dirname, '__recordings__', name)),
+    ),
+  ).toEqual(true);
+
+  server.close();
+  server = await startServer();
```
aiwilliams, line 83: This will fail to reset `server` if the `expect` fails I think.

charlieduong94:
Yeah, we can't follow the usual flow we have for internal projects here unfortunately. Perhaps we could build a github action sometime in the future.

charlieduong94:
Oh if you can update the `CHANGELOG.md` file describing the change and some documenting in `docs/testing.md` that would be great.

src/testing/__tests__/recording.test.ts
ndowmon: I feel like it'd be good to just restart the server in `beforeEach` and `afterEach` now that it's being configured differently in different tests... unless @charlieduong94 would prefer to leave them in `beforeAll`/`afterAll`.

charlieduong94: Yep, using `beforeEach`/`afterEach` would be better now.

ndowmon:
Shoot I did know about the github tagging, but it slipped my mind before the commit. My bad @aiwilliams. 

New commit is up with the `CHANGELOG.md` updated for `v0.6.1` and also a small change to `docs/testing.md`. I haven't made the `v0.6.1` tag yet, though.

src/testing/recording.ts
```diff
@@ -75,6 +77,7 @@ export function setupRecording({
   }
 
   return new Polly(name, {
+    ...options,
```
austin-rausch, line 80: How would you feel about doing using [this](https://lodash.com/docs/4.17.15#defaultsDeep) lodash function and allowing deep properties to be specified? I have a need to overwrite the `matchRequestsBy` values and will probably end up implementing it anyways.

CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ and this project adheres to
 
 ## [Unreleased]
 
+## 0.6.1 - 2020-04-21
```
charlieduong94, line 11: Considering this is technically a new feature in the sdk, it would be more appropriate to do a minor version bump to (0.7.0) instead of a patch.

I tend to just add the version to the change log prior to adding the publish commit to master since there are sometimes a couple of changes making their way into the project all at once.

src/testing/recording.ts
ndowmon: We talked offline but for the sake of documenting, I'm working on this implementation.

CHANGELOG.md
charlieduong94: Oh and fyi, I'm stealing 0.6.1 to get a hotfix out

ndowmon: The minor version bump makes sense, I was a bit hesitant since it's much smaller than the other minor versions. 
Not sure if you're suggesting just adding changes + additions under the [Unreleased] header until a new version is pushed. If so, I agree.

charlieduong94: >Not sure if you're suggesting just adding changes + additions under the [Unreleased] header until a new version is pushed. If so, I agree.

Yeah, that's what I'm suggesting.

ndowmon:
@charlieduong94 @austin-rausch added Rausch's request + testing and its usage in `testing.md` docs.

PR #59, opened 2020-04-21
NO-TICKET - added -s or --step option to the collect command.

austin-rausch:


austin-rausch:
I can add more tests around the defaulting behavior / original get start state function overriding, but I didn't think it was necessary.

charlieduong94:
> The docs say that --step step-fetch-users --step step-fetch-groups will work as well, does that come free with this? Perhaps we need a test to prove that sticks around?

A test like @aiwilliams mentioned here would be nice. I think it should come for free with `commander` but I could be wrong.

austin-rausch:
> > The docs say that --step step-fetch-users --step step-fetch-groups will work as well, does that come free with this? Perhaps we need a test to prove that sticks around?
> 
> A test like @aiwilliams mentioned here would be nice. I think it should come for free with `commander` but I could be wrong.

Yea, so I didn't spend a lot of time on it, but when providing multiple `--step`s the latest would overwrite the previous ones, maybe something in the `.option()` isn't right.

austin-rausch:
Still TODO: resolve dependent steps and enable those as well.

CHANGELOG.md
```diff
@@ -8,14 +8,6 @@ and this project adheres to
...
-
-- Added `--step / -s` to `collect` command for the CLI, takes either a single
-  step or comma separated steps or multiple values. E.G. `--step fetch-account`,
-  `--step fetch-account,fetch-users`, or `--step fetch-account --step fetch-users`.
-
-## 0.6.1 - 2020-04-21
-
 ### Fixed
 
 - Fixed issue with dependency graph failing to wait for step dependencies to
```
ndowmon: @austin-rausch I think I messed this up in my branch. this fix is in 0.6.1, but the `setupRecording` addition is not 🤦 
Sorry for the confusion

austin-rausch: Yea, I just fixed it.  I think it was me resolving the conflict wrong.

charlieduong94:
:kissing_cat:

PR #69, opened 2020-04-23
Setup publishing from workflow. Remove travis.yml

charlieduong94:
Got bored and decided to take a stab at replicating our internal publishing flow with workflows. I tried things out with a test project and got things to work well (see [this build](https://github.com/charlieduong94/workflow-test/actions/runs/85367102)).

I checked out some off the shelf solutions, but it seemed like all of them required passing in the workflow's `GITHUB_TOKEN` secret for writing tags for you. I wasn't super comfortable with that.


.github/workflows/build.yml
```diff
@@ -25,3 +25,45 @@ jobs:
...
+            echo "Found version commit tag. Publishing."
+            echo "::set-env name=publish::true"
+          else
+            echo "Version commit tag not found. Not publishing."
+          fi
+
+      - name: Publish
+        if: env.publish == 'true'
+        env:
+          NPM_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
```
charlieduong94, line 65: If we decide to use this, we'll need add a secret cc @aiwilliams

aiwilliams: > Secrets are environment variables that are encrypted and only exposed to selected actions. Anyone with collaborator access to this repository can use these secrets in a workflow.

I suppose we shouldn't add collaborators we don't trust to build workflows 😬 

Okay, secret is installed.

aiwilliams: Oh, note to everyone: we'll need to add the secret to each repository that does this, just as we had to do with Travis.

PR #78, opened 2020-04-24
NO-TICKET - add mutateRequest function to change the request object preflight.

austin-rausch:
The root of the issue is that a post body sent that is gunzipped won't get persisted correctly to the `.har` files because of this:
https://github.com/Netflix/pollyjs/blob/master/packages/%40pollyjs/persister/src/har/request.js#L42

The value of the `request.body` can't be changed post flight, because the object it inherits from is frozen once the request finishes so that the request can be uniquely identified.  Therefore I added in a function to change the value preflight...
https://github.com/Netflix/pollyjs/blob/master/packages/%40pollyjs/core/src/-private/http-base.js#L103

Example usage of this:

```
mutateRequest: (request) => {
    const contentEncoding = request.getHeader('content-encoding');
    if (contentEncoding === 'gzip') {
      const unzippedBody = zlib.unzipSync(request.body);
      request.body = unzippedBody.toString();
      request.setHeader('content-length', unzippedBody.byteLength);
      request.removeHeader('content-encoding');
    }
}
```

### DIFF
--- a/src/testing/recording.ts
+++ b/src/testing/recording.ts
@@ -16,14 +16,35 @@ interface SetupRecordingInput {
   redactedRequestHeaders?: string[];
   redactedResponseHeaders?: string[];
   mutateEntry?: (entry: any) => void;
+  mutateRequest?: (request: any) => void;
   options?: PollyConfig;
 }
 
 /**
- * Sets up a recording of all http requests and
+ * @description Sets up a recording of all http requests and
  * writes the data to disk when it is stopped.
- *
  * This leverages Polly.js to do all the heavy lifting.
+ * @param input.mutateRequest - allows mutating the `PollyRequest` object
+ * pre flight, this affects how the request is sent out and also how it
+ * is stored.  This is currently the only known way to get compressed (gzip, broli, etc.)
+ * requests to persist their body to the har file.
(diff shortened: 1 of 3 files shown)


src/testing/recording.ts
```diff
@@ -16,21 +16,43 @@ interface SetupRecordingInput {
   redactedRequestHeaders?: string[];
   redactedResponseHeaders?: string[];
   mutateEntry?: (entry: any) => void;
+  mutateRequest?: (request: any) => void;
   options?: PollyConfig;
 }
 
 /**
- * Sets up a recording of all http requests and
+ * @description Sets up a recording of all http requests and
```
aiwilliams, line 24: Nice, love the improvements to docs by adding these annotations!

austin-rausch:
`mutateEntity` doesn't always mutate the entity as well, it just gives you the ability to.  `onRequest` is a bad name anyways, it seems more like its registering an event handler. Also already merged...

aiwilliams:
FWIW, it does seem like `onRequest` would be a better fit. In a programming language that passes references and allows mutation all the time, and in a case where the function is not returning anything, `mutateRequest` sounds too assertive about the intended purpose, whereas `onRequest` sounds like an opportunity to do as though wilt.

Issue #65, opened 2020-04-22
Automatically redact known sensitive headers when recording


aiwilliams:
`Authorization` should always be redacted, there may be others.

austin-rausch:
Leaving this open as I think a broader discussion is in order.

Discussion should include:
- List of common header names that are sensitive (aside from Authorization of course)
- Possibly regex testing the `value` of each header against well known secret patterns similar to how we do in the aws integration.

PR #83, opened 2020-04-24
#65 - added redacting of authorization header for requests.

austin-rausch:


src/testing/recording.ts
```diff
@@ -20,6 +20,8 @@ interface SetupRecordingInput {
   options?: PollyConfig;
 }
 
+const SENSITIVE_HEADER_NAMES = ['authorization'].map((i) => i.toLowerCase());
```
aiwilliams, line 23: Perhaps there's no need to normalize the constant `SENSITIVE_HEADER_NAMES` at all, since we'll only put lowercase values in it?
```suggestion
const SENSITIVE_HEADER_NAMES = ['authorization'];
```

austin-rausch: Probably not, but I added it incase someone screws up.

PR #85, opened 2020-04-24
Allowing source objects to have null/undefined properties

softwarewright:
@aiwilliams and I noticed that when properties that line up with https://github.com/JupiterOne/data-model/tree/master/src/schemas are null they cause an exception to be thrown when calling createIntegrationEntity. This fix removes those properties from the whitelist of properties to be validated against the schema, but still allows those properties to be written to the entity created.

### DIFF
--- a/src/framework/data/createIntegrationEntity.ts
+++ b/src/framework/data/createIntegrationEntity.ts
@@ -202,7 +202,8 @@ function whitelistedProviderData(
   const whitelistedProviderData: ProviderSourceData = {};
   const schemaProperties = schemaWhitelistedPropertyNames(_class);
   for (const e of Object.entries(source)) {
-    if (schemaProperties.includes(e[0])) {
+    const hasValue = e[1] !== null && e[1] !== undefined;
+    if (hasValue && schemaProperties.includes(e[0])) {
       whitelistedProviderData[e[0]] = e[1];
     }
   }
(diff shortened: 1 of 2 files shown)


src/framework/data/createIntegrationEntity.ts
```diff
@@ -202,7 +202,8 @@ function whitelistedProviderData(
   const whitelistedProviderData: ProviderSourceData = {};
   const schemaProperties = schemaWhitelistedPropertyNames(_class);
   for (const e of Object.entries(source)) {
-    if (schemaProperties.includes(e[0])) {
+    const hasValue = e[1] !== null && e[1] !== undefined;
```
philidem, line 205: Slightly nitpicky, but I would recommend destructuring to make this more readable:

For example:

```typescript
const [key,value] of Object.entries(source)
```

philidem: Simplification:

```typescript
const hasValue = value != null;
```

`!= null` is the one exception that you can make for using that operator because it will check for `null` and `undefined`.

softwarewright: Good point, that does look better.

src/framework/data/createIntegrationEntity.ts
```diff
@@ -201,10 +201,9 @@ function whitelistedProviderData(
 ): Omit<ProviderSourceData, 'tags'> {
   const whitelistedProviderData: ProviderSourceData = {};
   const schemaProperties = schemaWhitelistedPropertyNames(_class);
-  for (const e of Object.entries(source)) {
-    const hasValue = e[1] !== null && e[1] !== undefined;
-    if (hasValue && schemaProperties.includes(e[0])) {
-      whitelistedProviderData[e[0]] = e[1];
+  for (const [key, value] of Object.entries(source)) {
+    if (value != null && schemaProperties.includes(key)) {
+      whitelistedProviderData[key] = value;
```
aiwilliams, line 206: Yah, much nicer!

PR #88, opened 2020-04-24
add gitleaks toml file to root and template.

austin-rausch:


austin-rausch:
> Weird name, "toml".

yea, some [asshole](https://en.wikipedia.org/wiki/Tom_Preston-Werner) came up with it and named it after himself.

PR #96, opened 2020-04-27
Fix issue with prettier disable not working correctly

charlieduong94:
I upgraded my setup to actually work with `eslint` correctly and started to feel the pain others have. Turns out I got the eslint prettier plugin and config mixed up. The config disables eslint styling rules and the plugin turns them on. I don't think we use any eslint configs that make styling choices, but it's probably better to ensure that all of them are off and do not conflict with prettier just to be safe.

Reference: https://prettier.io/docs/en/integrating-with-linters.html#disable-formatting-rules

Shoutout to @ndowmon for figuring out that removing the plugin got rid of formatting rules.

### DIFF
--- a/template/.eslintrc
+++ b/template/.eslintrc
@@ -3,7 +3,7 @@
     "eslint:recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:jest/recommended",
-    "plugin:prettier/recommended"
+    "prettier"
   ],
   "env": {
     "node": true,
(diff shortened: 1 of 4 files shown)


template/.eslintrc
```diff
@@ -3,7 +3,7 @@
     "eslint:recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:jest/recommended",
-    "plugin:prettier/recommended"
```
ndowmon, line 6: Out of curiosity, why use an `.eslintrc` file in the template, as opposed to putting it in the `package.json` like the sdk?

charlieduong94: I got feedback that others prefer the individual config files vs the `package.json` approach. Someday in the future I think we are going to shift the sdk to look a little more like the template projects.

PR #99, opened 2020-04-27
Add synchronization code

charlieduong94:
This adds the "meat" of the code that will perform synchronization that will be used by the `sync` and `run` commands. In managed environments, my goal here is to use the same code but pass in an `ApiClient` that is configured slightly differently.

I'm sorry the PR is a tad big but all of the code here is just scoped around performing synchronization so I'm hoping it won't be too hard to review. The `sync` and `run` commands are coming in another PR. I'll be circling back to add retry logic in a later PR as well.


charlieduong94:
hmm one sec. Going to add splitting up of uploaded payloads rnrq

src/framework/api/__tests__/index.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    expect(getApiBaseUrl({ dev: false })).toEqual(
+      'https://api.us.jupiterone.io',
+    );
+  });
+
+  test('defaults to returning the production base url', () => {
+    expect(getApiBaseUrl({ dev: false })).toEqual(
+      'https://api.us.jupiterone.io',
+    );
+  });
```
aiwilliams, line 34: I think these tests are the same?

src/framework/api/index.ts
```diff
@@ -0,0 +1,108 @@
...
+  account: string;
+}
+
+interface CreateApiClientFromApiKeyInput {
+  apiBaseUrl: string;
+  apiKey: string;
+}
+
+/**
+ * Creates an account using an API key
```
aiwilliams, line 25:
```suggestion
 * Creates an API client using an API key
```

src/framework/execution/logger.ts
```diff
@@ -12,7 +12,7 @@ const bunyanFormat = require('bunyan-format');
 
 interface CreateIntegrationLoggerInput {
   name: string;
-  invocationConfig: IntegrationInvocationConfig;
+  invocationConfig?: IntegrationInvocationConfig;
```
aiwilliams, line 15: Nothing worth changing right now, but I think it may be good to provide a reference to only the `instanceConfigFields` instead of the `invocationConfig`, if the only thing in the config used is the `instanceConfigFields`. This can communicate better what is necessary for the function, and makes tests simpler so that a complete config doesn't have to be built wherever this function is used.

src/framework/api/__tests__/index.test.ts
charlieduong94: oof will fix

src/framework/execution/logger.ts
charlieduong94: Yeah, that would be a good improvement.

charlieduong94:
I haven't thought about exposing the `scope` synchronization, but I could see it def being useful for quick uploads and for allowing developers to test out synchronization without having to provision an integration instance. 

I almost wonder if forcing developers to provision an integration instance (at least an "external" integration instance in the future) would reduce the chance of users mistakenly flooding their account with duplicate entities if they accidentally change the scope between runs. A scope can pretty much be any string value and can easily be fat fingered when input from the CLI, but an integration instance id is much harder to mess up since ids are unique and the sdk would error if the id does not match an instance.

I'm down for allowing for supporting the scope use case and it would be super easy to do so I'm down to add it, but perhaps we may want to document it's usage with a "use with caution" warning.

aiwilliams:
> I almost wonder if forcing developers to provision an integration instance (at least an "external" integration instance in the future) would reduce the chance of users mistakenly flooding their account with duplicate entities if they accidentally change the scope between runs.

My understanding of @philidem vision on the scope was, it would be stored on the data so that someone could also just bulk delete the data as they would when removing an integration instance. However, were we to have external integration instances - and even API tokens bound to the instance - that would make the scope thing not very useful I think.

PR #112, opened 2020-04-30
Project cleanup

philidem:
These PR contains various project cleanup items. Please look at individual commits for more details.


philidem:
**For convenience, here are the changes:**

- Revise errors so that there is a base IntegrationError class with support for message, code, and cause
- Revise code after tsconfig changes caused tsc to report additional errors
- Revise .eslintrc config so that eslint doesn't report errors in IDE when code does not match prettier style
- Modify tsconfig to make it more strict (NOTE: I think we should enable noImplicitAny later)
- Upgrade dev dependencies
- Move jest, eslint, prettier, husky, and lint-staged config out of package.json into top-level files

PR #113, opened 2020-04-30
Use IntegrationError instead of Error. Also, use UPPER_CASE for error codes

philidem:
I switched to using `IntegrationError` instead of `Error`.

Also, we were already using UPPER_CASE for error codes in synchronization API and other services so I switched to using UPPER_CASE instead of TitleCase for error codes.


philidem:
> Thoughts on using an `enum` for the integration error codes? Sometimes it's nice when projects that throw custom error codes have a single list that you can see all of the error codes. We could also put some code comment docs in with each value:
> 
> ```
> export enum IntegrationErrorCode {
>   /**
>   * ...stuff..
>   */
>   MALFORMED_API_KEY_ERROR = 'MALFORMED_API_KEY_ERROR',
> }
> ```

Hmm. I'm not sure if that would really help that much. I imagine we'd end up with a lot of error codes that get removed but still stick around in the `enum`.

src/framework/data/createIntegrationRelationship.ts
```diff
@@ -252,9 +256,10 @@ export function generateRelationshipKey(
   to: { _key: string } | string,
 ): string {
   if (!from || !to) {
-    throw new Error(
-      '"from" and "to" must be provided to generate a relationship _key!',
-    );
+    throw new IntegrationError({
+      code: 'GENERATE_RELATIONSHIP_KEY_MISSING_RELATIONSHIP_FROM_OR_TO',
+      message: '"from" and "to" must be provided to generate a relationship _type!',
+    });
```
aiwilliams, line 262: I've preferred something along the lines of `throw new IntegrationRelationshipTypeGenerationError(from, to)`, and putting the code and message generation in the error constructor to avoid duplication and to construct a message with more detail, to avoid "this could be the problem or that could be the problem". I see other errors doing some of this, such as `IntegrationLocalConfigFieldMissingError`, by putting the `code` in the constructor 👍

aiwilliams:
> I imagine we'd end up with a lot of error codes that get removed but still stick around in the `enum`.

I wonder, if the enum is defined at the top of the errors definition file, and the error constructor sets the code, it would be less likely? Since we cannot depend on `instanceof`, it would be nice where `code` is checked to have it compare to a documented description of the error.

PR #114, opened 2020-05-01
Add "yarn autobuild" command and adjust packaging to make possible

philidem:
- Added `dotenv-expand` which allows `${...}` variables in `.env` file
- Changed how package is created so that it was easier to support `yarn link`
- Added support for `yarn autobuild` command

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,3 +22,18 @@ details about how to develop integrations with this SDK.
 
 An example template project can be found
 [here](https://github.com/JupiterOne/integration-sdk/tree/master/template).
+
+## Integration SDK Development
+
+If you are making changes to the integration SDK and you want to test the
+changes in another integration project then it is recommended to automatically
+rebuild and link this project when changes are made.
+
+Steps to automatically build and link:
+
+- Run `yarn autobuild` in _this_ project from a terminal and wait for initial
+  build to complete.
+
+- In a separate terminal, run `yarn link @jupiterone/integration-sdk` in the
+  integration project. You can now use the integration SDK CLI in the other
+  project and it will use the latest code on your filesystem.
--- a/package.json
+++ b/package.json
@@ -2,13 +2,13 @@
   "name": "@jupiterone/integration-sdk",
   "version": "0.15.0",
   "description": "The SDK for developing JupiterOne integrations",
-  "main": "dist/framework/index.js",
-  "types": "dist/framework/index.d.ts",
+  "main": "src/framework/index.js",
+  "types": "src/framework/index.d.ts",
   "repository": "git@github.com:JupiterOne/integration-sdk.git",
   "author": "JupiterOne <dev@jupiterone.io>",
   "license": "MPL-2.0",
   "engines": {
-    "node": "10.x || 12.x"
+    "node": "10.x || 12.x || 14.x"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -2,35 +2,30 @@
...
-  "main": "dist/framework/index.js",
-  "types": "dist/framework/index.d.ts",
+  "main": "src/framework/index.js",
+  "types": "src/framework/index.d.ts",
   "repository": "git@github.com:JupiterOne/integration-sdk.git",
   "author": "JupiterOne <dev@jupiterone.io>",
   "license": "MPL-2.0",
   "engines": {
-    "node": "10.x || 12.x"
+    "node": "10.x || 12.x || 14.x"
```
aiwilliams, line 11: Seems the build matrix needs to cover this then, eh?

package.json
```diff
@@ -2,35 +2,30 @@
...
     "format": "prettier --write '**/*.{ts,js,json,css,md,yml}'",
     "type-check": "tsc --noEmit",
     "test": "jest",
     "test:ci": "yarn lint && yarn type-check && yarn test --ci",
-    "prebuild": "yarn test:ci",
-    "build": "tsc -p tsconfig.dist.json --declaration",
-    "prepublishOnly": "yarn build"
+    "compile": "tsc -p tsconfig.dist.json --declaration",
+    "prebuild": "rm -rf dist",
+    "build": "yarn compile && cp -R bin README.md LICENSE yarn.lock package.json ./dist",
```
aiwilliams, line 27: I rather miss the notion of using `files` to specify what goes into the built package. Seems so clean and native.

I think `yarn.lock` is not necessary or useful in the npm package, is it?

README.md
```diff
@@ -22,3 +22,18 @@ details about how to develop integrations with this SDK.
 
 An example template project can be found
 [here](https://github.com/JupiterOne/integration-sdk/tree/master/template).
+
+## Integration SDK Development
+
+If you are making changes to the integration SDK and you want to test the
+changes in another integration project then it is recommended to automatically
+rebuild and link this project when changes are made.
```
aiwilliams, line 30: I've wondered whether this workflow is supported by creating a Yarn workspace, that isn't checked in anywhere, but I've not messed with it enough to say for sure.

package.json
philidem: Well... I use it to do a production install when linking so it helps make sure the versions that were specified work properly. It's probably not strictly necessary tho.

package.json
philidem: Personally, I don't think we need to even specify the engine... I had just happen to notice that v14 of node was released recently so I didn't think we should block that from working.

package.json
philidem: Regarding `files`, I didn't think it was necessary because we're already packaging exactly what we need into the `dist` folder. I feel like the `files` entry will just add confusion if something is not packaged up as expected because they forgot to add entry to `files`.

aiwilliams: Yup, it's undesirable to specify `files` in this case for sure. I just don't fully understand why that mechanism is not used instead of this approach, which I assume has to do with `autobuild`, which I have never used.

package.json
aiwilliams: I wonder if this could be improved with something like ">10", since probably we just mean to say we aren't supporting known incompatibilities (if there are any).

README.md
charlieduong94: I think it definitely can be supported via workspaces. But I suppose for others not using workspaces, the `autobuild` command could be helpful.

PR #120, opened 2020-05-01
Fix template, replace test run in pre-push

aiwilliams:
The `yarn build` used to run the tests in `prebuild`, but that was lost with `autobuild` changes in last release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Fixed
+
+- Template missing necessary publish build.
```
austin-rausch, line 13: Minor, this doesn't really read like an action performed, it reads more like a statement or fact.

PR #121, opened 2020-05-01
Add IntegrationConfigLoadError type

aiwilliams:
This will be used by the managed integration environment to complain when it cannot load a config.

### DIFF
--- a/src/framework/execution/error.ts
+++ b/src/framework/execution/error.ts
@@ -17,11 +17,19 @@ export class IntegrationLocalConfigFieldTypeMismatch extends IntegrationError {
     });
   }
 }
+export class IntegrationConfigLoadError extends IntegrationError {
+  constructor(message: string) {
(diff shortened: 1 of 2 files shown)


src/framework/execution/error.ts
```diff
@@ -17,11 +17,19 @@ export class IntegrationLocalConfigFieldTypeMismatch extends IntegrationError {
...
+      message,
+    });
+  }
+}
 
 export class IntegrationConfigValidationError extends IntegrationError {
   constructor(message: string) {
     super({
-      code: 'LOCAL_CONFIG_FIELD_TYPE_MISMATCH',
+      code: 'CONFIG_FIELD_TYPE_MISMATCH',
```
aiwilliams, line 32: Local is not in the name of the class, so to match the pattern of other code names, dropped it from this.

PR #134, opened 2020-05-05
Add step event publishing functions to logger

charlieduong94:
As per, what we have in the [development document](https://github.com/JupiterOne/integration-sdk/blob/master/docs/development.md#logger) around logging, this felt like a good time to experiment with instrumenting the logger with new functions that perform both event publishing and traditional logging.

### DIFF
--- a/src/framework/execution/logger.ts
+++ b/src/framework/execution/logger.ts
@@ -1,11 +1,15 @@
 import Logger from 'bunyan';
+import PromiseQueue from 'p-queue';
 
 import {
   IntegrationLogger,
+  IntegrationStep,
   IntegrationInstance,
   IntegrationInvocationConfig,
   IntegrationInstanceConfigFieldMap,
 } from './types';
+
+import { SynchronizationJobContext } from '../synchronization';
 
 // eslint-disable-next-line
 const bunyanFormat = require('bunyan-format');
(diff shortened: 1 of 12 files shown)


src/framework/execution/logger.ts
```diff
@@ -120,3 +134,84 @@ function instrumentVerboseTrace(logger: Logger): Logger {
...
+    stepSuccess: (step: IntegrationStep) => {
+      const name = 'step-end';
+      const description = `Completed step "${step.name}".`;
+      logger.info({ step: step.id }, description);
+
+      publishEvent(name, description);
+    },
+    stepFailure: (step: IntegrationStep, err: Error) => {
+      const name = 'step-failure';
+      const description = `Step "${step.name}" failed to complete due to error.`;
```
aiwilliams, line 203: I think we'll soon need an error code that is displayed to users in the event, so that they can provide that to us that we are able to quickly locate the context of the error in the logs. Shall I create a separate Issue to discuss/track that feature?

charlieduong94: > Shall I create a separate Issue to discuss/track that feature?

Yes please! 

On a related note, I think we should start logging out the integration job id at the start of runs so that users that experience some kind of unexpected issue can also pass that along to us.

aiwilliams: Great idea, I'll add both. Thanks!

PR #140, opened 2020-05-05
Reject and stop execution if duplicate key is detected

charlieduong94:


src/framework/execution/jobState.ts
```diff
@@ -1,16 +1,45 @@
 import { JobState, IntegrationStep } from './types';
 
 import { FileSystemGraphObjectStore } from '../storage';
+import { IntegrationDuplicateKeyError } from './error';
+import { Entity, Relationship } from '../types';
+
+export class DuplicateKeyTracker {
+  keySet = new Set<string>();
```
austinkelleher, line 8: Consider making this a `private readonly` property to prevent accidental direct access. We should always force usage of `registerKey`

charlieduong94: Good idea, will change

ndowmon:
What happens if an entity with a duplicate key has been persisted in the graph in a previous invocation? e.g.
1. Invocation 1 returned & persisted an entity with `_class: Class1` and `_key: a`
2. Later, Invocation 2 returns an entity with `_class: Class2` and `_key: a`. Invocation 2 does _not_ return an entity with `_class: Class1` and `_key: a`. 

It looks like that's handled as described in the [synchronization docs](https://bitbucket.org/lifeomic/jupiter-persister/src/master/docs/synchronization.md) at **Diffing Algorithm/Process Old State**, but I just want to make sure we can't persist these duplicates.

charlieduong94:
The synchronizer will not persist anything if it detects a duplicate key.

> 1. Invocation 1 returned & persisted an entity with `_class: Class1` and `_key: a`
> 2. Later, Invocation 2 returns an entity with `_class: Class2` and `_key: a`. Invocation 2 does not return an entity with `_class: Class1` and `_key: a`.

In the situation you are describing here, since invocation 2 is not sending a duplicate key, the synchronizer just treats the new information (`_class` changing from `Class1` to `Class2`) as an update for the entity that matches `_key: a`

PR #143, opened 2020-05-06
Report errorCode and errorId upon step failure event

charlieduong94:


src/framework/execution/__tests__/logger.test.ts
```diff
@@ -294,7 +300,9 @@ describe('step event publishing', () => {
       events: [
         {
           name: 'step-failure',
-          description: 'Step "Mochi" failed to complete due to error.',
+          description: expect.stringContaining(
+            `Step "Mochi" failed to complete due to error. (errorCode=${error.code}`,
```
aiwilliams, line 304: Oh, just noticed there is no assertion that `errorId=` is included in the `description` text.

charlieduong94: I'll add that rnrq

PR #144, opened 2020-05-06
Log validation error if validateInvocation function throws

charlieduong94:
To fulfill the need of providing more meaningful messages to end users, I've added the ability to allow for an error's message to be added to the published log event when it's `expose` flag is set to true.


docs/development.md
```diff
@@ -126,6 +126,47 @@ ensuring the integration has received a valid set of `instanceConfigFields`. It
...
+    throw new IntegrationValidationError(
+      'Config requires a clientId and clientSecret to be provided',
+    );
+  }
+
+  return validate(config);
+}
+```
+
+Using the abover examplle, the following message will be logged if the config's
```
austin-rausch, line 155:
```suggestion
Using the above example, the following message will be logged if the config's
```

src/framework/execution/__tests__/logger.test.ts
```diff
@@ -310,3 +317,89 @@ describe('step event publishing', () => {
...
+    const { description, errorId } = createErrorEventDescription(
+      errorWithExposeTrue,
+      'testing',
+    );
+    expect(description).toEqual(
+      `testing (errorCode=${UNEXPECTED_ERROR_CODE}, errorId=${errorId}, reason=soba)`,
+    );
+  });
+
+  test('does not add reason if expose is not set', () => {
```
austin-rausch, line 383: Why tho?  Because the stack / message could be really large?

docs/development.md
```diff
@@ -126,6 +126,47 @@ ensuring the integration has received a valid set of `instanceConfigFields`. It
...
+function validateInvocation(context: IntegrationExecutionContext) {
+  const { config } = context.instance;
+
+  if (!config.clientId || !config.clientSecret) {
+    throw new IntegrationValidationError(
+      'Config requires a clientId and clientSecret to be provided',
+    );
+  }
+
+  return validate(config);
```
aiwilliams, line 151: I think this should not `return`, as it implies something will be done with the value.

docs/development.md
```diff
@@ -126,6 +126,47 @@ ensuring the integration has received a valid set of `instanceConfigFields`. It
...
+
+`Error occurred while validating integration configuration. (errorCode=CONFIG_VALIDATION_ERROR, errorId=<generated error id>, reason=Config requires a clientId and clientSecret to be provided)`;
+
+Optionally, users can pass in a more specific error code to display when
+throwing the error.
+
+```typescript
+throw new IntegrationValidationError(
+  'Config requires a clientId and clientSecret to be provided',
+  'INVALID_CONFIGURATION',
```
aiwilliams, line 166: I wonder how someone will know when to take this option, and what value to provide that would communicate more than can be in the message? I think I would drop this option, and then when we look for `errorCode` in the logs, we'll have a consistent relationship.

src/errors.ts
```diff
@@ -23,10 +24,17 @@ export class IntegrationError extends Error {
    */
   readonly fatal: boolean | undefined;
 
+  /**
+   * An option flag that will allow for error messages
+   * to be logged in the integration event log if set to true
```
aiwilliams, line 29:
```suggestion
   * The `errorCode` and `errorId` will always be visible to users.
   * Setting `expose: true` will allow for error messages
   * to be included in the integration event log.
```

src/framework/execution/__tests__/logger.test.ts
charlieduong94: I'm not really worried about the message length too much and we don't display the stack in published logs.

I thought about allowing all error messages to be shown, but settled on only allowing for certain errors to be exposed to users. I didn't want some silly javascript/typescript error to end up getting surfaced. Surfacing errors like that make us look pretty bad and is often not useful for end users.

docs/development.md
charlieduong94: will fix now otherwise I'll forget

src/framework/execution/__tests__/logger.test.ts
aiwilliams: I think this is a good question. Perhaps looking at older integrations we could find cases where we don't expose the message, but I wonder if we could always just expose it. I wouldn't expect the `message` to be multi-line.

src/framework/execution/executeIntegration.ts
```diff
@@ -69,7 +69,12 @@ async function executeIntegration(
 ): Promise<ExecuteIntegrationResult> {
   await removeStorageDirectory();
 
-  await config.validateInvocation?.(context);
+  try {
+    await config.validateInvocation?.(context);
+  } catch (err) {
+    context.logger.validationFailure(err);
+    throw err;
```
aiwilliams, line 76: Will this be caught and logged again? My hope is that it will not. In the production environment, it can be confusing to see the same thing logged multiple times in different formats (the Bunyan log output from `context.logger.validationFailure(err)` and the Lambda runtime logging of the same, uncaught error.

src/framework/execution/__tests__/logger.test.ts
aiwilliams: I think that we want to catch all errors and for those which are unexpected, the message would be something like https://bitbucket.org/lifeomic/jupiter-managed-integration-sdk/src/c53e0f7e0403b636adcf004798b0a49c8b9d8753/src/integration/types/errors.ts#lines-41. That is, I agree, we want to control what is shown for sure, but it seems there is always a message we can show.

charlieduong94: I suppose we can just replace the logged message with a canned one if we find that the error throw is not an instance of `IntegrationError`

docs/development.md
charlieduong94: will fix

src/framework/execution/__tests__/logger.test.ts
austin-rausch: > I'm not really worried about the message length too much and we don't display the stack in published logs.
> 
> I thought about allowing all error messages to be shown, but settled on only allowing for certain errors to be exposed to users. I didn't want some silly javascript/typescript error to end up getting surfaced. Surfacing errors like that make us look pretty bad and is often not useful for end users.

Okay yea.  We'll still receive the error in logs and something like `Type error: cannot read property "blahablahblah" of null` looks bad.

docs/development.md
charlieduong94: I was thinking that validation error codes would be defined on a per integration basis to provide some more useful feedback, but after coming back to this a day later, it doesn't make much sense since the message does that job. Will remove.

src/framework/execution/executeIntegration.ts
charlieduong94: Yeah, I can see multiple errors getting confusing and adding noise (our monitor has the same issue atm). I'm thinking we mark the errors that have been already logged and perform a check at the "catch-all" error handler to ensure duplicate logs are not written.

charlieduong94: There needs to be a way to handle gracefully handle execution failure and abort the sync/integration job anyway so I think I will handle this in the next pr.

PR #145, opened 2020-05-07
Allow schema validation to be disabled in managed environments

charlieduong94:
This takes a similar approach to what the managed integration sdk did.


src/cli/__tests__/cli-run.test.ts
```diff
@@ -44,9 +44,28 @@ beforeEach(() => {
 
 afterEach(async () => {
   delete process.env.JUPITERONE_API_KEY;
+  delete process.env.ENABLE_GRAPH_OBJECT_SCHEMA_VALIDATION;
   await polly.disconnect();
 });
 
+test('enables graph object schema validation', async () => {
```
aiwilliams, line 51: Excellent.

PR #146, opened 2020-05-08
Abort synchronization job on failure

charlieduong94:
I setup the logger to track previously logged errors so that we can avoid duplicate logging at the final `catch`. Mutating the error as it bubbled up felt a little dirty, so I set up tracking via a `Set`. I don't feel too strongly about any approach though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to
 - An `errorCode`, `errorId`, and `reason` is now displayed in step failure
   events and logs.
 - Event logging around errors thrown in `validateInvocation` function.
+- Abort synchronization job when `run` and `sync` command fails.
 
 ### Changed
 
--- a/src/cli/commands/run.ts
+++ b/src/cli/commands/run.ts
@@ -12,6 +12,7 @@ import {
   initiateSynchronization,
   uploadCollectedData,
   finalizeSynchronization,
+  abortSynchronization,
 } from '../../framework/synchronization';
 import { executeIntegrationInstance } from '../../framework/execution';
 import { createIntegrationInstanceForLocalExecution } from '../../framework/execution/instance';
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to
 - An `errorCode`, `errorId`, and `reason` is now displayed in step failure
   events and logs.
 - Event logging around errors thrown in `validateInvocation` function.
+- Abort synchronization job when `run` and `sync` command fails.
```
aiwilliams, line 17: I think this is a good example of why we should consider separating the execution code into a separate package from the code the integrations themselves depend on. Should we have a bug in the runner, we can avoid upgrading all the open source integrations or worse, not upgrading them and leaving them on old versions.

src/cli/commands/run.ts
```diff
@@ -51,24 +52,42 @@ export function run() {
...
+        const synchronizationResult = await finalizeSynchronization({
+          ...synchronizationContext,
+          partialDatasets: executionResults.metadata.partialDatasets,
+        });
+
+        log.displaySynchronizationResults(synchronizationResult);
+      } catch (err) {
+        await logger.flush();
+
+        if (!logger.isHandledError(err)) {
```
aiwilliams, line 78: I like this approach quite a bit actually, since we're using a stateful object as a logger 😄 It puts the responsibility in the right place I think, keeps us from having to keep track in the places where we do the logging by having to tag the error object.

src/framework/synchronization/index.ts
```diff
@@ -180,3 +189,33 @@ async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+
+  try {
+    const response = await apiClient.post(
+      `/persister/synchronization/jobs/${job.id}/abort`,
+      { reason },
+    );
+    abortedJob = response.data.job;
+  } catch (err) {
+    const errorMessage = 'Error occurred while aborting synchronization job.';
+    throw synchronizationApiError(err, errorMessage);
```
aiwilliams, line 217: I'm curious more than anything... why alloc a local var only to use it in one place?

CHANGELOG.md
aiwilliams: cc/ @philidem

src/framework/synchronization/index.ts
charlieduong94: Oh, this is copy pasta'd from another function. I had a logger.error there, but with how errors are now handled, I ended up removing that. There's a little inconsistency in the sync functions, so lemme clean that up rnrq.

Issue #124, opened 2020-05-04
Support loading integration from single index file


philidem:
Instead of auto-loading steps, we should move toward having a single index file that contains metadata and imports steps (instead of scanning directory).

aiwilliams:
This is related to #123

aiwilliams:
@charlieduong94 is this closed by #150 ?

charlieduong94:
Yes, this should be closed.

PR #150, opened 2020-05-11
Support loading invocation config from src/index

charlieduong94:
Addresses #124. I'm only deprecating the convention style loading for now and not removing things entirely so that others that are actively working on integrations can still pick up some changes (data model upgrades or small additions) without breaking their integration.

I want to add better validation and potentially introduce some type-gen or other tool to help keep static types and runtime validation code in sync. That may become a larger change and what's here is enough to allow us to start making deploys easier with the managed runtime. When we do the 1.0.0 release, we'll drop the convention loading entirely.

### DIFF
--- a/src/framework/__tests__/config.test.ts
+++ b/src/framework/__tests__/config.test.ts
@@ -14,6 +14,8 @@ import {
   restoreProjectStructure,
 } from '../../__tests__/loadProjectStructure';
 
+import * as log from '../../log';
+
 jest.mock('../../log');
 
 afterEach(() => {
@@ -65,6 +67,20 @@ describe('loadConfig', () => {
         },
       ],
     });
+  });
+
+  test('logs deprecation message for projects with directory structure', async () => {
+    const typescriptProjectPath = getProjectDirectoryPath('typeScriptProject');
+
+    await loadConfig(path.resolve(typescriptProjectPath, 'src'));
+
+    const warnSpy = jest.spyOn(log, 'warn');
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+    expect(warnSpy).toHaveBeenCalledWith(
+      expect.stringMatching(
+        /Automatically loading configurations from a directory structure is deprecated/,
+      ),
+    );
   });
 });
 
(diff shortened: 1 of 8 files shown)


aiwilliams:
> I'm only deprecating the convention style loading for now and not removing things entirely so that others that are actively working on integrations can still pick up some changes

Thank you!

src/framework/__tests__/config.test.ts
```diff
@@ -66,6 +68,20 @@ describe('loadConfig', () => {
...
+  test('logs deprecation message for projects with directory structure', async () => {
+    const typescriptProjectPath = getProjectDirectoryPath('typeScriptProject');
+
+    await loadConfig(path.resolve(typescriptProjectPath, 'src'));
+
+    const warnSpy = jest.spyOn(log, 'warn');
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+    expect(warnSpy).toHaveBeenCalledWith(
+      expect.stringMatching(
+        /Automatically loading configurations from a directory structure is deprecated/,
```
aiwilliams, line 81: Most excellent.

src/framework/__tests__/config.test.ts
```diff
@@ -205,6 +221,46 @@ describe('JavaScript projects', () => {
...
+          id: 'fetch-users',
+          name: 'Fetch Users',
+          types: ['my_user'],
+          executionHandler: expect.any(Function),
+        },
+      ],
+    });
+  });
+
+  test('throws error if configuration is not ', async () => {
```
aiwilliams, line 255: ... not what? 😄

charlieduong94: F

src/framework/config.ts
```diff
@@ -13,14 +14,17 @@ import {
...
- * Supports a convention over configuration approach
- * to building integrations.
- *
- * ./src/instanceConfigFields
- * ./src/steps/*
- * ./src/getStepStartStates
- * ./src/validateInvocation
+ * Loads integration invocation configuration.
  */
 export async function loadConfig(
```
philidem, line 29: Hopefully we don't use `loadConfig` in the deployment host project. I could see how this might be used by the CLI but in deployed environment I would expect something like:

```typescript
const invocationConfig = require('@jupiterone/graph-project');
await runIntegration(invocationConfig); // <--- something else runs the integration using the invocation config
```

charlieduong94: Yep, we won't be using the `loadConfig` function deployed env with integrations switching over to the `index` file.

Issue #156, opened 2020-05-13
Provide guidance and error type for integrations raising errors


aiwilliams:
The gitlab integration stores an [API response code in `IntegrationError.code`](https://github.com/JupiterOne/graph-gitlab/blob/3c8fd0f19f4a16915788f0338926e0d2c10b2187/src/provider/GitlabClient.ts#L131), and there may be another integration doing something similar. However, I think now that was not the best choice.

In https://github.com//JupiterOne/integration-sdk/blob/master/src/errors.ts#L7, and everywhere in the integration-sdk currently, the values for `code` are constant-like strings, and there are error classes that assign `code` so that we don't see the strings repeated where the errors are thrown (great!).

We should avoid having each integration develop their own set of codes for common things, wo that we can make some things consistent for runtime monitoring and debugging. We'll need a clear way for integrations to raise certain types of errors, and limit the use of new error codes as much as reasonable.

Consider this proposal:

```ts
/**
 * A type representing errors generated or handled by the integration runtime.
 * Integrations should not throw this error type directly. See below for
 * options.
 *
 * All errors occurring during execution of an integration should ultimately end
 * up as an `IntegrationError`, or a subclass thereof, each expressing a `code`
 * reflecting the class of the error. In case an unhandled `Error` occurs, which
 * has no `code`, then `"UNEXPECTED_ERROR"` will be used in logging the error.
 *
 * Integration code should throw these exceptions:
 *
 * ```
 *   // Thrown during `validateInvocation`, anything the user should see that
 *   // can help them fix the problem on their own.
 *   throw new IntegrationValidationError("Something that will be displayed to user so they can fix configuration")
 *
 *   // Typically thrown during `validateInvocation`. User will be informed of
 *   // what we're calling to test authentication, and the response code.
 *   throw new IntegrationProviderAuthenticationError({
 *     cause: err,
 *     endpoint: "https://something.com/api/authcheckwhatever",
 *     status?: err.status,
 *     message?: err.statusText
 *   });
 *
 *   // Thrown during resource fetching steps. The runtime will show these along
 *   // with information about what could not be obtained thanks to insufficient
 *   // access.
 *   throw new IntegrationProviderAuthorizationError({
 *     type: [], // the `_type`s that could not be obtained
 *     cause: err,
 *     endpoint: "https://something.com/api/resource",
 *     status?: err.status,
 *     message?: err.statusText
 *   });
 *
[...]

PR #157, opened 2020-05-13
Add some types of execution errors

aiwilliams:
Advances #156. I think we can start with these, but we'll also need something like @charlieduong94 mentioned, `logger.authError`, both for the integration-sdk to use when that type of error is thrown from a step, as well as by the step itself when it doesn't want to terminate, only report that and move on to some other work.


src/errors.ts
```diff
@@ -1,27 +1,80 @@
...
+ * "Message for production analysis")`. For example:
+ *
+ * ```ts
+ * try {
+ *   // something that could go wrong, but it is handled by the integration
+ * } catch (error) {
+ *   // log it for production analysis, but carry on because we know what we're doing
+ *   logger.warn({ err: error }, "Nothing we can't handle, but someone may come asking");
+ * }
+ * ```
```
charlieduong94, line 63: :heart:

PR #161, opened 2020-05-14
Fix uploadData from uploading the entire dataset for each batch.

austinkelleher:


src/framework/synchronization/index.ts
```diff
@@ -191,7 +191,7 @@ async function uploadData<T extends UploadDataLookup, K extends keyof T>(
       if (batch.length) {
         await apiClient.post(
           `/persister/synchronization/jobs/${job.id}/${type}`,
-          { [type]: data },
```
charlieduong94, line 194: oof thanks for the fix

PR #164, opened 2020-05-14
Add test for upload function to prevent regression with batching

charlieduong94:


charlieduong94:
I should have never rushed in a last minute chunking change

PR #172, opened 2020-05-17
Only transfer value if it is one of the supported types (skip object …

erkangz:
…or object[])

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,19 @@ and this project adheres to
 
 ## Unreleased
 
+### Added
+
 - Export `SetupRecordingInput` from `@jupiterone/integration-sdk/testing` to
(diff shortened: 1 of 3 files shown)


aiwilliams:
Excellent. I’m still hoping we can get consensus on the pattern of writing tests in the integrations that show what should be included in the generated entity so we can do more with this function over time, and folks will know when the generated entity in the integrations is no longer correct (vs TS types and hand-coded property transfers).

aiwilliams:
Version 2.0.0 is coming from https://github.com/JupiterOne/integration-sdk/pull/168 in a few minutes. These changes here can be 2.0.1 after rebasing.

aiwilliams:
I'm going to pull this branch to update the CHANGELOG and add a test.

PR #179, opened 2020-05-20
Add logger functions to publish events

### LINKED ISSUE #178: Add logger functions for publishing events and error details

philidem:
This change adds `context.logger.publishEvent(...)` and `context.logger.publishErrorEvent(...)` functions and allows semi-structured properties to be added to error description (fixes #178)

**Example usage:**

```javascript
context.logger.publishEvent({
  name: 'some event',
  description: 'some description',
});

context.logger.publishErrorEvent({
  err: new Error('some error!'),
  message: 'Something bad happened',
  name: 'error',
  // arbitrary event data that will be serialized into event description
  eventData: {
    responseText: 'Account is disabled'
  },
  // stuff that should only be logged
  logData: {
    region: process.env.AWS_REGION
  }
})
```


src/framework/execution/__tests__/logger.test.ts
```diff
@@ -431,7 +432,102 @@ describe('createErrorEventDescription', () => {
       'testing',
     );
     expect(description).toEqual(
-      `testing (errorCode=${error.code}, errorId=${errorId}, reason=soba)`,
+      `testing (errorCode="${error.code}", errorId="${errorId}", reason="soba")`,
     );
   });
 });
+
+describe('#publishEvent', () => {
```
aiwilliams, line 440: I had hoped to keep from having integrations publish random events to the job log because I had noticed that when that feature was available, there was no consistency in the logging across integrations, and the information presented to users was haphazard. I do think we need to communicate more to users than we are currently able to communicate, but I think it to be particularly errors that are non-terminal during a step, and which should not cause a Pager Duty event, so I had hoped that we could use` context.logger.warn({ err, ... }, "Optional message here")`, and then re-use types found in https://github.com/JupiterOne/integration-sdk/blob/master/src/framework/execution/error.ts. This would allow us to develop consistent presentation of events as well as dashboards for debugging.

So it would come out as:

```ts
try {
  ... some API call ...
} catch(err) {
  // I can handle this, but want to note it
  logger.warn({
    err: new IntegrationProviderAPIError({
      cause: err,
      endpoint: "https://provider.com/api/resource",
      status: response.status,
      statusText: response.statusText,
    })
  });
}
```

src/framework/execution/__tests__/logger.test.ts
```diff
@@ -431,7 +432,102 @@ describe('createErrorEventDescription', () => {
...
+    await logger.flush();
+
+    expect(postSpy).toHaveBeenCalledTimes(1);
+    expect(postSpy).toHaveBeenNthCalledWith(1, expectedEventsUrl, {
+      events: [{ name: 'the name', description: 'the description' }],
+    });
+  });
+});
+
+describe('#publishErrorEvent', () => {
```
aiwilliams, line 475: Similar to `logger.warn` described elsewhere, we've been working toward having `logger.error` log an error, and then publish the error event (this already happens with thrown errors). This approach, along with the [other logger functions](https://github.com/JupiterOne/integration-sdk/blob/master/src/framework/execution/types/logger.ts#L45), is meant to help avoid random events in the job event log, but also provides a level of indirection so that when `j1-integration collect` is run, and there is [no synchronization job](https://github.com/JupiterOne/integration-sdk/blob/5932e1bee60e96b838aa8b372389518803c2393a/src/framework/execution/logger.ts#L192), `publishEvent` does nothing. That is, `logger.*` methods will output to the local log and, depending on the nature of their logging, `publishEvent` is used to record them in the job event log when one is available.

I see that you'd like to add additional data to the logged event for the user to see, and I suspect from previous conversations that this will be related to troubleshooting exceptional situations. I think we should consider having the error types (such as `IntegrationProviderAPIError` or whatever other error type we should create) receive the additional error context (the data) and instead of serializing it to the single line message, improve the data storage for the event to support storing that error object with its additional properties. Then we can show the error in the job log with the short message, and expand it to show the complete error.

philidem:
We discussed this PR and decided to add these "publish event" functions for temporary use but we will make other future changes to always publish events when using the other integration logger functions.

PR #181, opened 2020-05-20
Initial changes to refactor integration execution

austinkelleher:
A user may want to execute some steps but not tie the execution to a specific integration instance, but rather just follow the same patterns as the managed integrations.


austinkelleher:
I opened a PR for the incorrect branch...

PR #182, opened 2020-05-20
Initial changes to decompose integration execution functionality

austinkelleher:
A user may want to execute some steps but not tie the execution to a specific integration instance, but rather just follow the same patterns as the managed integrations.

### DIFF
--- a/src/framework/config.ts
+++ b/src/framework/config.ts
@@ -4,11 +4,12 @@ import path from 'path';
 
 import { IntegrationError } from '../errors';
 import {
+  ExecutionContext,
   GetStepStartStatesFunction,
   IntegrationInstanceConfigFieldMap,
   IntegrationInvocationConfig,
   IntegrationStep,
-  InvocationValidationFunction,
+  IntegrationInvocationValidationFunction,
 } from '../framework/execution';
 import * as log from '../log';
 
@@ -101,7 +102,7 @@ export function loadInstanceConfigFields(
  */
 export function loadValidateInvocationFunction(
   projectSourceDirectory: string = path.join(process.cwd(), 'src'),
-): InvocationValidationFunction | undefined {
+): IntegrationInvocationValidationFunction | undefined {
   return loadModuleContent(
     path.resolve(projectSourceDirectory, 'validateInvocation'),
   );
@@ -110,10 +111,10 @@ export function loadValidateInvocationFunction(
 /**
  * Loads getStepStartStates function from ./src/getStepStartStates.(t|j)s
  */
-export function loadGetStepStartStatesFunction(
+export function loadGetStepStartStatesFunction<T extends ExecutionContext>(
   projectSourceDirectory: string = path.join(process.cwd(), 'src'),
 ) {
-  return loadModuleContent<GetStepStartStatesFunction>(
+  return loadModuleContent<GetStepStartStatesFunction<T>>(
     path.resolve(projectSourceDirectory, 'getStepStartStates'),
   );
 }
--- a/src/framework/execution/dependencyGraph.ts
+++ b/src/framework/execution/dependencyGraph.ts
@@ -9,22 +9,22 @@ import {
   TypeTracker,
 } from './jobState';
 import {
-  IntegrationExecutionContext,
-  IntegrationStep,
-  IntegrationStepExecutionContext,
+  Step,
   IntegrationStepResult,
-  IntegrationStepResultStatus,
-  IntegrationStepStartStates,
+  StepResultStatus,
+  StepStartStates,
+  ExecutionContext,
+  StepExecutionContext,
 } from './types';
 
 /**
  * This function accepts a list of steps and constructs a dependency graph
  * using the `dependsOn` field from each step.
  */
-export function buildStepDependencyGraph(
-  steps: IntegrationStep[],
-): DepGraph<IntegrationStep> {
-  const dependencyGraph = new DepGraph<IntegrationStep>();
+export function buildStepDependencyGraph<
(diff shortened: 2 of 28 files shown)


src/framework/execution/dependencyGraph.ts
```diff
@@ -104,16 +104,15 @@ export function executeStepDependencyGraph(
    * NOTE: the untouched input graph is used for determining
    * results since the working graph gets mutated.
    */
-  function stepHasDependencyFailure(step: IntegrationStep) {
+  function stepHasDependencyFailure(stepId: string) {
```
philidem, line 107: Was it better to use a `string` for argument instead of `IntegrationStep`? It seems like there is less safety for making this a string. Also, we should consider using `Opaque` type when you string identifiers like this. It would allow you to do something like this:

```
export type Opaque<K, T> = T & { __TYPE__: K };
export type StepId = Opaque<'StepId',string>;
```

I started using this `Opaque` type in other projects. See https://evertpot.com/opaque-ts-types/

austinkelleher: `IntegrationStep` has been completely replaced with `Step` in this file. Most functions in this file only require the step ID, but were taking the entire step. The new type generics make this very verbose and I was trying to simplify it but only passing what the function actually needs. It's more pure and easier to maintain this way.

src/framework/config.ts
```diff
@@ -4,11 +4,12 @@ import path from 'path';
 
 import { IntegrationError } from '../errors';
 import {
+  ExecutionContext,
   GetStepStartStatesFunction,
   IntegrationInstanceConfigFieldMap,
   IntegrationInvocationConfig,
   IntegrationStep,
-  InvocationValidationFunction,
+  IntegrationInvocationValidationFunction,
```
aiwilliams, line 12: 👍 Good rename here, matches the other `Integration*` things like `IntegrationInvocationConfig` etc.

src/framework/execution/dependencyGraph.ts
```diff
@@ -8,22 +8,22 @@ import {
...
  * using the `dependsOn` field from each step.
  */
-export function buildStepDependencyGraph(
-  steps: IntegrationStep[],
-): DepGraph<IntegrationStep> {
-  const dependencyGraph = new DepGraph<IntegrationStep>();
+export function buildStepDependencyGraph<
+  TStepExecutionContext extends StepExecutionContext
+>(steps: Step<TStepExecutionContext>[]): DepGraph<Step<TStepExecutionContext>> {
+  const dependencyGraph = new DepGraph<Step<TStepExecutionContext>>();
```
aiwilliams, line 26: FWIW, I'm not a huge fan of `TStepExecutionContext` over just `T` in many cases, where there aren't but a few lines of code (a good practice for most functions!), but it certainly is clear.

src/framework/execution/types/context.ts
```diff
@@ -2,23 +2,28 @@ import { IntegrationInstance, IntegrationInstanceConfig } from './instance';
...
+export interface ExecutionContext {
+  logger: IntegrationLogger;
+}
+
 /**
  * @param TConfig the integration specific type of the `instance.config`
  * property
  */
-export interface IntegrationExecutionContext<
+export type IntegrationExecutionContext<
```
aiwilliams, line 13: I'm still a bit fuzzy on the differences between `type` and `interface`, but I think this change will prevent the use of `T extends IntegrationExecutionContext` right? Any reason to think that would never be something folks should do?

src/framework/execution/types/validation.ts
```diff
@@ -1,6 +1,11 @@
-import { IntegrationExecutionContext } from './context';
+import { ExecutionContext, IntegrationExecutionContext } from './context';
 import { IntegrationInstanceConfig } from './instance';
 
-export type InvocationValidationFunction<
+export type InvocationValidationFunction<T extends ExecutionContext> = (
+  context: T,
+) => Promise<void> | void;
+
+// TODO: This shouldn't be necessary anymore
```
aiwilliams, line 8: It's unclear to me what isn't necessary anymore, nor what needs to be done about it 😄

src/framework/execution/types/context.ts
charlieduong94: You can extend `types` (see [this playground](https://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=10&pc=2#code/C4TwDgpgBAKlC8UDeUAeAuKA7ArgWwCMIAnKAXwCgKBLLYEgMwEMBjaAISglXqwBMAzrGRQKUcVBCZchEhUoUWAeywDgUApk6IkYiRigBGADR7xUo-KA)). Older versions of typescript did not support it though (see [docs](https://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases)).

It seems like the only real difference between an `interface` and `type` is that an `interface` allows for declaration merging (see [example](https://www.typescriptlang.org/play/?ssl=15&ssc=2&pln=1&pc=1#code/C4TwDgpgBAKlC8UDeUAeAuKA7ArgWwCMIAnKAXwCgKBLLYEgMwEMBjaAISglXqwBMAzrGRQKUcVBCZchEhUo06jVhxFiJAL2n4ixeVRYB7LAOBQCmToiTrxGKAFYANLcmZnrrY-lA))

aiwilliams: I see. I wonder if that declaration merging feature is a good thing or not. Reminds me of re-opening classes across files, or monkey-patching interfaces one does not own. I appreciate the power of being able to do what is necessary, but it comes with so much responsibility 😄

PR #186, opened 2020-05-20
Warn when undeclared types are detected

### LINKED ISSUE #74: Warn when entity and relationship types collected from a step do not match the step's declared types
The `JobState` object that gets created for each step can store a set of types that it has run across. After the step has completed, we can log a warning that unexpected types were detected. This information could also be stored on the `IntegrationStepResult` and summarized at the end.

charlieduong94:
step_type_detection

Closes #74


src/log.ts
```diff
@@ -53,9 +88,26 @@ function getStepStatusText(status: IntegrationStepResultStatus) {
   }
 }
 
-export function displaySynchronizationResults(job: SynchronizationJob) {
-  info('\nSynchronization results:\n');
-  info(`Synchronization job status: ${chalk.cyan(job.status)}`);
-  info(`Entities uploaded: ${chalk.cyan(job.numEntitiesUploaded)}`);
-  info(`Relationships uploaded: ${chalk.cyan(job.numRelationshipsUploaded)}`);
+function logUndeclaredAndExtraneousTypes(
```
aiwilliams: I'm not sure about the use of `And`, because I keep thinking (when reading the tests as well) that this is referring to two kinds of types, some that are undeclared and some that are extraneous, though I think it is referring only to a single kind of type: those which appeared in the collection but were not declared. Additionally, the data is tracked as `undeclaredTypes`, not `undeclaredAndExtraneousTypes`.

What do you think about just `logUndeclaredTypes` (and removing "extraneous" elsewhere as well)?

charlieduong94: Updated. I originally had extraneous types being logged later decided to stop doing so because for some integration developers that have a different provider setup, it could cause end up causing unnecessary noise.

PR #187, opened 2020-05-21
Fix tags as string

erkangz:


src/framework/data/tagging.ts
```diff
@@ -61,21 +61,44 @@ const TRUE_BOOLEAN_REGEX = /^true$/i;
...
-          m[k] = v;
-        }
-        return m;
-      }, {});
+    if (typeof tags === 'string') {
+      if (tags.includes(',')) {
+        (entity as any).tags = tags
+          .split(',')
+          .filter((t) => t.trim().length > 0)
+          .map((t) => t.trim());
```
aiwilliams, line 81: I think this could be changed to avoid trimming twice (and below as well):
```
.split(',')
.map((t) => t.trim())
.filter((t) => t.length > 0)
```

erkangz: Good idea. Do u want to include in one of your upcoming PRs?

PR #194, opened 2020-05-27
Append additional message details for events related to provider auth error

### LINKED ISSUE #136: Allow integrations to report resource access authorization errors to job event log
When an integration is not allowed to read a resource endpoint, there needs to be a consistent way of reporting the event to the job log.

charlieduong94:
Closes #136 . I decided to not add a `logger.authError` function and opted to instead just enhance the existing messages with more information if we happen to come across an `IntegrationProviderAuth*` error.

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -5,7 +5,7 @@ export const UNEXPECTED_ERROR_REASON =
 export const PROVIDER_AUTH_ERROR_DESCRIPTION =
   ' Failed to access provider resource.' +
   ' This integration is likely misconfigured or has insufficient permissions required to access the resource.' +
-  " Please ensure your integration's configuration settings are setup correctly.";
+  " Please ensure your integration's configuration settings are set up correctly.";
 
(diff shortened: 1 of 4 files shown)


src/errors.ts
```diff
@@ -2,6 +2,11 @@ export const UNEXPECTED_ERROR_CODE = 'UNEXPECTED_ERROR';
 export const UNEXPECTED_ERROR_REASON =
   'Unexpected error occurred executing integration! Please contact us in Slack or at https://support.jupiterone.io if the problem continues to occur.';
 
+export const PROVIDER_AUTH_ERROR_DESCRIPTION =
+  ' Failed to access provider resource.' +
+  ' This integration is likely misconfigured or has insufficient permissions required to access the resource.' +
+  " Please ensure your integration's configuration settings are setup correctly.";
```
austinkelleher, line 8: This is super minor, but "setup" is technically a noun. I think "set up" is correct wording.

PR #201, opened 2020-06-01
Shift to monorepo structure

charlieduong94:
This shifts the project into a monorepo structure. atm the sdk is not broken up yet and is just moved into `packages/integration-sdk`. Doing this helped me incrementally break things up more easily.

I broke things up into multiple prs to help with reviewing, but I also recommend just pulling down the `setup-lerna-publishing` branch from PR #209 and browsing files from your own machine. It will help with getting a better overall idea of how the project will be structured.

Next PR to look at after this one is #202 .


jest.config.js
```diff
@@ -1,20 +1,5 @@
...
-  ],
-  testPathIgnorePatterns: ['<rootDir>/dist/', '<rootDir>/node_modules/'],
-  collectCoverage: false,
-  globals: {
-    'ts-jest': {
-      isolatedModules: true,
-    },
-  },
-  testEnvironment: 'node',
+  projects: ['<rootDir>/packages/*'],
```
charlieduong94, line 4: This allows for tests to be run from the root while allowing each package to still have their own test configuration.

package.json
```diff
@@ -1,85 +1,34 @@
...
+  "private": true,
+  "workspaces": [
+    "packages/*"
+  ],
   "scripts": {
-    "lint": "eslint . --cache --fix --ext .ts,.tsx",
-    "format": "prettier --write '**/*.{ts,js,json,css,md,yml}'",
-    "type-check": "tsc --noEmit",
+    "lint": "eslint --ext .ts,.tsx,.js,.jsx .",
+    "type-check": "tsc -b packages/*",
```
charlieduong94, line 9: Experimenting with tsc's new composite builds. This setup is different from the typical `--noEmit` flow we usually do because code actually gets transpiled. The main benefit of this new change is that typescript can store some metadata and do faster incremental builds.

charlieduong94: Ran into issues with this on windows... For now we'll have to specify each package. I'll work on a script to make this less painful as more packages get introduced.

PR #202, opened 2020-06-01
Add integration sdk core package

charlieduong94:
This adds the `@jupiterone/integration-sdk-core` package. This package just contains some shared types and utilities that an integration would directly use. My hope is to keep this package pretty lean.

Previous PR is #201, next PR is #203


packages/integration-sdk-core/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@jupiterone/integration-sdk-core",
+  "version": "0.1.0",
```
aiwilliams, line 3: At some point along the path, these versions will all match?

packages/integration-sdk-core/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "description": "The SDK for developing JupiterOne integrations",
+  "main": "dist/src/index.js",
+  "types": "dist/src/index.d.ts",
+  "repository": "git@github.com:JupiterOne/integration-sdk.git",
```
aiwilliams, line 7: This is stale now, may be worth addressing as these files are created.

charlieduong94: Good catch, changing now

packages/integration-sdk-core/package.json
charlieduong94: Yeah, after all of the monorepo changes have been merged, we'll version all of the packages and then publish.

PR #209, opened 2020-06-01
Setup lerna publishing

charlieduong94:
This sets up lerna publishing. There are a few minor changes the build commands but it's nothing major. I changed this to use lerna's independent versioning because there are likely going to be packages added to this that don't really need to be updated if an integration package changes and vice versa.

Going forward, I would like to start following https://www.conventionalcommits.org/en/v1.0.0/. Lerna supports performing the semantic versioning for us based on the commit messages.


aiwilliams:
> Going forward, I would like to start following https://www.conventionalcommits.org/en/v1.0.0/. Lerna supports performing the semantic versioning for us based on the commit messages.

I would be super appreciative of any effort we can make as a team to develop consistency in our commit messages, especially if that consistency can pay off in automation and meaningful improvements to our use of semver.

I do wish this particular standard did not promote bad sentence structure, and they are terribly inconsistent in their use of periods. 😆

aiwilliams:
I've been talking about this more with @fomentia, and I'm growing concerned that using commit messages to drive versioning may be problematic. One big reason I'm concerned: producing a meaningful CHANGELOG is more than the sum of parts. However, I'll remain optimistic and see where this goes. 😄

fomentia:
I have some concerns.

1. Adding automation around commit messages is a side effect. Commit messages are not code.
2. Commit messages as code encourages rebasing during the review process, which is evil.
3. (((People))) are not going to follow these rules perfectly and especially aren't going to want to look up the _16_ specs every time they make a commit.

Also, does the person who wrote that document not know how git works? The image I'm greeted with when I open the site is atrocious.

Also, as @aiwilliams mentioned, the document is inconsistent with itself.

charlieduong94:
Those are very valid concerns. I'm convinced that we shouldn't use the conventional commits for versioning now.

PR #211, opened 2020-06-02
Fix convertProperties array of null

aiwilliams:


charlieduong94:
So I was thinking about introducing a changelog per package, but considering we are trying to keep the integration related projects versioned together, perhaps maintaining the changelog for all of the `integration-sdk-*` packages in `packages/integration-sdk` (the shell project) is the best thing to do.

charlieduong94:
I was also thinking of just "resetting" to `1.0.0` with these new packages, but since these packages are really just an evolution of the older `integration-sdk` I wonder if it makes sense start at `4.0.0` so that we can maintain the history and not confuse people with a versioning reset. Thoughts?

aiwilliams:
> start at `4.0.0` so that we can maintain the history and not confuse people with a versioning reset.

I think this would be nice.

> So I was thinking about introducing a changelog per package, but considering we are trying to keep the integration related projects versioned together, perhaps maintaining the changelog for all of the integration-sdk-* packages in packages/integration-sdk (the shell project) is the best thing to do.

I thought you had mentioned somewhere along the line that you wanted to do independent versions, is that not the case? If were have together versioning, then this approach to CHANGELOG in the workspace project sounds good to me.

aiwilliams:
> start at 4.0.0 so that we can maintain the history and not confuse people with a versioning reset.

>> I think this would be nice.

I just remembered that we would be pointing to two new packages (no longer `@jupiterone/integration-sdk`), so now I feel you on starting at 1.0.0 for those new packages 😄 I'd be fine either way, no longer seeing a big advantage to 4.0.0 (if we stop publishing `integration-sdk`).

charlieduong94:
> I thought you had mentioned somewhere along the line that you wanted to do independent versions, is that not the case? If were have together versioning, then this approach to CHANGELOG in the workspace project sounds good to me.

Yeah, independent versioning I think will be helpful when we start adding more packages to this repo that are not necessarily related to integrations. 

For the `integration-sdk` prefixed packages though, it might make it more clear which versions of a given package can work with each other. Ex. `integration-sdk-core` 1.0.0 might not be compatible with `integration-sdk-cli` 2.0.0. We could alternatively maintain version compatibility table somewhere I suppose, but I feel like keeping those packages sync'd would be nice.

PR #214, opened 2020-06-03
Publish sdk packages

charlieduong94:


packages/integration-sdk/README.md
```diff
@@ -5,6 +5,9 @@ Note: This package has been been deprecated and should no longer be used.
 The functionality provided by this package has been broken up into multiple
 smaller packages.
 
-If you are developing an integration, you can use the
-`@jupiterone/integration-sdk-core` and `@jupiterone/integration-sdk-cli`
-packages.
+This is now a shell package used to store historical information about the older
+`@jupiterone/integration-sdk` and also track changes made to the the
```
charlieduong94, line 9: dammit just notice the double `the`. Mind if I fix that after the merge so I don't have to delete all the tags?

.github/workflows/build.yml
```diff
@@ -69,5 +69,4 @@ jobs:
         run: |
           echo "//registry.npmjs.org/:_authToken=${NPM_AUTH_TOKEN}" > .npmrc
           yarn
-          yarn build
```
aiwilliams, line 72: Personally, I'm really glad for the use of documented lifecycles and features of `nmp`/`yarn` (such as `prepack` and `files`). This compounds the value of learning about those tools, and has the effect of making things more predictable across projects.

aiwilliams:
> > Are you using a tool to manage the versions and inter-dependencies between the packages/*?
> 
> Yeah, Lerna does the inter-dependency updates for us when doing the versioning

Oh, perfect, I knew we were using Lerna, but didn't know it would update all these. 👍

PR #217, opened 2020-06-03
Add @jupiterone/integration-sdk-dev-tools package

charlieduong94:
The config exported here is pretty much the same config used in this project.

### DIFF
--- a/packages/integration-sdk-dev-tools/README.md
+++ b/packages/integration-sdk-dev-tools/README.md
@@ -1,0 +1,60 @@
+# @jupiterone/integration-sdk-dev-tools
+
+This package contains some common dependencies and configuration files used in
+JupiterOne integrations.
+
+This module is opinionated and bundles in the following dependencies:
+
+- `typescript`
+- `prettier`
+- `jest`
+- `eslint`
+- `husky`
+- `lint-staged`
+
+For convenience, this package also comes bundled with the
+`@jupiterone/integration-sdk-cli` module.
+
+## Installation
+
+```
+npm install -D @jupiterone/integration-sdk-dev-tools
+
+# or
+
+yarn add -D @jupiterone/integration-sdk-dev-tools
+```
+
+## Usage
+
+Create `prettier.config.js` at root of your project that contains:
+
+```javascript
+module.exports = require('@jupiterone/integration-sdk-dev-tools/config/prettier');
+```
+
+Create `lint-staged.config.js` at root of your project that contains:
+
+```javascript
+module.exports = require('@jupiterone/integration-sdk-dev-tools/config/lint-staged');
+```
+
+Create `.huskyrc.js` at root of your project that contains:
+
+```javascript
+module.exports = require('@jupiterone/integration-sdk-dev-tools/config/husky');
+```
+
+Create `jest.config.js` at root of your project that contains:
+
+```javascript
+module.exports = require('@jupiterone/integration-sdk-dev-tools/config/jest');
+```
+
+Create `tsconfig.json` at root of your project that contains:
+
+```json
+{
+  "extends": "./node_modules/@jupiterone/integration-sdk-dev-tools/config/typescript"
+}
+```
--- a/packages/integration-sdk-dev-tools/config/jest.js
+++ b/packages/integration-sdk-dev-tools/config/jest.js
@@ -1,0 +1,18 @@
+module.exports = {
+  preset: 'ts-jest',
+  clearMocks: true,
+  restoreMocks: true,
+  testMatch: [
+    '<rootDir>/**/*.test.ts',
+    '!**/node_modules/*',
+    '!**/dist/*',
+    '!**/*.bak/*',
+  ],
+  collectCoverage: false,
+  globals: {
+    'ts-jest': {
+      isolatedModules: true,
+    },
+  },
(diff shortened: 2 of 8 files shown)


packages/integration-sdk-dev-tools/config/husky.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  hooks: {
+    'pre-commit': 'lint-staged',
+    'pre-push': 'yarn test:ci',
```
aiwilliams, line 4: I still long for not-CI test run on `git push`, one that provides reasonable confidence but doesn't make me wait for a CI run on my machine, and then another CI run in the PR, and then another CI run in master. That is, would be great to just run something like `yarn type-check && yarn test`, which I acknowledge may be exactly what `test:ci` is doing ATM. I saw another project somewhere around here that was doing `pre-push: 'yarn prepush'`, that's nice.

packages/integration-sdk-dev-tools/config/jest.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
+  preset: 'ts-jest',
+  clearMocks: true,
+  restoreMocks: true,
```
aiwilliams, line 4: Yah!

packages/integration-sdk-dev-tools/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+- `typescript`
+- `prettier`
+- `jest`
+- `eslint`
+- `husky`
+- `lint-staged`
+
+For convenience, this package also comes bundled with the
+`@jupiterone/integration-sdk-cli` module.
```
aiwilliams, line 16: I wonder if this is the best thing to do for folks who will be using the CLI without a need for all these dev tools, but perhaps that isn't ever going to be the case.

charlieduong94: The `integration-sdk-cli` package will be more useful if we decide to tack on the `integration` subcommand to the next gen version of the `j1` cli. Idk if doing that will be very useful, but it could be depending on what functionality we decide to add to the `integration-sdk-cli` (external integration stuff). 

For now though, it's just kinda nice having really small, well focused packages.

aiwilliams: I may be misunderstanding your comment... I totally love the idea of a `-cli` package, and small packages.

I was only wondering if the `-cli` needs to always come with the `-dev-tools`, which appear to be all about formatting/managing code in an integration project. I suppose there is no use of the `-cli` today without maintaining some code, so it is fine for them to travel together in that way. Of course, the `-cli` itself will have the `-dev-tools` as it's own `devDependencies` as well, I just think it may be more clear to have the integration projects depend on both themselves in `devDependencies` (not depend on the transitive property).

packages/integration-sdk-dev-tools/config/husky.js
charlieduong94: Okay I have a few ideas on how we can make this nicer. 

For prepush testing, what do you think about using jest's `--changedSince` feature for the `prepush`?
Docs: https://jestjs.io/docs/en/cli.html#--changedsince

I have not personally used it, but it could provide a little bit of confidence while allowing pushes to be quick(er) than a full test run.

Additionally, we could enable typescript's `incremental` option to make typechecking faster. It's on in this repo because we use composite builds, but we can also turn it on projects that do not use `composite` and `references`.

packages/integration-sdk-dev-tools/README.md
charlieduong94: ah okay sorry I think I misunderstood your original comment.

So I originally thought about making `dev-tools` a project that only contained common config and development deps, but then @philidem made a comment about bundling the cli with it (in integrations slack channel). After thinking about it a little, it kinda made sense to just include it in the `dev-tools` package because the CLI is really is just a kind of development tool. Our own internal `dev-tools` package comes with a few cli commands for doing various dev related things so it seemed to fit the pattern.

I don't feel too strongly about having the cli bundled in the `dev-tools` project though so if you would prefer we keep the `dev-tools` project a little more pure, I'm down remove the CLI from it.

aiwilliams: Oh, haha, I misunderstood, thinking the `-cli` came with the `-dev-tools`, instead of what you've done (`-dev-tools` comes with `-cli`), which is excellent. Sorry!

packages/integration-sdk-dev-tools/config/husky.js
aiwilliams: That `--changedSince` is pretty interesting, and `incremental` type checking would be awesome where supported as well! Do these kinds of workflow optimizations make sense to you? Others following along? I know there is some value in running the whole shebang.

charlieduong94: I think the incremental type checking is a clear win. As for `--changedSince`, if it's only used for testing in `prepush` then it should relatively harmless since CI will run the full test suite anyway. I also suspect as developers are making changes to a project they will end up running a full test run out of habit.

charlieduong94:
I've changed the husky prepush to call `yarn prepush`. I'll change the draft template to make the `prepush` command run tests with `--changedSince`

PR #218, opened 2020-06-03
Fix regression of _type option createIntegrationRelationship

aiwilliams:
Changes in https://github.com/JupiterOne/sdk/pull/200/files were lost in the shuffle to monorepo. This fixes the regression and advances docs around relationship mappings and what happens with these properties.


charlieduong94:
aww damn, sorry I didn't realize another change slipped through

aiwilliams:
> sorry I didn't realize another change slipped through

No worries! Moving to monorepo was a heavy lift, and I had the opportunity to add some docs 😄

PR #222, opened 2020-06-04
Fix createIntegrationRelationship mappings _key

aiwilliams:


aiwilliams:
@philidem I know we want to break this function up, but that will be a major version change. This captures the correct behavior without all the noise that will come with interface changes.

PR #232, opened 2020-06-12
Add interface for emitting metrics to be published

charlieduong94:
This builds on top of #231 and allows for `metric` events to be listened for. Our managed environment will handle collecting the `metric` events (pr for that coming soon).

### DIFF
--- a/packages/integration-sdk-cli/src/commands/run.ts
+++ b/packages/integration-sdk-cli/src/commands/run.ts
@@ -16,6 +16,7 @@ import {
 } from '@jupiterone/integration-sdk-runtime';
 
 import { loadConfig } from '../config';
+import { Metric } from '@jupiterone/integration-sdk-core';
 
 export function run() {
   return createCommand('run')
@@ -29,6 +30,8 @@ export function run() {
       const apiKey = getApiKeyFromEnvironment();
       const apiBaseUrl = getApiBaseUrl({ dev: !!process.env.JUPITERONE_DEV });
       log.debug(`Configuring client to access "${apiBaseUrl}"`);
+
+      const startTime = Date.now();
 
       const apiClient = createApiClientWithApiKey({
         apiBaseUrl,
@@ -48,12 +51,16 @@ export function run() {
         integrationInstanceId,
       });
 
+      logger = synchronizationContext.logger;
+
       const eventPublishingQueue = createEventPublishingQueue(
         synchronizationContext,
       );
-      logger.on('event', (event) => eventPublishingQueue.enqueue(event));
+      const metrics: Metric[] = [];
 
-      logger = synchronizationContext.logger;
+      logger
+        .on('event', (event) => eventPublishingQueue.enqueue(event))
+        .on('metric', (metric) => metrics.push(metric));
 
       const invocationConfig = await loadConfig();
 
@@ -94,6 +101,12 @@ export function run() {
         });
 
         log.displaySynchronizationResults(abortResult);
+      } finally {
+        logger.publishMetric({
+          name: 'total-duration',
+          value: Date.now() - startTime,
+          unit: 'Milliseconds',
+        });
       }
     });
 }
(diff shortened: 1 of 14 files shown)


packages/integration-sdk-cli/src/commands/run.ts
```diff
@@ -87,6 +101,12 @@ export function run() {
         });
 
         log.displaySynchronizationResults(abortResult);
+      } finally {
+        logger.publishMetric({
+          name: 'total-duration',
+          value: Date.now() - startTime,
+          unit: 'Milliseconds',
```
aiwilliams, line 108: I'm interested in learning more about the metric names you've already thought about, or whether there is a metric naming standard you think we should adopt (there are [many to choose from](https://www.google.com/search?q=metrics+naming+convention&oq=naming+metrics&aqs=chrome.2.69i57j0l4.5870j0j7&sourceid=chrome&ie=UTF-8)). I think this is worth discussion early on. For example, I would prefer `duration-total`, so that all `duration-*` metrics are grouped instead of all `total-*` things across different metrics (like `total-requests` or `total-steps`).

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -218,7 +219,11 @@ export function executeStepDependencyGraph<
         if (isStepEnabled(stepId) && stepDependenciesAreComplete(stepId)) {
           removeStepFromWorkingGraph(stepId);
           promiseQueue.add(() =>
-            executeStep(step).catch(handleUnexpectedError),
+            timeOperation({
+              logger: executionContext.logger,
+              metricName: `${stepId}-duration`,
```
aiwilliams, line 224: I wonder if this should be `duration-step`, with `dimensions: { stepId }`, so there is not an explosion of metrics and we can see averages across all steps easily within an integration.

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -60,15 +61,18 @@ export async function executeIntegrationInstance(
...
-      instance,
-      logger,
-    },
-    config,
-  );
-
-  return result;
+  return timeOperation({
+    logger,
+    metricName: 'collection-duration',
```
aiwilliams, line 66: `duration-collect`?

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -153,23 +163,35 @@ async function getPartialDatasets() {
...
-        if (Array.isArray(parsedData.relationships)) {
-          await uploadData(context, 'relationships', parsedData.relationships);
-        }
-      } catch (err) {
-        throw synchronizationApiError(err, 'Error uploading collected data');
-      }
-    },
+  await timeOperation({
+    logger: context.logger,
+    metricName: 'synchronization-upload-duration',
```
aiwilliams, line 168: `duration-sync-upload` to match the name of the `sync` command?

aiwilliams: Hum, perhaps just `duration-sync` to match `duration-collect`, and those are then `duration-<command>` 🤔

packages/integration-sdk-cli/src/commands/run.ts
charlieduong94: Okay, I'm down with changing the metrics to be prefixed with the type of thing they are tracking.

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
charlieduong94: Seems reasonable

packages/integration-sdk-runtime/src/synchronization/index.ts
charlieduong94: hmm when I think of `duration-sync` I think of the duration of the entire synchronization process. I think I would prefer to stick with `duration-sync-upload` and have another metric  like `duration-sync-finalize` for tracking how long it takes our backend process to run. Later on we can figure out a good way to track the total duration (including the time waiting for sync to start) and name that `duration-sync`.

aiwilliams: Makes sense, thanks @charlieduong94

PR #253, opened 2020-07-27
[J1-4221] Add getEntity() to the jobState

ndowmon:
Gets an existing entity in the job state by `{ _key, _type }`. Throws if there is not *exactly* one match.

### DIFF
--- a/packages/integration-sdk-core/src/types/jobState.ts
+++ b/packages/integration-sdk-core/src/types/jobState.ts
@@ -4,6 +4,10 @@ export interface GraphObjectFilter {
   _type: string;
 }
 
+export interface GraphObjectLookupKey extends GraphObjectFilter {
+  _key: string;
+}
+
 export type GraphObjectIteratee<T> = (obj: T) => void | Promise<void>;
 
 /**
@@ -63,6 +67,12 @@ export interface JobState {
    * to add a single relationship to the collection.
    */
   addRelationships: (relationships: Relationship[]) => Promise<void>;
+
+  /**
+   * Gets an entity by _key and _type
+   * Throws when !== 1 entity
+   */
+  getEntity: (lookupKey: GraphObjectLookupKey) => Promise<Entity>
 
   /**
    * Allows a step to iterate all entities collected into the job state, limited
--- a/packages/integration-sdk-core/src/types/jobState.ts
+++ b/packages/integration-sdk-core/src/types/jobState.ts
@@ -4,7 +4,8 @@ export interface GraphObjectFilter {
   _type: string;
 }
 
-export interface GraphObjectLookupKey extends GraphObjectFilter {
+export interface GraphObjectLookupKey {
+  _type: string;
   _key: string;
 }
 
--- a/packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
+++ b/packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
@@ -4,6 +4,8 @@ import { Sema } from 'async-sema';
 import {
   Entity,
   Relationship,
+  IntegrationError,
+  GraphObjectLookupKey,
   GraphObjectFilter,
   GraphObjectIteratee,
 } from '@jupiterone/integration-sdk-core';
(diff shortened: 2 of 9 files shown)


packages/integration-sdk-core/src/types/jobState.ts
```diff
@@ -4,6 +4,10 @@ export interface GraphObjectFilter {
   _type: string;
 }
 
+export interface GraphObjectLookupKey extends GraphObjectFilter {
+  _key: string;
+}
+
```
aiwilliams, line 10: I think this should not extend `GraphObjectFilter` because it will only get a single entity, which is by definition the unique combination of `_type` and `_key` - we won't expect to support other properties when `GraphObjectFilter` is expanded to support more properties.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -56,6 +58,32 @@ export class FileSystemGraphObjectStore {
...
+          entities.push(e);
+        }
+      }
+    );
+
+    if (entities.length !== 1) {
+      throw new IntegrationError({
+        code: 'GET_ENTITY_ERROR',
+        message: `Failed to find entity with _type=${_type}, _key=${_key}`,
+      });
```
aiwilliams, line 81: I think it would be good to consider throwing the [existing duplicate key error type](https://github.com/JupiterOne/sdk/blob/0cb90d3c432902870b0ca2b4d11794ec9482ea4d/packages/integration-sdk-core/src/errors.ts#L146).

packages/integration-sdk/CHANGELOG.md
```diff
@@ -9,6 +9,13 @@ and this project adheres to
 
 ## Unreleased
 
+### Added
+
+- Added `getEntity({ _type, _key })` function
+- `j1-integration visualize`:
+  - Nodes (entities) are colored by `_type`.
+  - Mapped relationships are shown as dashed lines (edges)
```
aiwilliams, line 17: Thank you!

packages/integration-sdk/README.md
```diff
@@ -11,3 +11,15 @@ This is now a shell package used to store historical information about the older
 
 If you are developing an integration, use the `@jupiterone/integration-sdk-core`
 and `@jupiterone/integration-sdk-cli` packages.
+
+## Versioning this project
```
aiwilliams, line 15: 💯

packages/integration-sdk-core/src/types/jobState.ts
ndowmon: Fair!

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
ndowmon: Sounds good. I broke this up to throw `IntegrationMissingKeyError` when length === 0 and `IntegrationDuplicateKeyError` when length > 1.

aiwilliams:
Hmm, tests are failing for some reason, but LGTM.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
```diff
@@ -242,6 +242,47 @@ describe('addRelationships', () => {
...
+
+    const nonMatchingEntities = times(25, () =>
+      generateEntity({ _type }),
+    );
+
+    await store.addEntities(storageDirectoryPath, [
+      ...nonMatchingEntities,
+    ]);
+
+    await expect(store.getEntity({ _type, _key })).rejects.toThrow(`Failed to find entity with _type=${_type}, _key=${_key}`);
```
austinkelleher, line 282: Looks like this test assertion is failing because of missing parens in the string.

ndowmon: Good catch, fixed.

Issue #62, opened 2020-04-22
createIntegrationRelationship has confusing method signature and auto complete suggestions


philidem:
```typescript
export declare function createIntegrationRelationship(options: DirectRelationshipOptions | DirectRelationshipLiteralOptions | MappedRelationshipOptions | MappedRelationshipLiteralOptions): Relationship
```

The union of so many types makes usage of the function hard to understand. We should consider decomposing into separate functions.

aiwilliams:
I really like the idea of `createEntityRelationship` and `createMappedRelationship` 👍

PR #258, opened 2020-07-28
Pass the source `_type` to mapped relationships.

ndowmon:
Allow mapped relationships to receive the source `_type`, either from the `source` property on `MappedRelationshipOptions`, or the optional `_mapping.sourceEntityType` property on `MappedRelationshipOptionsLiteral`.

Since this was related to #62 as well, I deprecated `createIntegrationRelationship` and exported `createDirectRelationship` and `createMappedRelationship`. These changes are in a separate commit from the primary functionality above.

### DIFF
--- a/packages/integration-sdk-core/src/data/createIntegrationRelationship.ts
+++ b/packages/integration-sdk-core/src/data/createIntegrationRelationship.ts
@@ -129,11 +129,9 @@ type TargetEntity = TargetEntityProperties & {
 type AdditionalRelationshipProperties = { [key: string]: any };
 
 /**
- * Create an `IntegrationRelationship`.
- *
- * `DirectRelationshipOptions` and `MappedRelationshipOptions` are recommended
- * over literal forms. Older integrations may need to use the literal forms to
- * control values for some reason or other.
+ * @deprecated
+ * @see createDirectRelationship
+ * @see createMappedRelationship
  */
 export function createIntegrationRelationship(
   options:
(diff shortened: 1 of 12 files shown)


packages/integration-sdk-core/src/data/createIntegrationRelationship.ts
```diff
@@ -142,38 +140,69 @@ export function createIntegrationRelationship(
...
+ *
+ * `DirectRelationshipOptions` is recommended over `DirectRelationshipOptionsLiteral`. Older integrations may need to use the literal forms to control values for some reason or other.
+ */
+export function createDirectRelationship(
+  options: DirectRelationshipOptions | DirectRelationshipLiteralOptions,
+): ExplicitRelationship {
+  //
+  if ('fromType' in options) {
+    return createRelationship(options);
+  //
```
aiwilliams, line 161:
```suggestion
  if ('fromType' in options) {
    return createRelationship(options);
```

packages/integration-sdk-core/src/data/createIntegrationRelationship.ts
```diff
@@ -142,38 +140,69 @@ export function createIntegrationRelationship(
...
+ *
+ * `MappedRelationshipOptions` is recommended over `MappedRelationshipOptionsLiteral`. Older integrations may need to use the literal forms to control values for some reason or other.
+ */
+export function createMappedRelationship(
+  options: MappedRelationshipOptions | MappedRelationshipLiteralOptions,
+): MappedRelationship {
+  // MappedRelationshipLiteralOptions
+  if ('_mapping' in options) {
+    return createMappedRelationshipLiteral(options);
+  // MappedRelationshipOptions
```
aiwilliams, line 185:
```suggestion
  if ('_mapping' in options) {
    return createMappedRelationshipLiteral(options);
```
To match the other function where removing the empty comments is suggested 😁  I don't feel these add to the clarity of the function.

Issue #73, opened 2020-04-23
Support mapped relationships in visualization


aiwilliams:
Mapped relationships are a bit more sophisticated than simple direct relationships (see https://github.com/JupiterOne/integration-sdk/blob/master/src/framework/data/createIntegrationRelationship.ts#L77). They should produce target entities if necessary and build the relationship.

cc/ @erkangz @softwarewright @philidem

PR #260, opened 2020-07-29
J1-4227: Match array/string _class in visualize

ndowmon:
See https://airtable.com/tblq5xeQXHTFS1YnX/viw8UgryHmlyw1g27/recHxh0xfVGqefgk7?blocks=hide

The `visualize` command will NOT match the following targetEntity. The local matcher needs to be updated to make sure the `j1-integration visualize` command works properly:

```
const entity = {
  _key: 'some-key',
  _class: ['User'],
  _type: 'user-type'
}

const targetEntity = {
  _key: 'some-key',
  _class: 'User',
  _type: 'user-type'
}
```

In addition, `RelationshipMapping` interfaces use the `targetFilterKeys` property to declare which properties should be used to match target entities. This PR adds logic to match objects on sets of `targetFilterKey`s.


ndowmon:
This code has gotten a little more complex than I was hoping, maybe could be split into a few files. However, I really wanted to keep the interface to `createMappedRelationshipNodesAndEdges` the same/similar to what it has been, so that `generateVisualization.ts` remains simple(ish).

ndowmon:
Also closes #73, didn't know of the existence of this issue in my previous PR for mapped relationship visualization.

PR #277, opened 2020-08-06
Fix jobState::getEntity by adding await in front of floating promise

### LINKED ISSUE #276: Testing with `getEntity` fails to find known entity
When running a test, `jobState.iterateEntities` seems to list all entities correctly, but if we try to query an entity with `jobState.getEntity` with matching `_type` and `_key` it throws an error.

E.g. `entities = [{ name: 'stefan', _key: 'jfrog-user-stefan', _type: 'jfrog-user' }, {...}, {...}, ...]`, `await jobState.getEntity({ _type: 'jfrog-user', _key: 'jfrog-user-stefan', });` throws an error (`"Failed to find entity with _type=jfrog_user, _key=jfrog-user-stefan"`). The same, however, doesn't happen when we just run the `j1-integration collect`, it's able to correctly return it then (same values).

neurovelho:
This fixes #276.

### DIFF
--- a/packages/integration-sdk-testing/src/__tests__/jobState.test.ts
+++ b/packages/integration-sdk-testing/src/__tests__/jobState.test.ts
@@ -63,6 +63,24 @@ describe('entities', () => {
     expect(jobState.collectedEntities).toEqual([]);
   });
 
+  test('getEntity returns entity from jobState when _type and _key match', async () => {
+    const jobState = createMockJobState();
+    await jobState.addEntities(inputEntities);
+
+    const result = await jobState.getEntity({ _type: 'test_a', _key: 'a' });
+
+    expect(result).toEqual(inputEntities[0]);
+  });
+
+  test('getEntity throws Error when entity cannot be found in jobState', async () => {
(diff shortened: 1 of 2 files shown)


packages/integration-sdk-testing/src/__tests__/jobState.test.ts
```diff
@@ -63,6 +63,26 @@ describe('entities', () => {
...
+    const jobState = createMockJobState();
+    await jobState.addEntities(inputEntities);
+
+    const result = await jobState.getEntity({ _type: 'test_a', _key: 'a' });
+
+    expect(result).toEqual(inputEntities[0]);
+  });
+
+  test('getEntity throws Error when entity cannot be found in jobState', async () => {
+    expect.hasAssertions();
```
fomentia, line 77: Do you actually need `expect.hasAssertions()`? If you're awaiting expects with promises, I don't think you should.

neurovelho: You are indeed correct. I removed this and the other one which was also unnecessary.

PR #286, opened 2020-08-17
Fixes #283 - Change jobState.getEntity to only take a _key

### LINKED ISSUE #283: Fix `getEntity` to take only a `_key`
Entities are supposed to be unique, and this is enforced by the `duplicateKeyTracker`. Therefore, there is no need to include the `_type` in the function. Dropping it will reduce coupling between the `_type` and `_key` properties, as we push hard to avoid binding the two value unnecessarily.

austinkelleher:
See #283 for info

Old API:

```ts
jobState.getEntity({
  _key: 'abc123',
  _type: 'my_type'
});
```

New API:

```ts
jobState.getEntity('abc123');
```

### DIFF
--- a/packages/integration-sdk/CHANGELOG.md
+++ b/packages/integration-sdk/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to
 
 ## Unreleased
 
+### Changed
+
+- [BREAKING] Fixed [#283](https://github.com/JupiterOne/sdk/issues/283) - Change
+  `jobState.getEntity` to only to a `_key`.
+
 ### 2.10.0 - 2020-08-06
 
 ### Changed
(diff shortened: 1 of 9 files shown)


packages/integration-sdk/CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to
 
 ## Unreleased
 
+### Changed
+
+- [BREAKING] Fixed [#283](https://github.com/JupiterOne/sdk/issues/283) - Change
+  `jobState.getEntity` to only to a `_key`.
```
aiwilliams, line 15:
```suggestion
  `jobState.getEntity` to require `_key` value only.
```

austinkelleher:
I'm going to wait to merge this change until I make the other documentation breaking changes, so we can batch those into a single major. It's just a number, but it can be confusing to see multiple back-to-back majors in an open source project.

PR #287, opened 2020-08-17
Fixes #284 - Add jobState.findEntity function

### LINKED ISSUE #284: Add `findEntity` to complement `getEntity`
`getEntity` throws an error when the entity is not found. This may force the developer to code this pattern:

```ts
let entity: Entity;
try {
  entity = await jobState.getEntity(...);
} catch (err) {
  entity = null;
}
```

When the entity is expected to possibly be not found, it is more expressive and less code to write:

```ts
  const entity = await jobState.findEntity(...);
```

austinkelleher:
#284

### DIFF
--- a/docs/integrations/development.md
+++ b/docs/integrations/development.md
@@ -297,7 +297,18 @@ await iterateEntities({ _type: 'my_integration_user' }, async (userEntity) => {
 ```
 
 Specific entities can be looked up using the entity `_key` property via the
-`getEntity` function.
+`findEntity` function. `findEntity` will return `null` if the entity does not
+exist.
+
+Example usage:
+
+```typescript
+const entity = await jobState.findEntity('myentitykey');
+```
+
+Similarly to `findEntity`, `getEntity` can also be used to look up an entity by
+`_key`. `getEntity` will throw a `IntegrationMissingKeyError` error if the
+entity does not exist.
 
 Example usage:
 
--- a/packages/integration-sdk-core/src/types/jobState.ts
+++ b/packages/integration-sdk-core/src/types/jobState.ts
@@ -76,6 +76,13 @@ export interface JobState {
   getEntity: (_key: string) => Promise<Entity>;
 
   /**
+   * Finds an entity by `_key` and returns `null` if the entity does not exist.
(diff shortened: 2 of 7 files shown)


docs/integrations/development.md
```diff
@@ -297,7 +297,18 @@ await iterateEntities({ _type: 'my_integration_user' }, async (userEntity) => {
...
+
+Example usage:
+
+```typescript
+const entity = await jobState.findEntity('myentitykey');
+```
+
+Similarly to `findEntity`, `getEntity` can also be used to look up an entity by
+`_key`. `getEntity` will throw a `IntegrationDuplicateKeyError` error if the
+entity already exists.
```
aiwilliams, line 311: I think this should be `IntegrationMissingKeyError`.

Related: looks like the [test job state does not throw the right kind of error](https://github.com/JupiterOne/sdk/blob/c93f69eb0fd90012e58f5accfe1f7c1b8bdf7634/packages/integration-sdk-testing/src/jobState.ts#L122).

I'm curious whether there is a way we can avoid a "mock" job state class at some point, because I've seen a couple times where it doesn't do the same thing as the real one.

packages/integration-sdk-core/src/types/jobState.ts
```diff
@@ -70,6 +70,13 @@ export interface JobState {
    */
   getEntity: (_key: string) => Promise<Entity>;
 
+  /**
+   * Finds an entity by `_key` and returns `null` if the entity does not exist.
+   * This function will also throw an error if multiple entities are found with
+   * the same `_key` value.
```
aiwilliams, line 76: This shouldn't be possible, right, since the JobState should be protecting from writing duplicate keys?

aiwilliams: Looking at this again, the reason I wondered about having this in the comment is, should users wrap `findEntity` in a `try/catch`, since the docs say it could throw an error. We introduced the function to keep from having to wrap with a `try/catch`. Perhaps the function can be sure to prevent throwing, always returning `null` if the underlying object store throws an error of not found?

PR #290, opened 2020-08-18
Publish

ndowmon:
- @jupiterone/cli@2.10.1
 - @jupiterone/integration-sdk-cli@2.10.1
 - @jupiterone/integration-sdk-core@2.10.1
 - @jupiterone/integration-sdk-dev-tools@2.10.1
 - @jupiterone/integration-sdk-private-test-utils@2.10.1
 - @jupiterone/integration-sdk-runtime@2.10.1
 - @jupiterone/integration-sdk-testing@2.10.1

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jupiterone/cli",
-  "version": "2.10.0",
+  "version": "2.10.1",
   "description": "The JupiterOne cli",
   "main": "dist/src/index.js",
   "types": "dist/src/index.d.ts",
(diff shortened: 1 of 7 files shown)


aiwilliams:
Will this re-target to `master`? I wonder if that will break publishing? https://github.com/JupiterOne/sdk/blob/f9d470165d9e69beb854845651ccd8edf426e5ac/.github/workflows/build.yml#L41

ndowmon:
> Will this re-target to `master`? I wonder if that will break publishing?

I'm not as familiar with the publish workflow. Now that it's retargeted it seems to me that the workflow should publish but could you take a look @aiwilliams?

PR #302, opened 2020-08-24
Fixes #301 - StepRelationshipMetadata in Step should require

### LINKED ISSUE #301: StepRelationshipMetadata in Step should require RelationshipClass from @jupiterone/data-model
Related: https://github.com/JupiterOne/sdk/issues/285

We recently introduced metadata in `Step`, which is used to auto-generate documentation. The `StepRelationshipMetadata` currently has a `_class` property with the type `string`. We should change that to be `RelationshipClass` from `@jupiterone/data-model`. See here: https://github.com/JupiterOne/sdk/blob/master/packages/integration-sdk-core/src/types/step.ts#L83

Old:

```typescript
export interface StepRelationshipMetadata extends StepGraphObjectMetadata {
  _class: string;
  sourceType: string;
  targetType: string;
}
```

New:

```
import { RelationshipClass } from '@jupiterone/data-model';

export interface StepRelationshipMetadata extends StepGraphObjectMetadata {
  _class: RelationshipClass;
  sourceType: string;
  targetType: string;
}
```

austinkelleher:
RelationshipClass from @jupiterone/data-model

### DIFF
--- a/packages/integration-sdk-core/src/types/step.ts
+++ b/packages/integration-sdk-core/src/types/step.ts
@@ -4,7 +4,7 @@ import {
   StepExecutionContext,
 } from './context';
 import { IntegrationInstanceConfig } from './instance';
-
+import { RelationshipClass } from '@jupiterone/data-model';
 export interface StepStartState {
   /**
    * Indicates the step is disabled and should not be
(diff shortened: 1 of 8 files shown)


austinkelleher:
@ndowmon Good question. This is a breaking change. We have not released the new version of the SDK yet. The original metadata changes haven't been released.

ndowmon:
@austinkelleher Thanks for the clarification :)

aiwilliams:
That is a good question @ndowmon. I now wonder if we should be taking this approach. 🤔 Are there alternative ways to ensure the data is generated with valid relationship classes? Perhaps it should be in the schema alone?

PR #313, opened 2020-08-25
Fixes #301 - Register inputEntities & inputRelationships in mockJobState

### LINKED ISSUE #301: StepRelationshipMetadata in Step should require RelationshipClass from @jupiterone/data-model

ndowmon:


packages/integration-sdk-testing/src/jobState.ts
```diff
@@ -46,6 +46,19 @@ export function createMockJobState({
...
+    });
+    typeTracker.registerType(e._type);
+  });
+  inputRelationships.forEach((r) => {
+    duplicateKeyTracker.registerKey(r._key as string, {
+      _type: r._type,
+    });
+    typeTracker.registerType(r._type as string);
+  });
+
```
aiwilliams, line 61: How do you feel about moving this logic below the `const ...` function declarations, just before the `return ...`:
```ts
addEntities(inputEntities);
addRelationships(inputRelationships);
```
I might be missing a finer detail 😄

ndowmon: @aiwilliams the `addEntities` and `addRelationships` functions also push graph objects onto the `collected<GraphObjects>` list. That breaks tests since `collected<GraphObjects>` should only contain graph objects that were added _during_ the step.

I could alternatively refactor a bit to something like this:

```typescript
const addEntitiesToDuplicateKeyTracker = (newEntities: Entity[]) => {
  newEntities.forEach((e) => {
    duplicateKeyTracker.registerKey(e._key, {
      _type: e._type,
    });
    typeTracker.registerType(e._type);
  });
};
addEntitiesToDuplicateKeyTracker(inputEntities);

const addEntities = async (newEntities: Entity[]): Promise<Entity[]> => {
  addEntitiesToDuplicateKeyTracker(newEntities);
  collectedEntities = collectedEntities.concat(newEntities);
  return Promise.resolve(newEntities);
}

const addRelationshipsToDuplicateKeyTracker = (newRelationships: Relationship[]) => {
  newRelationships.forEach((r) => {
    duplicateKeyTracker.registerKey(r._key as string, {
      _type: r._type,
    });
    typeTracker.registerType(r._type as string);
  });
};
addRelationshipsToDuplicateKeyTracker(inputRelationships);

const addRelationships = async (newRelationships: Relationship[]) => {
  addRelationshipsToDuplicateKeyTracker(newRelationships);
  collectedRelationships = collectedRelationships.concat(newRelationships);
  return Promise.resolve();
};

const iterateEntities = async <T extends Entity = Entity>(
  filter,
  iteratee,
) => {
  const filteredEntities = [...inputEntities, ...collectedEntities].filter(
    (e) => e._type === filter._type,
  );
  for (const entity of filteredEntities as T[]) {
    await iteratee(entity);
  }
};
```

It removes duplicate code which may be worth the change, but it's still not particularly attractive.

PR #322, opened 2020-08-29
Cleanup development.md typos

cdeevfrr:
I found a few errors in this doc and wanted to throw up a PR to see if they should be changed.

### Fixes a broken link to IntegrationInvocationConfig type in Readme
I was getting errors from this link, and it looks like the repo file structure may have changed. It looks like this fixes it, but I don't have enough context to be 100% certain.

It may be worth checking to see if the same change needs to happen in other places too, but I intended this to be a quick-fix, and don't want to invest more time until I learn if this is helpful or not.

### A few header link failures

I found some links that looked like this:

image

and found a [source](https://gist.github.com/rachelhyman/b1f109155c9dafffe618) saying that github has a special format for header links.

### Small typos

A few small typos (missing letters or duplicate words)

### DIFF
--- a/docs/integrations/development.md
+++ b/docs/integrations/development.md
@@ -14,7 +14,7 @@ This SDK supports building JupiterOne integrations using either JavaScript or
 TypeScript.
 
 The execution process expects the integration to produce an object conforming to
-the [`IntegrationInvocationConfig`](/src/framework/execution/types/config.ts)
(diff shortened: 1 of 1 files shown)


cdeevfrr:
That sounds good - please let me know if this workflow is sending you a bunch of emails, I'm used to being loose with commits and strict about PR merging, but if you're getting notified for every commit I'll be a lot more careful.

austinkelleher:
Thanks a lot for the fixes! This is very helpful.

PR #324, opened 2020-08-30
Fixes #321 - j1-integration document output in alphabetical order by …

### LINKED ISSUE #321: Make `j1-integration document` output ordered
The order of documentation that's generated from the `j1-integration document` command seems to be based on the dependency tree of steps. Changes to the dependency tree cause git diffs that don't clearly show what if anything changed.

Documentation could be ordered alphabetically to suppress this behavior.

Might be good to fix alongside #309 and #316

austinkelleher:
…_type


austinkelleher:
@ndowmon I think ordering alphabetically by resource name makes sense. I'll make the changes. Thanks for the feedback.

ndowmon:
> For example, we probably would maybe prefer...

Thanks for submitting to my fierce demands.

PR #328, opened 2020-09-05
Add shouldReportErrorToOperator to avoid alerts

aiwilliams:
We're getting Sentry alerts for a few errors that are considered to be a kind that may only be resolved by the user changing something in their configuration. These are handled in the sdk runtime code in some special way, avoiding a `logger.error` call, and then re-thrown. The managed runtime sends them to Sentry, which we don't want to do. This adds a function to allow the managed runtime to see if it should not report these.

The places where these error types are handled in the SDK:

* [`IntegrationProviderAuthorizationError`](https://github.com/search?q=org%3AJupiterOne+IntegrationProviderAuthorizationError&type=Code)
* [`IntegrationProviderAuthenticationError`](https://github.com/search?q=org%3AJupiterOne+IntegrationProviderAuthenticationError&type=Code)
* [`IntegrationValidationError`](https://github.com/search?q=org%3AJupiterOne+IntegrationValidationError&type=Code)

I'd like to consider a change that makes the SDK not throw these kinds of errors, but I think it will take a bit more care because the managed runtime may need more changes to live without the error being thrown.

### DIFF
--- a/packages/integration-sdk-runtime/src/logger/index.ts
+++ b/packages/integration-sdk-runtime/src/logger/index.ts
@@ -11,19 +11,22 @@ import {
   IntegrationInstanceConfigFieldMap,
   IntegrationInvocationConfig,
   IntegrationLogger as IntegrationLoggerType,
-  IntegrationProviderAuthenticationError,
-  IntegrationProviderAuthorizationError,
   IntegrationStepExecutionContext,
-  IntegrationValidationError,
   InvocationConfig,
+  isProviderAuthError,
   Metric,
-  PROVIDER_AUTH_ERROR_DESCRIPTION,
+  shouldReportErrorToOperator,
   StepExecutionContext,
   StepMetadata,
   SynchronizationJob,
   UNEXPECTED_ERROR_CODE,
   UNEXPECTED_ERROR_REASON,
 } from '@jupiterone/integration-sdk-core';
+
+export const PROVIDER_AUTH_ERROR_HELP =
+  ' Failed to access provider resource.' +
+  ' This integration is likely misconfigured or has insufficient permissions required to access the resource.' +
+  " Please ensure your integration's configuration settings are set up correctly.";
 
 // eslint-disable-next-line
 const bunyanFormat = require('bunyan-format');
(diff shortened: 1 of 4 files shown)


packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -158,6 +161,24 @@ export class IntegrationLogger extends EventEmitter
...
+   * * warn(err, ...)
+   * * error(err, ...)
+   *
+   * This can be used by outer `try/catch` blocks to avoid logging the same
+   * error twice:
+   *
+   * ```js
+   * if (!logger.isHandledError(err)) {
+   *   logger.error(err);
+   * }
```
aiwilliams, line 177: I don't care for this pattern. Why can't the logger just consult whether it's seen an err before and avoid re-logging it? This is closely related to the question, why does the SDK re-throw errors that it handles?

packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -279,10 +300,10 @@ export class IntegrationLogger extends EventEmitter
       `Error occurred while validating integration configuration.`,
     );
 
-    if (isUserConfigError(err)) {
-      this.warn({ errorId, err }, description);
-    } else {
+    if (shouldReportErrorToOperator(err)) {
       this.error({ errorId, err }, description);
+    } else {
+      this.warn({ errorId, err }, description);
```
aiwilliams, line 306: I'm a little uneasy about the impression this code gives, that `.error()` is somehow going to report to the operator. We used to have it that way, when there was a metric on `level: 50` log events that would trigger an alert and PD notification, but it isn't this way now.

The [code that calls this `validationFailure()` function](https://github.com/JupiterOne/sdk/blob/77a669a26de908cac6040a5ef480da2d622cd747/packages/integration-sdk-runtime/src/execution/executeIntegration.ts#L101) re-throws the error, which means that the code catching that also needs to know whether the error is meant to be reported to an operator.

I'm thinking we should consider a path toward avoiding throwing errors out of the SDK which do not need to be reported to the operator. The managed integration runtime should be able to report everything thrown out of the sdk.

What do you think @austinkelleher @ndowmon ?

austinkelleher: I agree that the managed integration runtime should be able to report everything thrown out of the SDK

austinkelleher (approved):
Thanks for making this change.

PR #330, opened 2020-09-08
Fixes #310 - Create metadata graph

### LINKED ISSUE #310: Proposal: `j1-integration document` should generate a graph of entity/relationship metadata
The `j1-integration document` command has enough information to be able to generate a graph of metadata about entities & relationships generated in an integration. 

It would be great to publish this information, possibly in the same way that some AWS documentation is currently published: https://support.jupiterone.io/hc/en-us/articles/360041433273-Data-Model-for-AWS-IAM-Access-and-Trusts

Which is sourced from: https://github.com/JupiterOne/docs/blob/master/docs/data-model/aws-iam.md

ndowmon:
This change adds the `j1-integration visualize-types` command, which uses [visjs](https://visjs.org/) to generate a graph of metadata types found in the integration.

As an intermediate step, the process of fetching metadata information was pulled out of the `j1-integration document` command and placed in a standalone location in the code. As a result, `j1-integration` `document`, `visualize-types`, and `types` will now generate a json at `<project-dir>/docs/types.json` which is leveraged by both `j1-integration document` and `j1-integration.

Usage is below.

```sh
yarn j1-integration visualize-types -y azure_role_assignment -y azure_user -y azure_resource_group
```
Screen Shot 2020-09-07 at 8 54 43 PM

An additional comment - the `j1-integration visualize-types` output is a little quick and dirty, and the output isn't _great_ for exposing to customers.  Exposing the `types.json` file was not necessary, but was strategic. I could easily see a future that leverages the J1 graph viewer as a package that uses the open-sourced `types.json` file to generate an attractive and interactive graph at <support.jupiterone.io> or even within the app at <tenant.us.jupiterone.io/integrations>.


packages/integration-sdk-cli/src/commands/visualize-types.ts
```diff
@@ -0,0 +1,145 @@
...
+function getNodesFromStepEntityMetadata(
+  entities: StepEntityMetadata[],
+  options?: {
+    types?: string[],
+    edges?: Edge[],
+  }
+): Node[] {
+  let targetTypes: Set<string> | undefined = undefined;
+  if (options?.types !== undefined && options.edges !== undefined) {
+    targetTypes = new Set();
```
charlieduong94, line 111: Minor, but imo it would be slightly cleaner to just move the variable declaration of `targetTypes` into this `if`. This would allow you to avoid the optional chaining in the filter function.

packages/integration-sdk-cli/src/commands/visualize-types.ts
```diff
@@ -0,0 +1,145 @@
...
+  log.info('\nGenerating local graph from types metadata...\n');
+
+  const edges = getEdgesFromStepRelationshipMetadata(metadata.relationships, { types });
+  const nodes = getNodesFromStepEntityMetadata(metadata.entities, { types, edges });
+  
+
+  const visHtml = generateVisHTML(
+    nodes,
+    edges,
+  )
```
charlieduong94, line 88: iirc, the other cli command has a snapshot test. It would be nice to do the same here and also unit test the construction of the metadata graph.

charlieduong94 (approved):
This is pretty cool! I can def see this being really helpful for new devs coming into an existing integration. Thanks for your work!

packages/integration-sdk-cli/src/commands/types.ts
```diff
@@ -0,0 +1,154 @@
...
+  projectPath: string;
+  typesFilePath: string;
+}
+
+export function types() {
+  return createCommand('types')
+    .description('Collects types metadata for all steps')
+    .option(
+      '-p, --project-path <directory>',
+      'Absolute file path to the integration project directory. Defaults to the current working directory.',
```
aiwilliams, line 25: Nit, I'd drop "file path" for just "path", since it is a directory and not a file.
```suggestion
      'Absolute path to the integration project directory. Defaults to the current working directory.',
```

packages/integration-sdk-cli/src/commands/types.ts
```diff
@@ -0,0 +1,154 @@
...
+      process.cwd(),
+    )
+    .option(
+      '-t, --types-file-path <path>',
+      'Absolute file path to the JSON file that should be created/overwritten. Defaults to {CWD}/docs/types.json.',
```
aiwilliams, line 30: Just to avoid saying "file" twice 😄 
```suggestion
      'Absolute path to the JSON file that should be created/overwritten. Defaults to {CWD}/docs/types.json.',    
```

packages/integration-sdk-cli/src/commands/types.ts
```diff
@@ -0,0 +1,154 @@
...
+export function types() {
+  return createCommand('types')
+    .description('Collects types metadata for all steps')
+    .option(
+      '-p, --project-path <directory>',
+      'Absolute file path to the integration project directory. Defaults to the current working directory.',
+      process.cwd(),
+    )
+    .option(
+      '-t, --types-file-path <path>',
```
aiwilliams, line 29: I really like `-o, --output-path` as an option for tools that output stuff. I see [`document` is using `-f, --documentation-file-path <directory>`](https://github.com/JupiterOne/sdk/blob/cca95a549d83c8096cfc50c671d6a30603738d4d/packages/integration-sdk-cli/src/commands/document.ts#L35), I think it'd be good to get that one changed.

packages/integration-sdk-cli/src/commands/types.ts
```diff
@@ -0,0 +1,154 @@
...
+
+export async function executeTypesAction(
+  options: TypesCommandArgs,
+): Promise<void> {
+  const { projectPath } = options;
+  const typesFilePath =
+    options.typesFilePath ||
+    getDefaultTypesFilePath(projectPath);
+
+  log.info('\nFetching metadata types from steps...\n');
```
aiwilliams, line 43: How about `Collecting metadata types from steps...\n'` since the command is described as a collector?

packages/integration-sdk-cli/src/commands/types.ts
```diff
@@ -0,0 +1,154 @@
...
+    collectGraphObjectMetadataFromSteps(config.integrationSteps),
+  );
+
+  const metadataString = JSON.stringify(metadata, undefined, 2);
+
+  await fs.writeFile(typesFilePath, metadataString, {
+    encoding: 'utf-8',
+  });
+
+  log.info('Successfully generated types metadata file!');
```
aiwilliams, line 58: Should we point to where it was created as [the `visualize` command does](https://github.com/JupiterOne/sdk/blob/cca95a549d83c8096cfc50c671d6a30603738d4d/packages/integration-sdk-cli/src/commands/visualize.ts#L16)? This can provide a consistent experience with the CLI.

packages/integration-sdk-cli/src/commands/types.ts
```diff
@@ -0,0 +1,154 @@
...
+} from '@jupiterone/integration-sdk-core';
+import { promises as fs } from 'fs';
+
+export interface TypesCommandArgs {
+  projectPath: string;
+  typesFilePath: string;
+}
+
+export function types() {
+  return createCommand('types')
```
aiwilliams, line 21: I kinda wish this and `visualize` would output their content to the console unless provided the `-o, --output-path` option.

aiwilliams (commented):
This is an excellent feature, thanks for advancing it @ndowmon !

Now that we're adding additional commands, I think there is a need to consider whether there is room to improve consistency across their interfaces. For example:

* Output to standard out when there is no option provided to write to a file; this will allow for piping output to other programs
* Use `-o, --output-file` as the option for all commands that can write to a file
* Make another command like `table` that outputs the content that `document` is sticking into the document file currently, to separate that work and make the `document` command use it
* Make a command `graph` that takes sub commands, like `j1-integration graph types` and `j1-integration graph collected` and outputs HTML files; allow an option like `--format png`
* Make the `document` command update documentation file to include output of `table` and `graph types --format png`, and link it into the documentation that is generated by `document`
* Reconsider the name of `visualize` (see `graph` above) and add an option of `-o, --output-file`; move default location into command doc (currently the path/name of the file is inside the [`generateVisualization` function](https://github.com/JupiterOne/sdk/blob/cca95a549d83c8096cfc50c671d6a30603738d4d/packages/integration-sdk-cli/src/commands/visualize.ts#L15)

These can come in other PRs, and I'm willing to move this into a proposal Issue too, if others think these a good idea to pursue.

ndowmon:
@aiwilliams @charlieduong94 @austinkelleher Thanks for the feedback. I made a few major changes in response:

- Remove intermediate step that creates `types.json` file and commits to repository. This isn't actually used currently and it could confuse developers if it starts showing up in commits without reason.
- Add tests for `visualize-graph`
- Change file options to `-o, --output-file` for consistency in `document` and `visualize-graph` commands.

@aiwilliams I think a lot of your suggestions deserve more conversation, so I'll move them into a new SDK issue for CLI improvements.

packages/integration-sdk-cli/src/commands/document.ts
```diff
@@ -19,20 +19,20 @@ const J1_DOCUMENTATION_MARKER_START =
...
     .option(
       '-p, --project-path <directory>',
-      'Absolute file path to the integration project directory. Defaults to the current working directory.',
+      'Absolute path to the integration project directory. Defaults to the current working directory.',
       process.cwd(),
     )
     .option(
-      '-f, --documentation-file-path <path>',
-      'Absolute file path to the Markdown file that should be created/updated. Defaults to {CWD}/docs/jupiterone.md.',
+      '-o, --output-file <path>',
```
austinkelleher, line 34: Ya this is a great change. Technically a breaking change, so we'll want to major this.

austinkelleher (dismissed):
Awesome work!

aiwilliams (approved):

austinkelleher (approved):

austinkelleher (approved):

Issue #296, opened 2020-08-20
Force mapped relationship `targetFilterKeys` to be array of array of strings


ndowmon:
The `RelationshipMapping.targetFilterKeys` [property](https://github.com/JupiterOne/sdk/blob/master/packages/integration-sdk-core/src/types/relationship.ts#L35) can be easily misused by developers who define just one set of target filter keys for a mapped relationship.

```
createMappedRelationship({
  _class: ROLE_ASSIGNMENT_PRINCIPAL_RELATIONSHIP_CLASS,
  _mapping: {
    relationshipDirection: RelationshipDirection.FORWARD,
    sourceEntityKey: sourceEntity._key,
    targetEntity: {
      name: targetEntity.name,
      _type: targetEntity._type,
    },
    targetFilterKeys: ['name', '_type'],
  }
})
```

The `targetFilterKeys` above is valid and at first glance seems right, but will identify a mapped relationship by _either_ a matching `name` _or_ a matching `_type`. The proper usage for `targetFilterKeys` should be `[ ['name', '_type'] ]`, which will require _both_ values to match.

ndowmon:
For this issue, make sure both the sdk _and_ runtypes are updated.

aiwilliams:
@ndowmon, this is by design, but I agree that it can be risky. The restricted form you propose would still allow for `[['_key']]`, so it does seem safer. @philidem @erkangz what do you think of restricting `targetFilterKeys` to `string[][]`?

erkangz:
Looks good to me

PR #334, opened 2020-09-10
Fixes #333 & #296: Jest matcher optional `schema`

### LINKED ISSUE #295: Jest matcher `toMatchGraphObjectSchema` fails for non-array entity classes
Given an entity with a `string` classname (`_class`),
```
expect({
  _key: 'some-access-role-key',
  _type: 'provider_access_role',
  _class: 'AccessRole',
  name: 'access-role',
  displayName: 'access-role',
}).toMatchGraphObjectSchema({
  _class: 'AccessRole',
  schema: {
    additionalProperties: false,
    properties: {
      _type: { const: 'provider_access_role' },
      _key: { type: 'string' },
      _class: { const: 'AccessRole' },
      name: { type: 'string' },
      displayName: { type: 'string' },
    }
  }
})
```

the `toMatchGraphObjectSchema` matcher fails with the following
```
Error validating graph object against schema (data={
  "_key": "some-access-role-key",
  "_type": "provider_access_role",
  "_class": "AccessRole",
  "name": "access-role",
  "displayName": "access-role"
}, errors=[
  {
    "keyword": "const",
    "dataPath": "._class",
    "schemaPath": "#/properties/_class/const",
    "params": {
      "allowedValue": [
        "AccessRole"
      ]
    },
    "message": "should be equal to constant"
  }
], index=0)
```

Workaround: the matcher _will_ work correctly by setting the entity's `_class: ['AccessRole']` and changing the graph object schema `schema.properties._class: { type: 'array', items: { const: 'AccessRole' } }`.

### LINKED ISSUE #333: Make `schema` optional in `toMatchGraphObjectSchema`
When developers _don't_ provide additional schema to `toMatchGraphObjectSchema`, the interface can confuse developers or reviewers:

```typescript
expect(result).toMatchGraphObjectSchema({
  _class: ['DomainZone'],
  schema: {},
});
```

Making schema optional will remove some ambiguity from above.

ndowmon:
Fixes #333:

```typescript
expect(response).toMatchGraphObjectSchema({
  _class: ['Service'],
  // service: {}, # this is unnecessary now.
});
```

~~Also fixes #295:~~

```typescript
expect({
  _key: 'key',
  _type: 'type',
  _class: 'AccessRole',
  name: 'name',
  displayName: 'display-name',
}).toMatchGraphObjectSchema({
  _class: 'AccessRole',  // This would have failed previously because `_class` was cast to `[_class]`.
});
```

~~| Case # | Entity `_class` | Matcher `_class` | Pass/Fail |
|-|-|-|-|
| 1  | `AccessRole` | `AccessRole` | Pass |
| 1  | `[AccessRole]` | `[AccessRole]` | Pass |
| 1  | `[AccessRole]` | `AccessRole` | Fail |
| 1  | `AccessRole` | `[AccessRole]` | Fail |~~

~~In other words, the matcher `_class` needs to match the intended `_class` _exactly_.~~

**Backed out the changes for #295**


packages/integration-sdk-testing/src/jest.ts
```diff
@@ -183,14 +183,15 @@ export function toMatchGraphObjectSchema<T extends Entity>(
   { _class, schema }: ToMatchGraphObjectSchemaParams,
 ) {
   // Copy this so that we do not interfere with globals.
-  // NOTE: The data-model should actuall expose a function for generating
+  // NOTE: The data-model should actually expose a function for generating
   // a new object of the `IntegrationSchema`.
   const dataModelIntegrationSchema = dataModel.IntegrationSchema;
-  _class = Array.isArray(_class) ? _class : [_class];
+  const classArray = Array.isArray(_class) ? _class : [_class];
```
austinkelleher, line 189: I think there is other usage of `_class` that will no longer be normalized to an array (e.g. here: https://github.com/JupiterOne/sdk/blob/a9bf99cd267dd23cc4a06467913119248f0e3b41/packages/integration-sdk-testing/src/jest.ts#L221). Is that going to cause issues? I think that all of our tests currently assume an array because of that. Maybe this behavior is better, but might cause breaking behavior in our existing tests?

packages/integration-sdk/CHANGELOG.md
```diff
@@ -15,6 +15,12 @@ and this project adheres to
   environments to check whether an `Error` is of a type that need not be alerted
   to deployment operators. All `Error`s other than those only an end user can
   resolve should be reported.
+- Made `ToMatchGraphObjectSchemaParams.schema` property optional in jest
+  matcher.
+
+### Fixed
+
+- Fixed `toMatchGraphObjectSchema` matcher for non-array `_class` values.
```
aiwilliams, line 23: I think we've not been consistent, but want to have these refer to the issue number when they "Fixes" or "Closes" one.

ndowmon: Thanks for the suggestion! Done.

packages/integration-sdk-testing/src/jest.ts
ndowmon: Yes, it looks like it would. I baked these out of this commit so we just address #333.

austinkelleher (approved):

Issue #248, opened 2020-07-20
Axios 422 error when data size is too large


softwarewright:
When exporting data if data request is too large then the export will fail out. We need to handle the 422 error and dynamically change the limit size of the request when downloading the data.

PR #335, opened 2020-09-13
Fixing bulkdownload payload size issue

softwarewright:
Fixes https://github.com/JupiterOne/sdk/issues/248

### DIFF
--- a/packages/cli/src/export/__tests__/utils/createRelationship.ts
+++ b/packages/cli/src/export/__tests__/utils/createRelationship.ts
@@ -1,7 +1,9 @@
 import {
   Entity,
-  createDirectRelationship,
 } from '@jupiterone/integration-sdk-core';
+import {
+  createDirectRelationship
+} from '@jupiterone/integration-sdk-core/src';
 import { RelationshipClass } from '@jupiterone/data-model';
 
 interface CreateRelationshipParams {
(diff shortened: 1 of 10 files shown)


packages/cli/src/export/__tests__/utils/createRelationship.ts
```diff
@@ -1,7 +1,9 @@
 import {
   Entity,
-  createDirectRelationship,
 } from '@jupiterone/integration-sdk-core';
+import {
+  createDirectRelationship
+} from '@jupiterone/integration-sdk-core/src';
```
softwarewright, line 6: Not sure why the direct import from `@jupiterone/integration-sdk-core` stopped working for `createDirectRelationship` in this project. Whenever I import the `createDirectRelationship` without the ending `/src` on `@jupiterone/integration-sdk-core`  the function comes in undefined. I'm still trying to understand why this happens if anyone has any ideas.

softwarewright: -_- Yup that was it, thanks Charlie.

austinkelleher (dismissed):
Thanks for fixing this!

charlieduong94 (dismissed):

charlieduong94 (approved):

PR #337, opened 2020-09-14
Only allow primitives & arrays on mapped relationship target entities

ndowmon:
In the Artifactory integration, a call was made to `createMappedRelationship` with an actual persisted entity, and should have instead been made to `createDirectRelationship`. This was not noticed in code review, and ended up making it into production. 

In production, the integration failed with:

```
(1:38:55 PM) - [job_abort] - Aborting integration job. (reason=Error uploading collected data (API response: code=RELATIONSHIP_UPLOAD_FAILED_VALIDATION, message="2 validation error(s): - Error in relationships[0]._mapping.targetEntity._rawData (reason=Expected string | number | boolean | string[] | number[] | boolean[], but was object in _mapping.targetEntity._rawData) - Error in relationships[1]._mapping.targetEntity._rawData (reason=Expected string | number | boolean | string[] | number[] | boolean[], but was object in _mapping.targetEntity._rawData) (code=RELATIONSHIP_UPLOAD_FAILED_VALIDATION, statusCode=400,
```

The typing for target entities expected `Partial<Entity>`, which was not strict enough to generate entities that pass persister validation.

This change introduces `PrimitiveEntity` and uses `Partial<PrimitiveEntity>` as the typing for target entities in mapped relationships, essentially omittig the `_rawData` property & data type from acceptable property types. This should guard developers from making this same mistake in the future.

For the record, the following used to pass type checking, but now fails:

```typescript
createMappedRelationship({
  _class: RelationshipClass.HAS,
  source: {
    _key: 'some-key',
    _class: 'EntityClass',
    _type: 'some-type',
  },
  target: {
    _rawData: [{
      name: 'default',
      rawData: {
        key: 'mapped relationship targets cannot have objects as property values!'
      }
    }]
  }
})
```

### DIFF
--- a/packages/integration-sdk-cli/src/visualization/createMappedRelationshipNodesAndEdges.ts
+++ b/packages/integration-sdk-cli/src/visualization/createMappedRelationshipNodesAndEdges.ts
@@ -1,20 +1,23 @@
-import { Node, Edge, NodeOptions } from "vis";
-import { 
+import { Node, Edge, NodeOptions } from 'vis';
+import {
   Entity,
   MappedRelationship,
   RelationshipDirection,
   RelationshipMapping,
   TargetEntityProperties,
-} from "@jupiterone/integration-sdk-core";
-import { isMatch, pick } from "lodash";
-import { getNodeIdFromEntity, NodeEntity, isNodeIdDuplicate } from "./utils";
+} from '@jupiterone/integration-sdk-core';
+import { isMatch, pick } from 'lodash';
+import { getNodeIdFromEntity, NodeEntity, isNodeIdDuplicate } from './utils';
 
 interface MappedRelationshipNodesAndEdges {
   mappedRelationshipNodes: Node[];
   mappedRelationshipEdges: Edge[];
 }
 
-export function isClassMatch(entityClasses: string | string[] | undefined, targetEntityClasses: string | string[] | undefined): boolean {
+export function isClassMatch(
+  entityClasses: string | string[] | undefined,
+  targetEntityClasses: string | string[] | undefined,
+): boolean {
   if (targetEntityClasses === undefined) {
     return true;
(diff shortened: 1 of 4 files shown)


aiwilliams (approved):
I think this is a good change. We may run into a scenario where some integrations are constructing entities and later (in the code) deciding they are the target of a mapped relationship. In those cases, they may be forced to refactor the code to avoid creating the target before deciding what type of relationship will be made. That's a good thing I think, because it isn't obvious that an entity created before that decision will not be managed by the integration, nor have `_rawData` persisted for it.

ndowmon:
@aiwilliams I'm not sure I'm completely following... under what circumstances would we create/persist an entity, but make that entity a `targetEntity` in a mapped relationship? You're not talking about that being a `sourceEntity` with `RelationshipDirection.REVERSE`, are you?

aiwilliams:
> under what circumstances would we create/persist an entity, but make that entity a `targetEntity` in a mapped relationship

That shouldn't happen. I'm referring to a scenario where a developer codes like this:

```ts
const entity = createIntegrationEntity(...);

// ....
// lots of lines, reader isn't aware the entity will be used in a mapped
// relationship because they think createIntegrationEntity() means create
// one that is owned by this instance, not used in a mapped relationship
// ....

if (someThing) {
  await jobState.addRelationship(createDirectRelationship(..., to: entity));
} else {
  await jobState.addRelationship(createMappedRelationship(..., target: entity));
}
```

I think it would be better to avoid that kind of thing. Since `entity._rawData` will exist, I would hope that `createMappedRelationship(..., target: entity)` will not lint for those integrations now, and they'll need to be refactored.

Does that clarify?

ndowmon:
Ah yes, I see what you're saying. I agree that the need to refactor is good.

charlieduong94 (approved):

PR #345, opened 2020-09-29
Add dotenv/config to Jest config in dev-tools

ndowmon:


austinkelleher (commented):
I think there is value in this change, but I personally avoided doing this in the Google Cloud integration in favor of making the testing strategy against the real environment a different npm script command.

Testing purely locally:

```
yarn test
```

Testing against the real environment:

```
yarn test:env
```

This stems from my personal preference that every open source project should ship with a local testable path. Doing the following validates that I have a project set up correctly and ready to develop against:

```
git clone ...
npm install
npm test
```

Thoughts @aiwilliams @ndowmon?

aiwilliams:
I was thinking this would be added to the template project, but I suppose it makes sense to add it here once we're sure it won't be a problem.

> Doing the following validates that I have a project set up correctly and ready to develop against

Completely agree, the setup process you've outlined should just work in every project! I wish that were true in our internal projects as well, with expensive tests that require Docker, localstack, etc. NOT executed in `yarn test`.

> every open source project should ship with a local testable path

I don't think I fully understand this yet for the open source integrations and the `dotenv/config`, so please bear with me, knowing I agree with your goal as stated above.

When `dotenv/config` is required, it will do nothing if there is no `.env` (I'm sure you're aware), the default state of a newly cloned repository. The tests do need to be written to work without a `.env` loaded, something like:

```ts
const config: IntegrationConfig = {
  clientId: process.env.CLIENT_ID || 'clientId',
  clientSecret: process.env.CLIENT_SECRET || 'clientSecret',
  someConfigData: process.env.SOME_CONFIG_DATA || 'yo,dog'
};
```

I see the Google Cloud integration [does something similar](https://github.com/JupiterOne/graph-google-cloud/blob/master/test/config.ts).

Often, values like `someConfigData` are replaced in specific tests or should default to values used in the `.env` when the recordings were made. The `.env` ends up primarily driving the secrets. Secrets should be redacted from recordings; the actual values aren't required for playback. I've seen three or four recently created integrations do some kind of `dotenv/config` loading in the tests to get the secrets for collecting the recordings.

So, I have a hard time seeing where there's a problem other than all the integrations I've worked on/recently reviewed want to require `dotenv/config` for `yarn test`.

austinkelleher:
@aiwilliams Yes, agreed. But how do we run our tests locally such that we can test exactly what the CI will test without having to delete the `.env` file or manually setting the environment variables to undefined? I have see on multiple occasions a test pass locally because of some local environment setup where we load the environment variables. Then when the tests run in the CI, it fails. I like to be able to test against the non-local environment and exclusively against what is in the recordings (e.g. fully offline testing).

I don't really care what the default behavior is as long as we have a way to do fully offline testing that will not be interfered with by environment variables. This is why I distinguish the difference between online and offline testing strategy in the GC integration using a specific environment variable that is set in the npm script as you noted. 

Something like this would also work though:

```json
{
  "scripts": {
    "test": "...", // Online testing by default
    "test:ci": "DISABLE_INTEGRATION_ENVIRONMENT_CONFIG=true yarn test" // Offline testing. Can run locally.
  }
}
```

ndowmon:
@aiwilliams @austinkelleher Ok, I see what you both are saying - I definitely see the value in allowing users to locally either use/not use environment variables. A _sort of_ similar pattern was required with the new Rapid7 integration:

```typescript
{
  "scripts": {
    "test": "jest",
    "test:local": "NODE_TLS_REJECT_UNAUTHORIZED=0 && yarn test",
    ...
  }
  ...
}
```

I see the value, and also agree with the suggestion of putting this into the `integration-template`.

If we're in agreement, I'll be happy to close this and open one in `integration-template`.

Issue #228, opened 2020-06-10
Validate direct relationships before upload


aiwilliams:
It is possible to create relationships between entities where one side or the other is not actually present in the collected set, which may be surfaced by the persister as a warning log event, but is really an error situation we want to expose as early in development as possible. Only mapped relationships can refer to an entity that isn't collected in the run.

Provide a command such as `j1-integration validate --direct-relationships` which walks relationships and verifies that their from/to `_key` values point to entities that exist in the current collection.

This will likely require #79 is advanced so that entities can be found quickly by `_type` & `_key`.

ndowmon:
The `getEntity()` function recently implemented ought make this part a bit easier:

> This will likely require #79 is advanced so that entities can be found quickly by _type & _key.

aiwilliams:
@ndowmon I think we need to reconsider the API of `getEntity`, to make it take nothing but a `_key` value. Consider how https://github.com/JupiterOne/sdk/blob/7eadfcece14ee9330297db07c2952e6af0c04675/packages/integration-sdk-runtime/src/execution/jobState.ts#L15 prevents duplicate keys across all values of `_type`. This is possible because an integration instance is a scope of keys. What do you think?

cc/ @austinkelleher

PR #346, opened 2020-09-29
Fixes #344: Direct relationship jest matcher

### LINKED ISSUE #344: Add `toMatchRelationshipSchema()` matcher
The `.toMatchRelationshipSchema()` matcher should ensure that all properties are non-arrays or objects.

ndowmon:


packages/integration-sdk-testing/src/jest.ts
```diff
@@ -222,12 +226,68 @@ export function toMatchGraphObjectSchema<T extends Entity>(
...
+      {
+        $ref: '#GraphObject',
+      },
+      {
+        additionalProperties: {
+          type: ['boolean', 'integer', 'null', 'number', 'string'],
+        },
+      },
+    ],
+  };
```
ndowmon, line 267: There is no `#Relationship` schema in the data model, but this defines what it _would_ look like.

The tricky part will be figuring how to share logic between `ExplicitRelationship` and `MappedRelationship`. Not sure if we will need one JSON schema for each or what.

packages/integration-sdk-testing/src/jest.ts
```diff
@@ -238,12 +298,38 @@ export function toMatchGraphObjectSchema<T extends Entity>(
...
+    return {
+      message: () => 'Success!',
+      pass: true,
+    };
+  } else {
+    return hasDistinctKeysResult;
+  }
+}
+
+function toHaveDistinctKeys(received: { _key: string }[]) {
```
austinkelleher, line 313: This is minor, but I think we could improve this function by not returning a the jest matcher result format and instead return only a boolean. That will make the function more pure and the caller can decide what to do with the resulting boolean. Also I think we can improve the TS typings. Something like the following (untested):

```ts
function graphObjectHasDistinctKeys<
  T extends Entity | ExplicitRelationship
>(received: T[]): boolean {
  const _keys = received.map((r) => r._key);
  return Array.isArray(_keys) && new Set(_keys).size === _keys.length;
}
```

Caller:

```ts
if (graphObjectHasDistinctKeys(received)) {
  return {
    message: () => 'Success!',
    pass: true,
  };
} else {
  return {
    message: () =>
      `Object \`_key\` properties array is not unique: [${_keys}]`,
    pass: false,
  };
}
```

austinkelleher (approved):
This is awesome! Thanks a lot for taking this on. I left a minor code improvement comment, but approving anyway.

packages/integration-sdk-testing/src/jest.ts
ndowmon: Good point. I had originally considered exposing the `toHaveDistinctKeys()` matcher from the SDK but now I'm not sure that makes sense.

ndowmon: I'll have to rebase this back off of master (it's branched off of `add-dotenv-config-to-jest-config` which it seems like I'll close without merging). I think it's a good suggestion and worth doing while I'm doing the rebase.

austinkelleher (approved):
Excellent!

aiwilliams:
Related to #228

PR #348, opened 2020-10-01
Fix createIntegrationEntity() when status != 'Online'|'Active'

ndowmon:


packages/integration-sdk-core/src/data/createIntegrationEntity.ts
```diff
@@ -164,7 +164,18 @@ function generateEntity({
...
+    );
+    const isInactive = new RegExp('inactive|disabled|offline', 'i').test(
+      source.status,
+    );
+
+    entity.active = isActive
+      ? true // if
+      : isInactive
+      ? false // else if
+      : undefined; // else
```
ndowmon, line 178: Hate the way eslint fixed this... I had it as

```typescript
entity.active = (
  isActive ? true : // if 
  isInactive ? false : // else if 
  undefined; // else 
);

aiwilliams (approved):
I think we had hopes of removing these special cases, but I think it good to fix it while it exists!

ndowmon:
Needed to fix a merge conflict; this needs a re-approval.

erkangz (dismissed):

austinkelleher (dismissed):

austinkelleher (approved):

PR #353, opened 2020-10-12
Support controlling syncMode with invocationConfig

aiwilliams:
The tests need to be updated, but this proposed change allows an integration to control the sync mode. Please let me know if this approach looks sound. My understanding of the persister code is that we must provide the sync mode when initializing the synchronization job. I'll get the `sync` command tests passing if the approach looks good. I'll also need to make a change to `managed-integration-runtime` package I think to make it invoke `invocationConfig.syncMode()`.

The Qualys integration is going to need to be able to upload new data in a rolling fashion, without deleting old data. Older integrations such as the Jira and Bitbucket integration already do this kind of thing by sending only CREATE operations, so we'll need this for them to update to the new SDK at some point as well.

### DIFF
--- a/packages/integration-sdk-cli/src/commands/run.ts
+++ b/packages/integration-sdk-cli/src/commands/run.ts
@@ -1,29 +1,31 @@
 import { createCommand } from 'commander';
 
-import * as log from '../log';
+import { Metric } from '@jupiterone/integration-sdk-core';
 import {
-  getApiKeyFromEnvironment,
-  getApiBaseUrl,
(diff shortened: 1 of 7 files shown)


erkangz (approved):

austinkelleher:
Looks like the tests are failing. Looks like a good change though!

aiwilliams:
We're moving to an approach where the integration can declare partial data sets at the beginning of the synchronization job, to allow the persister to begin processing immediately as data is uploaded. Related to #372 #371

PR #356, opened 2020-10-13
Publish disk-usage on interval

austinkelleher:
This will help show us how quickly disk usage is growing for an integration run.

### DIFF
--- a/packages/integration-sdk-cli/src/__tests__/cli-run-failure.test.ts
+++ b/packages/integration-sdk-cli/src/__tests__/cli-run-failure.test.ts
@@ -79,7 +79,7 @@ test('aborts synchronization job if an error occurs', async () => {
   });
 });
 
-test.skip('does not log errors that have been previously logged', async () => {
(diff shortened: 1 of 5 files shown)


packages/integration-sdk-cli/src/__tests__/cli-run-failure.test.ts
```diff
@@ -79,7 +79,7 @@ test('aborts synchronization job if an error occurs', async () => {
   });
 });
 
-test('does not log errors that have been previously logged', async () => {
+test.skip('does not log errors that have been previously logged', async () => {
```
austinkelleher, line 82: Will revisit this. It isn't actually broken...

austinkelleher:
I'd like to abstract all of this into the `FileSystemGraphObjectStore`, but it will require more refactoring across the codebase. There are quite a few assumptions throughout the code that we are using the file system to write out entities and relationships.

aiwilliams (approved):
Great work making sure the functions only work when there is a directory and while the integration is executing.

PR #359, opened 2020-10-16
Add link to data-model when jest matchers fail to validate schema

ndowmon:
I think it can be hard for external developers (or candidates working on coding challenges) to know where to look when `toMatchGraphObjectSchema()` matchers throw errors about invalid classes/schemas. This change is meant to make it easier for devs to know where to look for our data model schemas.


austinkelleher (approved):
Nice! It would be great if the URL could actually include links to the specific JSON file(s) that failed validation.

aiwilliams (approved):
Sweet!

PR #363, opened 2020-10-23
362 failure callback in logger

### LINKED ISSUE #362: Allow `IntegrationLogger` to expose external handler for step failure
The `executeIntegrationInstance` function continues step execution, even in the event of a step failure, unless the error contains the `fatal: true` property. This prevents external error handling systems to capture some errors encountered within steps. The SDK's `IntegrationLogger` should expose an external error handling callback that allows non-fatal encountered errors to be handled specifically by the calling function.

ndowmon:
Fixes #362 

Also removed the `step` property from logs, since there is already a `stepId` property attached in every step context: 
https://github.com/JupiterOne/sdk/blob/master/packages/integration-sdk-runtime/src/execution/dependencyGraph.ts#L314

### DIFF
--- a/packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
+++ b/packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
@@ -268,20 +268,8 @@ describe('step event publishing', () => {
     logger.stepSuccess(step);
 
     expect(infoSpy).toHaveBeenCalledTimes(2);
-    expect(infoSpy).toHaveBeenNthCalledWith(
-      1,
-      {
-        step: step.id,
-      },
-      `Starting step "Mochi"...`,
-    );
-    expect(infoSpy).toHaveBeenNthCalledWith(
-      2,
-      {
-        step: step.id,
-      },
-      `Completed step "Mochi".`,
-    );
+    expect(infoSpy).toHaveBeenNthCalledWith(1, {}, `Starting step "Mochi"...`);
+    expect(infoSpy).toHaveBeenNthCalledWith(2, {}, `Completed step "Mochi".`);
 
     const error = new IntegrationLocalConfigFieldMissingError('ripperoni');
     logger.stepFailure(step, error);
@@ -292,7 +280,6 @@ describe('step event publishing', () => {
       {
         err: error,
         errorId: expect.any(String),
-        step: step.id,
       },
       expect.stringContaining(
         `Step "Mochi" failed to complete due to error. (errorCode="${error.code}"`,
--- a/packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
+++ b/packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
@@ -667,3 +667,68 @@ describe('#publishErrorEvent', () => {
     });
   });
 });
+
+describe('#handleFailure', () => {
+  test('should invoke callback on handleFailure', () => {
+    const logger = createIntegrationLogger({
+      name,
+      invocationConfig,
+    });
+
+    const errorSpy = jest.spyOn(logger, 'error');
+    const publishEventSpy = jest.spyOn(logger, 'publishEvent');
+
+    const onFailureSpy = jest.fn();
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    // Property 'onFailure' is private and only accessible within class 'IntegrationLogger'.ts(2341)
+    logger.onFailure = onFailureSpy;
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    // Property 'handleFailure' is private and only accessible within class 'IntegrationLogger'.ts(2341)
+    logger.handleFailure({
+      err: new Error(),
+      errorId: 'SOME_ERROR_ID',
+      eventName: 'SOME_EVENT',
+      description: 'an error :(',
+    });
+
+    expect(onFailureSpy).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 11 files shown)


packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -260,14 +271,12 @@ export class IntegrationLogger extends EventEmitter
...
-    const name = 'step_failure';
+    const eventName = 'step_failure';
     const { errorId, description } = createErrorEventDescription(
       err,
       `Step "${step.name}" failed to complete due to error.`,
     );
-
-    this.error({ errorId, err }, description);
-    this.publishEvent({ name, description });
+    this.handleFailure({ eventName, errorId, err, description });
```
ndowmon, line 279: I originally removed the unnecessary `step` log property so that I could create a single consistent interface, sharing logic for both `stepFailure` and `validationFailure`.

packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -248,26 +259,24 @@ export class IntegrationLogger extends EventEmitter
   stepStart(step: StepMetadata) {
     const name = 'step_start';
     const description = `Starting step "${step.name}"...`;
-    this.info({ step: step.id }, description);
+    this.info({}, description);
```
aiwilliams, line 262: There is no need to pass in an empty object, it can be just `info(description)`.

packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -295,19 +304,29 @@ export class IntegrationLogger extends EventEmitter
...
-    const name = 'validation_failure';
+    const eventName = 'validation_failure';
     const { errorId, description } = createErrorEventDescription(
       err,
       `Error occurred while validating integration configuration.`,
     );
+    this.handleFailure({ eventName, errorId, err, description });
+  }
 
+  private handleFailure(options: {
```
aiwilliams, line 315: I think that if we're going to go generic and encourage switching on `eventName` in the `handleFailure` function, we should make the `eventName` a fixed set of values, such as an `enum`.

packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
```diff
@@ -680,3 +667,68 @@ describe('#publishErrorEvent', () => {
...
+    });
+
+    expect(onFailureSpy).toHaveBeenCalledTimes(1);
+    expect(publishEventSpy).toHaveBeenCalledTimes(1);
+
+    // should call error
+    expect(errorSpy).toHaveBeenCalledTimes(1);
+  });
+
+  test('should invoke callback on handleFailure', () => {
```
aiwilliams, line 704: This is the same test description as the previous test.

packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
```diff
@@ -680,3 +667,68 @@ describe('#publishErrorEvent', () => {
...
+    const onFailureSpy = jest.fn();
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    // Property 'onFailure' is private and only accessible within class 'IntegrationLogger'.ts(2341)
+    logger.onFailure = onFailureSpy;
+
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    // Property 'handleFailure' is private and only accessible within class 'IntegrationLogger'.ts(2341)
+    logger.handleFailure({
```
aiwilliams, line 722: Did you mean to test that another function will invoke the `onFailure` function?

aiwilliams (changes requested):

packages/integration-sdk-runtime/src/logger/index.ts
ndowmon: Thanks, that's a great idea. I went with string union instead of enum since it's very specific to this interface (at the moment)

packages/integration-sdk-runtime/src/logger/index.ts
ndowmon: 👍

packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
ndowmon: Meant to test the difference between an error that satisfies `shouldReportErrorToOperator`. Fixed.

packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
ndowmon: No, but that's a good idea. Added two tests to ensure `stepFailure` and `validationFailure` both invoke `handleFailure`.

packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
aiwilliams: Looking at the expectations later in the functions, I see the differences to be essentially

```
if error for operator {
  (expect(errorSpy).toHaveBeenCalledTimes(1)
} else {
  expect(warnSpy).toHaveBeenCalledTimes(1);
}
```

Perhaps there could be three tests and descriptions could be:

* `#handleFailure should invoke onFailure`
* `#handleFailure should log as warn when error is meant to be handled by a user`
* `#handleFailure should log as error when error is meant to be handled by an operator`

aiwilliams (approved):
I left some comments that I think could help improve the clarity of the tests.

ndowmon:
Thanks @aiwilliams great suggestions. I updated and versioned

aiwilliams (approved):

PR #364, opened 2020-10-26
Expose onFailure callback in createIntegrationLogger

ndowmon:
Add-on to #363. Unfortunately, that PR did not expose the callback in the `createIntegrationLogger` interface.

### DIFF
--- a/packages/integration-sdk-runtime/src/logger/index.ts
+++ b/packages/integration-sdk-runtime/src/logger/index.ts
@@ -40,6 +40,7 @@ interface CreateLoggerInput<
   invocationConfig?: InvocationConfig<TExecutionContext, TStepExecutionContext>;
   pretty?: boolean;
   serializers?: Logger.Serializers;
+  onFailure?: OnFailureFunction;
 }
 
 interface CreateIntegrationLoggerInput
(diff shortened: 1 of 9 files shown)


packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -40,6 +40,7 @@ interface CreateLoggerInput<
   invocationConfig?: InvocationConfig<TExecutionContext, TStepExecutionContext>;
   pretty?: boolean;
   serializers?: Logger.Serializers;
+  onFailure?: OnFailureFunction;
```
austinkelleher, line 43: Will this be considered a fatal error? If so, we should consider renaming this to something like `onFatalFailure` for clarity.

austinkelleher (dismissed):

packages/integration-sdk-runtime/src/logger/index.ts
ndowmon: No, this is for non-fatal failures. Fatal errors terminate the process and throw an error in the calling process, which is currently being handled just fine by the managed integration runtime. It's non-fatal errors that are handled within this package which drives the need for this callback.

austinkelleher (approved):

PR #365, opened 2020-10-27
Support tokens for authorization

fomentia:
We should be able to use the old JWT user keys and the new service tokens for authorization. Thus, we cannot depend on having the account ID embedded in the access token. This change to the API client interface requires callers to provide an account ID and an access token instead of only a JWT user key. Since a public helper function is being removed, this will require a major version bump.

### DIFF
--- a/packages/cli/src/__tests__/cli-export.test.ts
+++ b/packages/cli/src/__tests__/cli-export.test.ts
@@ -194,13 +194,18 @@ test('should log error when export fails', async () => {
 
 test('should throw error when missing api key', async () => {
   await expect(
-    createCli().parseAsync(['node', 'j1', 'export']),
+    createCli().parseAsync([
+      'node',
+      'j1',
+      'export',
+      `--account=${TEST_ACCOUNT}`,
+    ]),
   ).rejects.toThrow(
     /Missing option! Set the JUPITERONE_API_KEY environment variable or supply the --api-key option./,
   );
 });
 
-test('should throw error when missing api key', async () => {
+test('should throw error when missing account', async () => {
   await expect(
     createCli().parseAsync([
       'node',
--- a/packages/cli/src/__tests__/cli-export.test.ts
+++ b/packages/cli/src/__tests__/cli-export.test.ts
@@ -3,7 +3,7 @@ import { mocked } from 'ts-jest/utils';
 import * as runtime from '@jupiterone/integration-sdk-runtime';
 
 import { createCli } from '..';
-import { TEST_API_KEY } from './utils';
+import { TEST_API_KEY, TEST_ACCOUNT } from './utils';
 import * as log from '../log';
 import { vol } from 'memfs';
 import { createEntity, createRelationship } from '../export/__tests__/utils';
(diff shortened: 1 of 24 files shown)


packages/cli/src/__tests__/cli-export.test.ts
```diff
@@ -192,6 +196,19 @@ test('should throw error when missing api key', async () => {
   await expect(
     createCli().parseAsync(['node', 'j1', 'export']),
   ).rejects.toThrow(
-    /Missing apiKey! Set the JUPITERONE_API_KEY environment variable or supply the --api-key option./,
+    /Missing option! Set the JUPITERONE_API_KEY environment variable or supply the --api-key option./,
+  );
+});
+
+test('should throw error when missing api key', async () => {
```
ndowmon, line 203:
```suggestion
test('should throw error when missing account', async () => {
```

fomentia:
@ndowmon This test can intermittently fail since the uuid generated for the relationship when it is a duplicate can include "456": https://github.com/JupiterOne/sdk/blob/ccc8e01847896ad5790f5ab29b711ec3d794ac23/packages/integration-sdk-cli/src/visualization/__tests__/createMappedRelationshipNodesAndEdges.test.ts#L369

packages/cli/src/__tests__/cli-export.test.ts
aiwilliams: This looks to still be checking the api key, and there is another expectation below that provides the `--api-key` and throws for the missing `--acount`.

packages/cli/src/commands/j1Export.ts
```diff
@@ -23,6 +24,10 @@ export function j1Export() {
       'The directory where entities/relationships will be downloaded',
       DEFAULT_EXPORT_DIRECTORY,
     )
+    .option(
+      '--account <account>',
+      'The JupiterOne account you are importing entities/relationships into',
```
aiwilliams, line 29: This is `j1Export`, seems this should talk in terms of exporting, not importing.

aiwilliams (changes requested):
Please update the `CHANGELOG` and include as part of this PR.

aiwilliams (approved):

fomentia:
@ndowmon The build keeps failing due to the faulty test I previously mentioned. I'm not totally sure what the best way to fix it is.

```
FAIL packages/integration-sdk-cli/src/visualization/__tests__/createMappedRelationshipNodesAndEdges.test.ts
  ● #createMappedRelationshipNodesAndEdges › should return DUPLICATE KEY when duplicate non-matching entity found

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Array [
        Object {
          "dashes": true,
          "from": "123",
          "label": "HAS",
    -     "to": StringNotMatching /456/,
    +     "to": "6e6145d9-9707-418c-ab6a-9b9a9707a456",
        },
      ]

      364 |     } = createMappedRelationshipNodesAndEdges({mappedRelationships, explicitEntities});
      365 | 
    > 366 |     expect(mappedRelationshipEdges).toEqual([
          |                                     ^
      367 |       {
      368 |         from: '123',
      369 |         to: expect.not.stringMatching('456'),

      at Object.<anonymous> (src/visualization/__tests__/createMappedRelationshipNodesAndEdges.test.ts:366:37)
```

PR #367, opened 2020-10-29
Pass `onFailure` to children of `IntegrationLogger`

### LINKED ISSUE #362: Allow `IntegrationLogger` to expose external handler for step failure

ndowmon:
Fixes #362 

Looks like this will need a major bump after this.

### DIFF
--- a/packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
+++ b/packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
@@ -669,7 +669,7 @@ describe('#publishErrorEvent', () => {
 });
 
 describe('#handleFailure', () => {
-  test.only('should maintain onFailure implementation when calling logger.child()', () => {
+  test('should maintain onFailure implementation when calling logger.child()', () => {
     const onFailureSpy = jest.fn();
     const logger = createIntegrationLogger({
       name,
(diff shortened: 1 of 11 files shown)


packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
```diff
@@ -669,16 +669,32 @@ describe('#publishErrorEvent', () => {
 });
 
 describe('#handleFailure', () => {
-  test('should invoke onFailure and publishEvent', () => {
+  test.only('should maintain onFailure implementation when calling logger.child()', () => {
```
austinkelleher, line 672: Did you intend to leave the `.only`?

ndowmon: No I didn't. Good catch!

austinkelleher (approved):

austinkelleher (approved):

PR #370, opened 2020-11-03
Update build.yml

aiwilliams:
We're reducing the number of github actions because they cost money, and there are platforms/versions that aren't critical. Going to keep tests running on Windows for this because there are file path things we care about. Please approve an merge.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,8 +7,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [10.x, 12.x]
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        node-version: [12.x]
+        os: [ubuntu-latest, windows-latest]
 
     steps:
       - id: setup-node


.github/workflows/build.yml
```diff
@@ -7,8 +7,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node-version: [10.x, 12.x]
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        node-version: [12.x]
+        os: [ubuntu-latest, windows-latest]
```
ndowmon, line 11:
```suggestion
        os: [ubuntu-latest]
```

ndowmon (changes requested):

.github/workflows/build.yml
aiwilliams: The windows one is important I think because we support execution on windows, and there is code in the SDK dealing with path differences between the platforms.

ndowmon: Thanks for the explanation!

ndowmon (approved):

PR #371, opened 2020-11-07
Allow steps to declare always partial datasets

aiwilliams:
This change would allow an integration step to declare that it will always produce partial datasets. For example, a step that ingests Findings would declare that it is not ingesting the complete set, or one that ingests PRs, etc.

Some tests could be added, but I think it good to validate the approach before spending more time on it.


fomentia (approved):

aiwilliams:
@philidem I think the thing remaining here is to include the integration config partial datasets info when initiating the synchronization job, but I think we still want to include them in the summary, right?

philidem:
> @philidem I think the thing remaining here is to include the integration config partial datasets info when initiating the synchronization job, but I think we still want to include them in the summary, right?

What do you mean by "include them in the summary"?

aiwilliams:
> What do you mean by "include them in the summary"?

The call to finalize a job includes the complete list of partial datasets https://github.com/JupiterOne/sdk/blob/47e5b5484cc712831e28df8e8981391489b52920/packages/integration-sdk-runtime/src/synchronization/index.ts#L54

aiwilliams:
Talked with @philidem IRL, it makes sense to include it in the finalization call. The persister will consider it additive to the initialization set of partial types and essentially be insensitive to any repetition.

aiwilliams:
After further consideration and discussion with @philidem, we are NOT going to send any partial dataset info to the persister when initiating the synchronization job, because the persister is NOT going to process data as it comes in at this time. However, this PR is still good as it allows the step to declare the types its processing which are always to be considered partial datasets. This does mean that the integration will have no means of deleting entities/relationships of these types.

packages/integration-sdk-core/src/types/step.ts
```diff
@@ -97,6 +100,21 @@ export interface StepGraphObjectMetadataProperties {
...
+   * This is useful in steps that ingest an ever growing list of entities and
+   * their relationships, ensuring the synchronization system does not delete
+   * older data that will not be seen as ingestion progresses through the list
+   * over numerous executions. For example:
+   *
+   * * Vulnerability findings
+   * * SCM pull requests
+   * * Ticket systems
+   */
+  partialDatasets?: PartialDatasets[];
```
ndowmon, line 117: I'm actually wondering if this information should be an optional property on `StepEntityMetadata` and `StepRelationshipsMetadata`. That way, this information can easily make its way into automated documentation in the future.

packages/integration-sdk-runtime/src/execution/__tests__/executeIntegration.test.ts
```diff
@@ -341,6 +341,164 @@ describe('executeIntegrationInstance', () => {
...
+            {
+              types: ['test_b2'],
+            },
+          ],
+          status: StepResultStatus.PARTIAL_SUCCESS_DUE_TO_DEPENDENCY_FAILURE,
+        },
+      ],
+      metadata: {
+        partialDatasets: {
+          types: ['test_a', 'test_b2', 'test_b'],
```
ndowmon, line 417: I don't think you have a test about this but I just want to be clear - if `step-a` has a partial dataset, but is successful, it still has `StepResultStatus.SUCCESS`, correct? and `step-b`, which depends on `step-a`, will be marked as `StepResultStatus.SUCCESS` as opposed to `StepResultStatus. PARTIAL_SUCCESS_DUE_TO_DEPENDENCY_FAILURE`?

charlieduong94: Def would be good to add a test for the `SUCCESS` case.

aiwilliams: That's correct @ndowmon, added a test 👍

packages/integration-sdk-core/src/types/step.ts
aiwilliams: I think this is a great idea. I'll take a stab at reworking it....

ndowmon (approved):
This is really nice. Thanks @aiwilliams !

packages/integration-sdk/CHANGELOG.md
```diff
@@ -14,10 +14,11 @@ and this project adheres to
 - `ExecutionContext.history?: ExecutionHistory` provides information about the
   `lastExecution` and `lastSuccessfulExecution`. Integrations may use this to
   limit data ingestion.
-
 - Allow `executeIntegrationInstance` to take a custom `GraphObjectStore`
 - Expose `BucketMap` and `FileSystemGraphObjectStore` from
   `@jupiterone/integration-sdk-runtime`
+- Steps may declare `partialDatasets` in their metadata to indicate that they
+  will never ingest a complete set of some `_type`s of entities/relationships.
```
aiwilliams, line 21: D'oh, missed updating this, will get in CHANGELOG for release PR.

PR #372, opened 2020-11-07
Upload entities and relationships continuously during ingestion

aiwilliams:
After reviewing the code, I think the easiest way to get this feature working is through the `GraphObjectStore` interface, which the `JobState` implementation interacts to get entities and relationships out of memory. I am thinking to have an implementation of the `GraphObjectStore` that wraps a `FileSystemGraphObjectStore` and a reference to an initiated `SynchronizationJob` and something like a `PersisterService`. When the `GraphObjectStore.flush()` function is invoked (including threshold-induced invocations), it will queue up requests to upload the data.

The managed runtime would use this this all the time, the CLI could be left unchanged or we could easily support an option like `--continuous-upload`.

Please let me know if you think this approach is sound and I'll continue to push it forward.

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/__tests__/dependencyGraph.test.ts
+++ b/packages/integration-sdk-runtime/src/execution/__tests__/dependencyGraph.test.ts
@@ -6,25 +6,29 @@ import waitForExpect from 'wait-for-expect';
 
 import {
   Entity,
-  Relationship,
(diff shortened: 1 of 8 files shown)


fomentia:
This seems like a good approach to me! 👍🏻

austinkelleher (approved):

PR #373, opened 2020-11-13
Refactor tests in executeIntegration.test.ts

austinkelleher:
Making some additional changes to the SDK and wanted to clean this up in the process because it was difficult to add new tests to this test suite due to the opinionated nature of the tests.

I think we should generally avoid using Jest hooks to create a bunch of global test variables and keep each test's state local to the individual test. Global test variables like this encourage bad test practices, can hide/introduce bugs in tests, and can make it difficult to add tests.


aiwilliams (approved):
💯

PR #375, opened 2020-11-13
Expose BucketMap and FileSystemGraphObjectStore

austinkelleher:
Going to be using `BucketMap` in the managed `GraphObjectStore` implementation. Exposing `FileSystemGraphObjectStore` is not a necessary change, but I figured it didn't hurt to expose this to potential implementors who may want to extend it now that we are allowing a custom `graphObjectStore` to be passed into `executeIntegrationInstance`.

### DIFF
--- a/packages/integration-sdk-runtime/src/__tests__/index.test.ts
+++ b/packages/integration-sdk-runtime/src/__tests__/index.test.ts
@@ -1,0 +1,11 @@
+import { BucketMap, FileSystemGraphObjectStore } from '../index';
+
+describe('#storage', () => {
+  test('should expose BucketMap', () => {
+    expect(BucketMap).not.toEqual(undefined);
+  });
+
+  test('should expose FileSystemGraphObjectStore', () => {
+    expect(FileSystemGraphObjectStore).not.toEqual(undefined);
+  });
+});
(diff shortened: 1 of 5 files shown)


packages/integration-sdk-runtime/src/__tests__/index.test.ts
```diff
@@ -0,0 +1,11 @@
+import { BucketMap, FileSystemGraphObjectStore } from '../index';
```
austinkelleher, line 1: Small test to validate that nobody removes these from the index accidentally

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/BucketMap.ts
```diff
@@ -6,7 +6,11 @@
  * and also tracks the total count of objects stored.
  */
 export class BucketMap<T = any> extends Map<string, T[]> {
-  totalItemCount = 0;
+  private internalTotalItemCount = 0;
```
austinkelleher, line 9: Bonus changes to prevent consumers from mutating this value.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/BucketMap.ts
```diff
@@ -6,7 +6,11 @@
  * and also tracks the total count of objects stored.
  */
 export class BucketMap<T = any> extends Map<string, T[]> {
```
aiwilliams, line 8: Should this be pulled up out of the `FileSystemGraphObjectStore` directory? Seems it's part of the `storage` module, used by the `FileSystemGraphObjectStore`, but not dependent/exclusive to it.

aiwilliams (approved):

PR #378, opened 2020-11-18
Stop handling IntegrationDuplicateKeyError as fatal

ndowmon:


packages/integration-sdk/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to
 
 ## Unreleased
 
+### Changed
+
+- Stop handling `IntegrationDuplicateKeyError` as fatal
```
aiwilliams, line 14:
```suggestion
- Stop handling `IntegrationDuplicateKeyError` as fatal. Previously, this type of error would terminate the integration completely and no data would be ingested. Now, the step that raises this error will not complete, and any steps that depend on the failed step will not execute, but all other steps will complete and partial datasets will be processed.
```

aiwilliams (changes requested):
Sweet, thanks @ndowmon!

I'd like to have `CHANGELOG.md` include enough information to communicate the impact of changes.

packages/integration-sdk/CHANGELOG.md
aiwilliams: I think in the future it could be worthwhile to have the `jobState` just ignore the duplicate data and have it send us something in Sentry, so that it can ingest all the data except the duplicate, and we have what we need in Sentry to investigate the duplicate and adjust the integration to stop producing them. That said, this solution is a great step forward to avoid a total crash!

austinkelleher (commented):
Nice! This change looks good, but I'm not sure if this will be sufficient to actually allow duplicate keys. I think the synchronization `/entities` and `/relationships` allow passing an optional querystring parameter that indicates whether duplicate keys are allowed.

aiwilliams (dismissed):

packages/integration-sdk-runtime/src/execution/__tests__/executeIntegration.test.ts
```diff
@@ -913,13 +913,24 @@ describe('executeIntegrationInstance', () => {
...
+      integrationStepResults: [
+        {
+          encounteredTypes: ['duplicate_entity'],
+          status: 'failure',
+        },
+      ],
+      metadata: {
+        partialDatasets: {
+          types: [],
+        }
```
ndowmon, line 931: In this case, does the one successful `duplicate_entity` get sent to the persister? If so, `duplicate_entity` should be specified as `partial`.

packages/integration-sdk-runtime/src/execution/__tests__/executeIntegration.test.ts
```diff
@@ -956,13 +967,30 @@ describe('executeIntegrationInstance', () => {
...
+        },
+        {
+          id: 'b',
+          encounteredTypes: [],
+          status: 'failure',
+        },
+      ],
+      metadata: {
+        partialDatasets: {
+          types: [],
```
ndowmon, line 990: In this case I think `duplicate_entity` should definitely be added to `partial`, right?

aiwilliams: I don't think it should be in `partialDatasets.types` because the test steps do not declare the types of entities or relationships they will produce.

aiwilliams (commented):
Thanks for thinking to add these tests! I think that if you add a couple more that prove `partialDatasets.types` will included declared types it will help demonstrate how those declarations are important.

aiwilliams (approved):

PR #381, opened 2020-11-20
Compress local integration graph object files

austinkelleher:
Here are my findings for compression/decompression of local integration data:

```
Google Cloud Integration
------------------------
No compression: 
  Size = 2.5M
  Time = 26s
gzip compression:
  Size = 516k
  Time = 27s
brotli compression:
  Size = 480k
  Time = 27s
```

Summary:
- Compressing and decompressing all of the integration files did not impact (some runtime variance accounting for 1 second surely)  the integration total run time.
- Compressing the files resulted in a 5x reduction in total file size
- Brotli produced the smallest file sizes

The current implementation requires an environment variable to be specified. There was a lot of code churn when trying to implement a new option at the execution context level. We will always want this to be enabled in the managed runtime and we'd want local runs to stay the same so that the user can view the local graph files in a human-readable format. Moving the option up to the `executeIntegrationInstance` level is something we should consider doing though.


packages/integration-sdk-private-test-utils/__fixtures__/synchronization-compressed/.j1-integration/graph/relationship-1.json
```diff
@@ -0,0 +1,2 @@
+%hι.�W�C�C4�[��?x�7�fEA�`6�+?"�շX�P�-�
```
mknoedel, line 1: In case you are wondering, this Google Translated is 
ᬥ Ũᰇ kyart ⺖ Ci 䏐 䌴 jjut  㽸 turbsyeob ፅ 䇺 Bi Fang 㼍 ⋒ hyaerb ᑘ Luan Ma ⷤਰ kkwik The Lai twap Long ߑ Nu  䬱 bwarh Cha edge jektae  chwet Yan Guo  Mulhouse Johnson ถ byaen Huan ssyeotseot 䳂  䠄ጡ ㅽ

aiwilliams (approved):
Looks good!

ndowmon (approved):

PR #386, opened 2020-12-03
Add registerEventEmitters

ndowmon:
Lots of code for a not huge change, but I feel testing this one was important, and unfortunately these types of emitters are not testable using Jest.

### DIFF
--- a/packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
+++ b/packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
@@ -1,0 +1,89 @@
+import { IntegrationLogger } from '.';
+
+type LifecycleErrorCallback = (err: Error) => void;
+
+function multipleResolveListener(callback: LifecycleErrorCallback) {
+  // conforms to type MultipleResolveListener, see node_modules/@types/node/globals.d.ts
+  return (type: 'resolve' | 'reject', promise: Promise<any>, value: any) => {
+    if (type === 'reject') {
+      callback(value);
+    }
+  };
+}
+
+function unhandledRejectionListener(callback: LifecycleErrorCallback) {
+  // conforms to type UnhandledRejectionListener, see node_modules/@types/node/globals.d.ts
+  return (reason: {} | null | undefined, promise: Promise<any>) => {
+    callback(reason as Error);
+  };
+}
+
+function uncaughtExceptionListener(callback: LifecycleErrorCallback) {
+  // conforms to type UncaughtExceptionListener, see node_modules/@types/node/globals.d.ts
+  return (error: Error) => {
+    callback(error);
+  };
+}
+
+/**
+ * Because individual integrations contain code in their `validateInvocation` and `integrationSteps`
+ * that could include unhandled rejections or multiple resolves, this function is exported as a convenience
+ * to catch and handle events emitted from an integration. This emitter should be instrumented as early
+ * as possible in the node process in order to handle any instrumentation exceptions.
+ *
+ * Before exiting the node process, unregister these event emitters using `unregisterEventEmitters`
+ */
+export function registerEventEmitters(callback: LifecycleErrorCallback) {
+  process.on('multipleResolves', multipleResolveListener(callback));
+  process.on('unhandledRejection', unhandledRejectionListener(callback));
+  process.on('uncaughtException', uncaughtExceptionListener(callback));
+}
+
+/**
+ * Call this function before exiting the node process when using `registerEventEmitters`
+ */
+export function unregisterEventEmitters(callback: LifecycleErrorCallback) {
+  process.nextTick(() => {
+    process.removeListener(
+      'multipleResolves',
+      multipleResolveListener(callback),
+    );
+    process.removeListener(
+      'unhandledRejection',
+      unhandledRejectionListener(callback),
+    );
+    process.removeListener(
+      'uncaughtException',
+      uncaughtExceptionListener(callback),
+    );
+  });
+}
+
+function integrationLoggerEventEmitterCallback(
+  getLogger: () => Pick<IntegrationLogger, 'error'>,
+): LifecycleErrorCallback {
+  return (err: Error) => getLogger().error(err);
+}
+
+/**
+ * Most often, unhandled rejections should be handled using logger.error().
+ * As a convenience, registerIntegrationLoggerEventEmitters passes logger.error() as the callback
+ * to registerEventEmitters.
+ *
+ * Before exiting the node process, unregister these event emitters using
+ * `unregisterIntegrationLoggerEventEmitters`
+ */
+export function registerIntegrationLoggerEventEmitters(
+  getLogger: () => Pick<IntegrationLogger, 'error'>,
+) {
+  registerEventEmitters(integrationLoggerEventEmitterCallback(getLogger));
+}
+
+/**
+ * Call this function before exiting the node process when using `registerIntegrationLoggerEventEmitters`
+ */
+export function unregisterIntegrationLoggerEventEmitters(
+  getLogger: () => Pick<IntegrationLogger, 'error'>,
+) {
+  unregisterEventEmitters(integrationLoggerEventEmitterCallback(getLogger));
+}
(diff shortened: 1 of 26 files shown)


mknoedel (dismissed):
Really nice, looks great!

ndowmon:
Just added an additional commit to include the event (`multipleResolves`, `unhandledRejection`, `uncaughtException`) when logging an error.

packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/**
+ * Call this function before exiting the node process when using `registerEventEmitters`
+ */
+export function unregisterEventEmitters(callback: LifecycleErrorCallback) {
+  process.nextTick(() => {
+    process.removeListener(
+      'multipleResolves',
+      multipleResolveListener(callback),
```
austinkelleher, line 51: I'm not sure that this will behave as you expect. `removeListener` requires a direct reference to the original function that was registered. This will create a new function instead.

packages/integration-sdk-runtime/testEventEmitters.js
```diff
@@ -0,0 +1,27 @@
+require('./test/mocks/mockWriteJsonToPath');
```
austinkelleher, line 1: Are these files going to end up being bundled in when the packages are built? Can you ensure that this is ignored? It may make sense to rename this file to match our same test naming convention, but ignore the file with jest (e.g. `testEventEmitters.test.ts`)

It would also be great if we could make all of these files TypeScript and execute with `ts-node` instead.

austinkelleher: I see there is a `/test` directory. Maybe we should move this into there instead of at the top-level of the package?

austinkelleher (changes requested):
This is looking great! Thanks for the changes. I left a few comments.

packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
ndowmon: Great catch, Austin.

What do you think about this as an implementation?

```typescript
// first, define event emitter. maintains singleton & throws if defineEventEmitterCallback is called more than once
/**
 * First define event emitter. Maintains singleton & throws if defineEventEmitter is called
 * more than once
 */
defineEventEmitterCallback((err, event) => { logger.error({ err, event });

/**
 * registers based on the callback defined in defineEventEmitterCallback. throws if 
 * defineEventEmitterCallback has not been called
 */
registerEventEmitters();

// EXECUTION CODE HERE

/**
 * unregisters the event emitters based on the callback defined above. throws if 
 * registerEventEmitters was not called
 */
unregisterEventEmitters();
 ```

Alternatively, a slightly simpler implementation but one that requires `registerEventEmitters` to also maintain the callback singleton

```typescript
/**
 * Maintains a callback singleton & throws if registerEventEmitters was already called.
 */
registerEventEmitters((err, event) => { logger.error({ err, event });

// EXECUTION CODE HERE

/**
 * SAME AS ABOVE
 */
unregisterEventEmitters();

Comment (austinkelleher) [reply packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts:51]: __Discussed offline__

 I think `registerEventEmitters` should return the functions that were registered and the `unregisterEventEmitters` function should take the return value of `registerEventEmitters`.

Perhaps something like this:

```typescript
function registerEventEmitters(callback) {
  const multipleResolveListenerFn = multipleResolveListener(callback);
  // Register all the goodies
  return {
    'multipleResolves': multipleResolveListenerFn
  }
}
const registeredEventEmitters = registerEventEmitters(cb)
unregisterEventEmitters(registeredEventEmitters)
```

packages/integration-sdk-runtime/testEventEmitters.js
ndowmon: Yeah, I had initially put this file in the root because I knew I would need to traverse from `test` to `src` (or `dist`) and I thought I would have had module resolution issues if my entrypoint was within `test`. But it seems that since I invoke `node` (or `ts-node`) from the root, the entrypoint is high enough to not cause resolution issues.

Resolved this in an incoming commit.

austinkelleher (dismissed):

packages/integration-sdk-runtime/test/multipleResolves.ts
```diff
@@ -0,0 +1,52 @@
...
+  registerIntegrationLoggerEventEmitters,
+  unregisterIntegrationLoggerEventEmitters,
+} from '../src';
+import {
+  LOCAL_INTEGRATION_INSTANCE,
+  createMockIntegrationLogger,
+} from './util/fixtures'
+import { expect } from './util/expect';
+
+function callbackThrowsMultipleResoles(err) {
```
aiwilliams, line 14:
```suggestion
function callbackThrowsMultipleResolves(err) {
```

packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
```diff
@@ -0,0 +1,117 @@
...
+// conforms to type UncaughtExceptionListener, see node_modules/@types/node/globals.d.ts
+type UncaughtExceptionListener = (error: Error) => void;
+
+function createUncaughtExceptionListener(callback: LifecycleErrorCallback): UncaughtExceptionListener {
+  return (error) => {
+    callback(error, 'uncaughtException');
+  };
+}
+
+interface RegisteredEventEmitters {
```
aiwilliams, line 36: Aren't these really registered event listeners? They all have the name `*Listener`. 🤔

packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
```diff
@@ -0,0 +1,117 @@
...
+ * that could include unhandled rejections or multiple resolves, this function is exported as a convenience
+ * to catch and handle events emitted from an integration. This emitter should be instrumented as early
+ * as possible in the node process in order to handle any instrumentation exceptions.
+ *
+ * Before exiting the node process, unregister these event emitters using `unregisterEventEmitters`
+ * 
+ * @param callback
+ * @returns {RegisteredEventEmitters} Pass these listeners into `unregisterEventEmitters`
+ */
+export function registerEventEmitters(callback: LifecycleErrorCallback): RegisteredEventEmitters {
```
aiwilliams, line 53: I wonder if this is really registering error event handlers. 🤔

ndowmon: Thanks for the review @aiwilliams! I agree with those naming conventions.

packages/integration-sdk-runtime/test/multipleResolves.ts
ndowmon: 👍  great catch!

packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
ndowmon: Agreed 😄  check the new commit 💯

packages/integration-sdk-runtime/test/index.ts
```diff
@@ -1,19 +1,23 @@
...
+} from './lateLoggerRegistration';
+import {
+  executeIntegrationInstanceWithUnregisteredEventEmitter,
+} from './unregisterEventHandlers';
 
 const eventEmitterTests = {
   unhandledRejection: executeIntegrationInstanceWithUnhandledRejection,
   multipleResolves: executeIntegrationInstanceWithMultipleResolves,
   lateRegisteredLogger: executeIntegrationInstanceWithLateRegisteredLogger,
+  unregisterEventHandlers: executeIntegrationInstanceWithUnregisteredEventEmitter,
```
aiwilliams, line 20: Needs a rename?

aiwilliams (approved):

packages/integration-sdk-runtime/test/index.ts
ndowmon: Of course i missed one 🤦

aiwilliams (approved):

PR #387, opened 2020-12-04
Local execution event emitters

ndowmon:
I tested this by removing all but two fake steps in the `graph-bamboo` project, linking `@jupiterone/integration-sdk-runtime` and executing with the following fake execution handlers:

```typescript
async function emitsMultipleResolve() {
  await Promise.all([
    (async () => {
      await Promise.resolve();
      throw new Error('Multiple Resolves Error #1');
    })(),
    (async () => {
      await Promise.resolve();
      throw new Error('Multiple Resolves Error #2');
    })(),
  ]);
}

function emitsUnhandledRejection() {
  async function throwsException() {
    await Promise.resolve();
    throw new Error('Unhandled Rejection');
  }
  void throwsException();
}
```

The console printed the following, showing both **Multiple Resolves Error #2** and **Unhandled Rejection**

```shell
nickdowmon@Nicks-MacBook-Pro graph-bamboohr % yarn start
yarn run v1.22.10
$ j1-integration collect
TypeScript files detected. Registering ts-node.

Configuration loaded! Running integration...

executing integration locally (linked)
23:21:58.754Z  INFO Local: Starting execution with config... (compressionEnabled=false)
23:21:58.757Z  INFO Local: Collected metric.
  metric: {
    "name": "disk-usage",
    "value": 558,
    "unit": "Bytes",
    "timestamp": 1607124118757
  }
23:21:58.762Z  INFO Local: Starting step "Fetch Account Details"... (stepId=fetch-account)
23:21:58.762Z  INFO Local: Starting step "Other"... (stepId=other)
23:21:58.762Z  INFO Local: Completed step "Fetch Account Details". (stepId=fetch-account)
23:21:58.767Z ERROR Local: Step "Other" failed to complete due to error. (errorCode="UNEXPECTED_ERROR", errorId="01d3a004-b8c3-4c80-8f74-1daf795f8a4e", reason="Unexpected error occurred executing integration! Please contact us in Slack or at https://support.jupiterone.io if the problem continues to occur.") (stepId=other, errorId=01d3a004-b8c3-4c80-8f74-1daf795f8a4e)
  Error: Multiple Resolves Error #1
      at /Users/nickdowmon/projects/github/graph-bamboohr/src/steps/account.ts:43:13
      at async Promise.all (index 0)
      at Object.emitsMultipleResolve [as executionHandler] (/Users/nickdowmon/projects/github/graph-bamboohr/src/steps/account.ts:40:3)
[...]

PR #388, opened 2020-12-05
Make registerIntegrationEventEmitters call logger.onFailure

ndowmon:
Merges into #387

### DIFF
--- a/packages/integration-sdk-runtime/src/logger/index.ts
+++ b/packages/integration-sdk-runtime/src/logger/index.ts
@@ -169,7 +169,7 @@ export class IntegrationLogger extends EventEmitter
   implements IntegrationLoggerType {
   private _logger: Logger;
   private _errorSet: Set<Error>;
-  private onFailure: OnFailureFunction;
+  readonly onFailure: OnFailureFunction;
 
   constructor(input: IntegrationLoggerInput) {
     super();
--- a/packages/integration-sdk-runtime/src/logger/registerEventHandlers.ts
+++ b/packages/integration-sdk-runtime/src/logger/registerEventHandlers.ts
@@ -95,16 +95,16 @@ export function unregisterEventHandlers({
   });
 }
 
-interface MinimalLogger {
+interface ErrorLogger {
   error: IntegrationLogger['error'];
   onFailure?: IntegrationLogger['onFailure'];
 }
 
 function integrationLoggerEventHandlerCallback(
-  getLogger: () => MinimalLogger,
+  getErrorLogger: () => ErrorLogger,
 ): LifecycleErrorCallback {
   return (err, event) => {
-    const logger = getLogger();
+    const logger = getErrorLogger();
     logger.error({ err, event });
     if (logger.onFailure) {
       logger.onFailure({ err });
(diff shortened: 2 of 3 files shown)


packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
```diff
@@ -73,25 +91,27 @@ export function unregisterEventEmitters({
...
+  getLogger: () => MinimalLogger,
 ): LifecycleErrorCallback {
-  return (err, event) => getLogger().error({ err, event });
+  return (err, event) => {
+    const logger = getLogger();
+    logger.error({ err, event });
+    if (logger.onFailure) {
+      logger.onFailure({ err });
+    }
+  };
```
ndowmon, line 114: Here's the change in this file. The rest is the addition of the `MinimalLogger` interface or changes the linter made.

packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -169,7 +169,7 @@ export class IntegrationLogger extends EventEmitter
   implements IntegrationLoggerType {
   private _logger: Logger;
   private _errorSet: Set<Error>;
-  private onFailure: OnFailureFunction;
+  readonly onFailure: OnFailureFunction;
```
ndowmon, line 172: Made `readonly` so we can call this directly from `IntegrationLogger`. Calling this should not be encouraged though.

austinkelleher (approved):

austinkelleher (dismissed):

mknoedel (approved):

packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
```diff
@@ -73,25 +91,27 @@ export function unregisterEventEmitters({
...
-      'uncaughtException',
-      uncaughtExceptionListener,
-    );
+    process.removeListener('multipleResolves', multipleResolveListener);
+    process.removeListener('unhandledRejection', unhandledRejectionListener);
+    process.removeListener('uncaughtException', uncaughtExceptionListener);
   });
 }
 
+interface MinimalLogger {
```
aiwilliams, line 100: What do you think of the name `ErrorLogger`?

packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
```diff
@@ -103,15 +123,17 @@ function integrationLoggerEventEmitterCallback(
  * `unregisterIntegrationLoggerEventEmitters`
  */
 export function registerIntegrationLoggerEventEmitters(
-  getLogger: () => Pick<IntegrationLogger, 'error'>,
+  getLogger: () => MinimalLogger,
```
aiwilliams, line 126: Then this could be `getErrorLogger()`.

packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
ndowmon: Clearer interface name. Thanks for the suggestion 👍

packages/integration-sdk-runtime/src/logger/registerEventEmitters.ts
ndowmon: 👍

aiwilliams (approved):

mknoedel (approved):

PR #391, opened 2020-12-10
Updated name of InvocationConfigNotFound error to InvocationConfigLoa…

kevincasey1222:
…dError

The InvocationConfigNotFound error can actually be caused not only by the
invocationConfig failing to load from src/index.ts, but also by any other
reason that the require statement fails. In such cases, the error
InvocationConfigNotFound is misleading and masks the real problem.


aiwilliams (approved):
Thanks for the contribution!

Issue #385, opened 2020-11-30
Improve performance of `FileSystemGraphObjectStore` `getEntity`


austinkelleher:
The `getEntity` function in `FileSystemGraphObjectStore` is currently iterating every entity on disk to catch duplicate entity keys. We should trust the `DuplicateKeyTracker` and return the first entity with the given key. On integrations that ingest a large amount of data, we do not want to iterate every local file.

See: https://github.com/JupiterOne/sdk/blob/master/packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts#L67

austinkelleher:
This is one small change, but a significant improvement would be to create an in-memory Map that tracks the file path of the file by `_key` and the index of the entity in the file. That would drastically improve the performance of local entity searches.

austinkelleher:
Fixed in #454

PR #395, opened 2020-12-16
Optimize graph object buffering and flushing

austinkelleher:
## Overview

The motivation of this change is to allow our `FileSystemGraphObjectStore` to buffer more data in memory before performing a flush. 

## Previously Implementation Details

Whenever we called `jobState.findEntity`, `jobState.iterateEntities`, or `jobState.iterateRelationships`, a flush would be performed. Multiple steps are typically running at the same time in an integration. This means that flushing is happening all of the time.

## New Implementation Details

We now have two in-memory graph object lookup tables:

> A = Map<Graph Object _key, Graph Object>
> B = Map<Graph Object _type, Map<Graph Object _key, boolean>>

When we perform a `jobState.findEntity`, we first check map `A` to see if we have this graph object buffered in memory. If we do not, then we perform our traditional lookup on disk. (There is actually more we can do to optimize this. See here: https://github.com/JupiterOne/sdk/issues/385#issuecomment-735889750)

When we perform a `jobState.iterateEntities` or a `jobState.iterateRelationships`, we first iterate over the data we have buffered in map `B`, then we perform our traditional disk file iteration method.


packages/integration-sdk/CHANGELOG.md
```diff
@@ -9,6 +9,23 @@ and this project adheres to
...
+- Allow `FileSystemGraphObjectStore` to specify `graphObjectBufferThreshold`,
+  which defines the maximum number of graph objects that the graph object store
+  can buffer into memory before flushing to disk. Machines with more memory
+  should consider increasing this value as the default is 500.
+
+### Removed
+
+- BREAKING: Remove `BucketMap` as it's no longer used in the
+  `FileSystemGraphObjectStore`. `BucketMap` was exposed externally, but should
+  not be used externally either.
```
ndowmon, line 27: The wording here is a little confusing. Can you clarify "`BucketMap` was exposed externally, but should
  not be used externally either."?

packages/integration-sdk-runtime/src/storage/memory.ts
```diff
@@ -0,0 +1,239 @@
...
+   * This will allow us to eventually run steps in parallel without locking when
+   * steps have no interdependence. If we cleared it out, it's possible that
+   * two steps could be running in parallel and one of the steps may clear out
+   * the maps while the other is still relying on it.
+   */
+  flushEntities(entities: Entity[]) {
+    for (const entity of entities) {
+      this.entityKeyToEntityMap.delete(entity._key);
+      const entityTypeKeysMap = this.entityTypeToKeysMap.get(entity._type);
+      entityTypeKeysMap?.delete(entity._key);
```
ndowmon, line 176: I wonder if we should really conditionally chain `entityTypeKeysMap?.delete(entity._key)` here. If `entityTypeKeysMap` is undefined, doesn't this represent an out-of-sync memory data store, something we would consider an exception?

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -12,34 +12,49 @@ import {
...
   }
 
   async addEntities(storageDirectoryPath: string, newEntities: Entity[]) {
-    this.entityStorageMap.add(storageDirectoryPath, newEntities);
+    this.localGraphObjectStore.addEntities(storageDirectoryPath, newEntities);
 
-    if (this.entityStorageMap.totalItemCount >= GRAPH_OBJECT_BUFFER_THRESHOLD) {
+    if (
+      this.localGraphObjectStore.getTotalEntityItemCount() >=
+      this.graphObjectBufferThreshold
```
ndowmon, line 56: IMO `graphObjectBufferThreshold` makes me think that the `localGraphObjectStore` can store 500 graph objects, but really it can store 500 of each class of graph object (500 entities & 500 relationships). Just curious if it's worth maintaining a `private readonly entityBufferThreshold: number;` and separately a `private readonly relationshipBufferThreshold: number;`, or else update docstrings on `FileSystemGrahpObjectStoreParams.graphObjectBufferThreshold` to reflect that totals for entities and relationships are kept separate?

packages/integration-sdk-runtime/src/storage/memory.ts
```diff
@@ -0,0 +1,239 @@
...
+   * two steps could be running in parallel and one of the steps may clear out
+   * the maps while the other is still relying on it.
+   */
+  flushRelationships(relationships: Relationship[]) {
+    for (const relationship of relationships) {
+      this.relationshipKeyToRelationshipMap.delete(relationship._key);
+      const relationshipTypeKeysMap = this.relationshipTypeToKeysMap.get(
+        relationship._type,
+      );
+      relationshipTypeKeysMap?.delete(relationship._key);
```
ndowmon, line 193: Same as above.

ndowmon (commented):
Looks good @austinkelleher, thanks for this improvement!

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
```diff
@@ -259,7 +259,7 @@ describe('getEntity', () => {
     ]);
 
     const entity = await store.getEntity({ _key, _type });
-    expect(store.entityStorageMap.totalItemCount).toEqual(0);
+    // expect(store.entityStorageMap.totalItemCount).toEqual(0);
```
charlieduong94: Might as well delete these commented out expectations right?

charlieduong94 (changes requested):
Just a small change requested regarding some commented out code. Otherwise this looks pretty good!

packages/integration-sdk-runtime/src/storage/memory.ts
charlieduong94: Yeah, it seems like that kind of situation would be an indicator of syncing issues. Throwing an exception sounds reasonable to me, but if you end up thinking that that the absence of the key map shouldn't cause a step to fail, perhaps consider logging out a warning or error message.

packages/integration-sdk/CHANGELOG.md
austinkelleher: I had originally planned to use it in a managed graph object store, but opted not to. Nobody should be using it externally, but it's technically exposed to the external API via:

```ts
import { BucketMap } from '@jupiterone/integration-sdk-runtime';
```

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
austinkelleher: Perhaps we can consider introducing that level of granularity later if necessary. I don't think it's worth maintaining the two properties for now though. I imagine that relationships generally consume less memory than entities, so it may be worth considering later.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
austinkelleher: I ended up removing this in a later PR with more changes. Sorry for the confusion.

packages/integration-sdk-runtime/src/storage/memory.ts
austinkelleher: Agreed. This would be indicative of a bug. I'm changing it to throw.

austinkelleher: Fixed

ndowmon (dismissed):
👍

charlieduong94 (dismissed):

charlieduong94 (dismissed):

ndowmon (approved):

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -285,6 +298,19 @@ export function executeStepDependencyGraph<
...
+
+      if (context.jobState.waitUntilUploadsComplete) {
+        try {
+          // Failing to upload all integration data should not be considered a
+          // fatal failure. We just want to make this step as a partial success
+          // and move on with our lives!
+          await context.jobState.waitUntilUploadsComplete();
+        } catch (err) {
+          context.logger.stepFailure(step, err);
+          status = StepResultStatus.FAILURE;
```
mknoedel, line 310: Kind of strange that there is no differentiation between a failure on the step and a failure when attempting to persist the data. Idk if it is worth adding a way to differentiate that or not.

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -285,6 +298,19 @@ export function executeStepDependencyGraph<
       }
 
       await context.jobState.flush();
+
+      if (context.jobState.waitUntilUploadsComplete) {
+        try {
+          // Failing to upload all integration data should not be considered a
+          // fatal failure. We just want to make this step as a partial success
```
mknoedel, line 305: Really unimportant, but 'make' should likely be 'mark'

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -12,54 +12,90 @@ import {
...
-      GRAPH_OBJECT_BUFFER_THRESHOLD
+      this.localGraphObjectStore.getTotalRelationshipItemCount() >=
+      this.graphObjectBufferThreshold
     ) {
-      await this.flushRelationshipsToDisk();
+      await this.flushRelationshipsToDisk(onRelationshipsFlushed);
     }
   }
 
   async getEntity({ _key, _type }: GraphObjectLookupKey): Promise<Entity> {
```
mknoedel, line 91: Thought: what information is typically gotten from previous entities? My thought would be that it would usually only be foreign keys, which is likely just an id. Is it worth considering when we flush `InMemoryGraphObjectStore`, we maintain a map similar to the implementation of `DuplicateKeyTracker`. 
Maybe we could get lucky and reduce roundtrips to the disk a good amount of the time. ¯\_(ツ)_/¯

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -166,7 +222,7 @@ export class FileSystemGraphObjectStore implements GraphObjectStore {
    * and prematurely end, causing the next step to start up before the
    * data it depends on is present on disk.
    */
-  async lockOperation(operation: () => Promise<any>) {
+  private async lockOperation<T>(operation: () => Promise<T>) {
```
mknoedel, line 225: Looking at the [documentation](https://www.npmjs.com/package/p-map) for `p-map`, it looks there is a concurrency parameter that we can use instead of using a semaphore.

packages/integration-sdk-runtime/src/storage/memory.test.ts
```diff
@@ -0,0 +1,214 @@
+import {
+  Entity,
+  ExplicitRelationship,
+  Relationship,
+} from '@jupiterone/integration-sdk-core';
+import { InMemoryGraphObjectStore } from './memory';
+import { v4 as uuid } from 'uuid';
+
+function createTestEntity(partial?: Partial<Entity>): Entity {
```
mknoedel, line 9: You pulled these out earlier in this PR  to `packages/integration-sdk-private-test-utils/src/graphObject.ts`

mknoedel (approved):
Looks like a great improvement!

packages/integration-sdk-runtime/src/storage/memory.test.ts
austinkelleher: Good call. Fixing.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
austinkelleher: Ya, I think there are ways we can get rid of this, but we should tackle it separately.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
austinkelleher: There are definitely lots of additional caching improvements that we can make!

ndowmon (approved):

mknoedel (dismissed):

ndowmon (approved):

PR #396, opened 2020-12-16
Support for continuous integration data uploads

austinkelleher:
Our previous integration flow had two primary phases:

Phase 1: Collect all of the data
Phase 2: Upload all of the data

These changes improve this drastically by mixing the two phases. We will now queue up graph object uploads in each step and ensure that all of the uploads have successfully completed before triggering dependent steps.


packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -284,6 +297,20 @@ export function executeStepDependencyGraph<
         status = StepResultStatus.FAILURE;
       }
 
+      await context.jobState.flush();
```
ndowmon, line 300: Interesting. Looks like we now flush after each new step. 😈  advocate: do we want to change the implementation of the maps in `localGraphDataStore`, which tracks `stepId` and `_key` for entities?

packages/integration-sdk-runtime/src/execution/uploader.ts
```diff
@@ -0,0 +1,110 @@
...
+        queue.pending >= uploadConcurrency ||
+        queue.size >= uploadConcurrency
+      ) {
+        if (onThrottleEnqueue) {
+          // Mainly just used for testing that our custom throttling works.
+          onThrottleEnqueue();
+        }
+
+        await queue.onIdle();
+      }
```
ndowmon, line 61: Should the `enqueue` function be `async` (does it have to be)? I'm thinking from the caller's perspective:

```typescript
    await graphObjectStore.addEntities(stepId, entities, async (entities) =>
      uploader?.enqueue({
        entities,
        relationships: [],
      }),
    );
 ```

Maybe the caller should only need to wait for entities to hit the graph object store, and not for these to be published to the persister. 🤷

File: packages/integration-sdk-runtime/src/synchronization/index.ts [hunk 3]
```diff
@@ -183,28 +208,12 @@ export async function uploadCollectedData(context: SynchronizationJobContext) {
...
-                parsedData.relationships,
-              );
-            }
-          } catch (err) {
-            throw synchronizationApiError(
-              err,
-              'Error uploading collected data',
-            );
-          }
+          await uploadGraphObjectData(
```
ndowmon, line 211: I see that `uploadCollectedData` is still being called from `synchronizeCollectedData` (which is used in the `j1-integration sync` command). I haven't looked at the follow-on PRs yet so maybe you will do this, but I think it'd be unsafe to leave this functionality in the off chance that it duplicates uploads to the persister.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
```diff
@@ -426,6 +426,186 @@ describe('iterateRelationships', () => {
...
+        );
+        return Promise.resolve();
+      },
+    );
+
+    expect(addRelationshipsFlushCalledTimes).toEqual(1);
+    expect(flushedRelationshipsCollected).toEqual([r1, r2, r3]);
+  });
+
+  test('#flushEntitiesToDisk should call flush callback when buffer threshold reached', async () => {
```
ndowmon, line 506: Seems like copy-paste error? The buffer threshold has not been reached here.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
```diff
@@ -426,6 +426,186 @@ describe('iterateRelationships', () => {
...
+    expect(addEntitiesFlushCalledTimes).toEqual(0);
+    expect(flushedEntitiesCollected).toEqual([]);
+
+    await store.flushEntitiesToDisk(onEntitiesFlushed);
+
+    expect(addEntitiesFlushCalledTimes).toEqual(1);
+    expect(flushedEntitiesCollected).toEqual(entities);
+  });
+
+  test('#flushRelationshipsToDisk should call flush callback when buffer threshold reached', async () => {
```
ndowmon, line 533: Seems like copy-paste error? The buffer threshold has not been reached here.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
```diff
@@ -426,6 +426,186 @@ describe('iterateRelationships', () => {
...
+    expect(addRelationshipsFlushedCalledTimes).toEqual(0);
+    expect(flushedRelationshipsCollected).toEqual([]);
+
+    await store.flushRelationshipsToDisk(onRelationshipsFlushed);
+
+    expect(addRelationshipsFlushedCalledTimes).toEqual(1);
+    expect(flushedRelationshipsCollected).toEqual(relationships);
+  });
+
+  test('#flush should call both entity and relationship flush callbacks when buffer threshold reached', async () => {
```
ndowmon, line 566: Seems like copy-paste error? The buffer threshold has not been reached here.

packages/integration-sdk-cli/src/commands/run.ts
```diff
@@ -13,11 +13,14 @@ import {
   getApiBaseUrl,
   getApiKeyFromEnvironment,
   initiateSynchronization,
-  uploadCollectedData,
+  // uploadCollectedData
```
ndowmon, line 16: I'm sure you'll do this in a future PR but this should be removed, not just commented out.

packages/integration-sdk-runtime/src/execution/uploader.ts
```diff
@@ -0,0 +1,110 @@
...
+          // step should get marked as "partial" and other steps can continue
+          // to run.
+          queue.pause();
+          queue.clear();
+
+          uploadErrors.push(err);
+        });
+    },
+
+    async waitUntilUploadsComplete() {
```
ndowmon, line 77: I saw something about `queue.isPaused` in `src/execution/uploader.ts`, should that be set somewhere? Here? IDK.

packages/integration-sdk-runtime/src/execution/jobState.ts
```diff
@@ -158,6 +174,22 @@ export function createStepJobState({
...
+          }),
+        async (relationships) =>
+          uploader?.enqueue({
+            entities: [],
+            relationships,
+          }),
+      ),
+
+    async waitUntilUploadsComplete() {
+      await uploader?.waitUntilUploadsComplete();
```
ndowmon, line 192: I may not have read the code correctly, but it does seem since `uploader.enqueue` is async and calls `await queue.onIdle()`, this will always return right away...

packages/integration-sdk-cli/src/commands/run.ts
charlieduong94, line 16: Delete this?

packages/integration-sdk-runtime/src/execution/__tests__/dependencyGraph.test.ts
```diff
@@ -622,6 +630,156 @@ describe('executeStepDependencyGraph', () => {
     expect(spyB).toHaveBeenCalledBefore(spyC);
   });
 
+  test('should mark steps failed executionHandlers with status FAILURE a dependent steps with status PARTIAL_SUCCESS_DUE_TO_DEPENDENCY_FAILURE when step upload fails', async () => {
```
ndowmon, line 633: I think there are typos here? Hard to understand.

```suggestion
  test('should mark steps with failed executionHandlers with status FAILURE and dependent steps with status PARTIAL_SUCCESS_DUE_TO_DEPENDENCY_FAILURE when step upload fails', async () => {
```

ndowmon: I guess this also doesn't seem like a test that's relevant to this PR. The scope seems to be about dependencies here, although the functionality is about creating a failing uploader. Consider trimming this, breaking it apart, or making the test name clearer.

packages/integration-sdk-runtime/src/execution/uploader.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    const uploader = createQueuedStepGraphObjectDataUploader({
+      stepId: uuid(),
+      uploadConcurrency: Infinity,
+      async upload(d) {
+        if (numQueued) {
+          await sleep(100);
+          uploaded.push(d);
+        } else {
+          numQueued++;
+          await sleep(200);
```
ndowmon, line 61: What's the purpose of these `sleep()`s? It's not very clear to me and makes the test a bit more confusing.

packages/integration-sdk-runtime/src/execution/uploader.test.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  return flushed;
+}
+
+async function createFlushedDataAndWaitForUploads(
+  uploader: StepGraphObjectDataUploader,
+  n: number,
+) {
+  const flushed = await createAndEnqueueUploads(uploader, n);
+  await uploader.waitUntilUploadsComplete();
```
ndowmon, line 43: I'm not convinced that all data is not uploaded at this point in the process. Can you add a test to prove that un-uploaded data can exist in in the upload queue and that `waitUntilUploadsComplete` does indeed properly wait for uploads to complete?

Alternatively, explain it to me like I'm a 5 year old.

ndowmon (commented):
Another sweet change. Added some comments for your review.

charlieduong94 (changes requested):

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
austinkelleher: This is confusing...I had slightly changed the behavior in the previous PR and ended up changing it back to the original in this PR. Should have just rolled back the change in the first PR. See here: https://github.com/JupiterOne/sdk/pull/395/files#diff-8912f1fec8d408545c592a8420e4d837119d01694950f32598718ba7abc57d7aL287

packages/integration-sdk-runtime/src/execution/uploader.ts
austinkelleher: It does have to be async. See this comment: https://github.com/JupiterOne/sdk/pull/396/files#diff-38e59a3db7780e509c55171310fde11a5d109af248e2b859c4f0378ee65e049fR47

packages/integration-sdk-runtime/src/synchronization/index.ts
austinkelleher: I think we still want this. I don't personally use the `sync` command. `run` collects & uploads. `sync` only uploads. The `sync` command assumes that you already have data on disk that now should be uploaded.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
austinkelleher: Thanks. Will fix.

packages/integration-sdk-runtime/src/execution/uploader.ts
austinkelleher: Good question, but no it should not be. We should only pause execution of the queue when we see an error. This function rethrows a collection of errors that we saw.

packages/integration-sdk-runtime/src/execution/jobState.ts
austinkelleher: I don't think that would be the case. The `enqueue` method is, in fact, `async`, but waiting for the enqueued function to actually settle is not async. The only reason why the `enqueue` function is async is so that we can throttle how many functions we are pushing into our queue. Throttling the `enqueue` will prevent us from buffering too many entities and relationships into memory while the uploads haven't actually completed. 

See: https://github.com/JupiterOne/sdk/pull/396/files#diff-38e59a3db7780e509c55171310fde11a5d109af248e2b859c4f0378ee65e049fR47

packages/integration-sdk-runtime/src/execution/__tests__/dependencyGraph.test.ts
austinkelleher: Will fix the grammar. Thanks for the suggestion. The test is relevant to the overall changes though. The test validates that our uploads respect the existing behavior of our internal dependency graph.

packages/integration-sdk-runtime/src/execution/uploader.test.ts
austinkelleher: Good question. Will add a comment for clarity. The sleeps validate that the promises are indeed executed concurrently instead of in serial. We enqueue a bunch of uploads and, as long as we haven't reached our concurrency limit, the promises will all be run concurrently.

packages/integration-sdk-runtime/src/execution/uploader.test.ts
austinkelleher: Perhaps I'm not following, but the function is indeed supposed to wait until uploads are complete. It's used as a utility function throughout these tests, so that we can just assert that we've collected the correct data in the correct order.

charlieduong94 (approved):

packages/integration-sdk-runtime/src/execution/uploader.ts
```diff
@@ -0,0 +1,110 @@
...
+      if (
+        queue.pending >= uploadConcurrency ||
+        queue.size >= uploadConcurrency
+      ) {
+        if (onThrottleEnqueue) {
+          // Mainly just used for testing that our custom throttling works.
+          onThrottleEnqueue();
+        }
+
+        await queue.onIdle();
```
mknoedel, line 60: I think we don't want to wait for onIdle here, as it loses the concurrency benefits. Basically we won't start on the next batch of uploads until every one from the last group is finished. We should likely wait on `onEmpty` and instead of checking if wither the pending or the size is greater than the upload concurrency, I think we should just be checking the `queue.size`

packages/integration-sdk-runtime/src/execution/uploader.test.ts
mknoedel: From what I can tell, the queue will never be that big, but there is a small amount of time when the final concurrent requests need to drain from the queue. I believe `waitUntilUploadsComplete` will wait for those final requests to resolve.

packages/integration-sdk-runtime/src/synchronization/index.ts
mknoedel: It seems kind of weird to me that there is any interaction with the persister at all in the open-source SDK as that is J1 specific. Am I wrong with the goal of the SDK?

packages/integration-sdk-cli/src/commands/run.ts
```diff
@@ -8,16 +8,19 @@ import {
...
   getAccountFromEnvironment,
   getApiBaseUrl,
   getApiKeyFromEnvironment,
   initiateSynchronization,
-  uploadCollectedData,
 } from '@jupiterone/integration-sdk-runtime';
 
 import { loadConfig } from '../config';
 import * as log from '../log';
+import { createPersisterApiStepGraphObjectDataUploader } from '@jupiterone/integration-sdk-runtime/dist/src/execution/uploader';
```
aiwilliams, line 21: Why does this have to be this way, digging into the `dist/` directory?

PR #397, opened 2020-12-16
Share graph object creation test utils across tests and cleanup

austinkelleher:
Just separated into this PR to make reviewing easier.


packages/integration-sdk-runtime/src/execution/__tests__/jobState.test.ts
```diff
@@ -8,35 +8,36 @@ import {
...
-      options?.normalizeGraphObjectKey,
-    ),
+    duplicateKeyTracker: new DuplicateKeyTracker(),
     typeTracker: new TypeTracker(),
     dataStore: new MemoryDataStore(),
+    ...partial,
   };
 }
 
+function createTestStepJobStateState(
```
ndowmon, line 28: `StateState`? Intentional?

ndowmon (approved):
Nice!

packages/integration-sdk-runtime/src/execution/__tests__/jobState.test.ts
austinkelleher: Good catch. Will fix. Thanks.

charlieduong94 (approved):

PR #399, opened 2020-12-16
Only write prettified files to the file system on local collection

austinkelleher:
Small change that should marginally improve integrations that produce a significant amount of graph objects locally.

- Less disk usage
- Faster reads
- Faster writes

### DIFF
--- a/packages/integration-sdk-cli/src/commands/collect.ts
+++ b/packages/integration-sdk-cli/src/commands/collect.ts
@@ -33,7 +33,7 @@ export function collect() {
       log.info('\nConfiguration loaded! Running integration...\n');
 
       const graphObjectStore = new FileSystemGraphObjectStore({
-        prettyFile: true,
+        prettifyFiles: true,
       });
 
       const results = await executeIntegrationLocally(
--- a/packages/integration-sdk-cli/src/commands/collect.ts
+++ b/packages/integration-sdk-cli/src/commands/collect.ts
@@ -2,6 +2,7 @@ import { createCommand } from 'commander';
 
 import {
   executeIntegrationLocally,
+  FileSystemGraphObjectStore,
   prepareLocalStepCollection,
 } from '@jupiterone/integration-sdk-runtime';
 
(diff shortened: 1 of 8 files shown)


packages/integration-sdk-runtime/src/fileSystem.ts
```diff
@@ -191,7 +195,7 @@ export async function removeStorageDirectory() {
 }
 
 function removeDirectory(directory: string) {
-  return new Promise((resolve, reject) =>
+  return new Promise<void>((resolve, reject) =>
```
austinkelleher, line 198: Bonus typescript typing fix.

packages/integration-sdk-runtime/src/fileSystem.ts
```diff
@@ -47,10 +48,13 @@ interface WriteDataToPathInput {
 export async function writeJsonToPath({
```
austinkelleher, line 48: Note: This is the main change.

packages/integration-sdk-cli/src/visualization/generateVisualization.ts
```diff
@@ -27,9 +27,11 @@ export async function generateVisualization(
     log.warn(`Unable to find any files under path: ${resolvedIntegrationPath}`);
   }
 
-  const { entities, relationships, mappedRelationships } = await retrieveIntegrationData(
```
austinkelleher, line 30: Nothing actually changed in this file. Prettier picked it up. After all of my changes get merged, I'll run prettier on the entire project to prevent this from happening.

packages/integration-sdk-cli/src/commands/collect.ts
```diff
@@ -30,6 +31,11 @@ export function collect() {
       const enableSchemaValidation = !options.disableSchemaValidation;
       const config = prepareLocalStepCollection(await loadConfig(), options);
       log.info('\nConfiguration loaded! Running integration...\n');
+
+      const graphObjectStore = new FileSystemGraphObjectStore({
+        prettyFile: true,
```
charlieduong94, line 36: There are multiple files right? I wonder if `prettyFiles` would be better. I don't care too much though, just summoning my inner Phil.

charlieduong94 (approved):
Seems gud. Maybe consider that naming feedback, but it's not a blocker.

packages/integration-sdk-cli/src/commands/collect.ts
ndowmon: I agree with ~~Phil~~ Charlie.

ndowmon: `prettifyFiles`?

ndowmon (approved):

PR #404, opened 2020-12-26
OPTIMIZATION: Support for omitting specific graph objects from file storage

austinkelleher:
Some entities and relationships do not need to be stored on the
file system at all. We only need to store graph objects on the file
system if we later intend to fetch the data from the job state or
iterate over the `_type`. This PR introduces support for specifying
metadata in each step that can be used by the `FileSystemGraphObjectStore`
that will be used to omit specific data from being written to disk
entirely. The data will still get uploaded.

A longer term optimization would be actually leveraging the dependency graph to automatically generate this information. We can also make a follow-up improvement that removes the entire locking behavior during these cases because no step will rely on the unindexed data, so there is no reason to lock.

### DIFF
--- a/packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
+++ b/packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
@@ -10,6 +10,9 @@ import {
   IntegrationMissingKeyError,
   Relationship,
   GraphObjectStore,
+  GraphObjectIndexMetadata,
+  GetIndexMetadataForGraphObjectTypeParams,
+  IntegrationStep,
 } from '@jupiterone/integration-sdk-core';
 
 import { flushDataToDisk } from './flushDataToDisk';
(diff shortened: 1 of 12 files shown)


packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -160,13 +224,29 @@ export class FileSystemGraphObjectStore implements GraphObjectStore {
...
+            });
+
+            if (typeof indexMetadata === 'undefined') {
+              return true;
+            }
+
+            return indexMetadata.enabled === true;
           });
 
+          if (indexable.length) {
```
aiwilliams, line 241: Do you think it would be worthwhile to add a test for the scenario when there are no indexable types?

aiwilliams (approved):

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
austinkelleher: Will do!

aiwilliams (approved):

PR #405, opened 2020-12-30
Fix memory leak in createIntegrationEntity

aiwilliams:


packages/integration-sdk-core/src/data/createIntegrationEntity.ts
```diff
@@ -240,15 +248,20 @@ function whitelistedProviderData(
   return whitelistedProviderData;
 }
 
-const schemaWhitelists = new Map<string[], string[]>();
```
mknoedel, line 243: Oh, yeah that makes sense. Good Catch!

mknoedel (approved):

PR #415, opened 2021-01-06
Trim integration config values

mknoedel:
This is in response to [this ticket](https://dev.azure.com/jupiterone/Platform/_sprints/taskboard/Integrations%20Team/Platform/Sprint%2062?workitem=2100) where a company accidentally added a space to their accountId. We could change the UI, but I feel that it also makes sense to  check before running integrations as welll.


charlieduong94 (approved):

charlieduong94:
LGTM, thanks for making this change!

PR #417, opened 2021-01-07
Trim integration config values

mknoedel:
Trim integration all unmasked config values prior to executing an integration.

Masked values are skipped because they may contain spaces at the start or the end. If you think there is a good place to put that information in the code let me know and I will add it!

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/__tests__/executeIntegration.test.ts
+++ b/packages/integration-sdk-runtime/src/execution/__tests__/executeIntegration.test.ts
@@ -20,10 +20,7 @@ import {
 import { InMemoryGraphObjectStore } from '@jupiterone/integration-sdk-private-test-utils';
 
 import * as integrationFileSystem from '../../fileSystem';
-import {
-  createIntegrationLogger,
-  IntegrationLogger as IntegrationLoggerImpl,
-} from '../../logger';
+import { IntegrationLogger as IntegrationLoggerImpl } from '../../logger';
 import { FlushedGraphObjectData } from '../../storage/types';
 import {
   executeIntegrationInstance,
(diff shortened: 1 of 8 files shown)


packages/integration-sdk-runtime/src/execution/__tests__/executeIntegration.test.ts
```diff
@@ -50,30 +48,6 @@ export interface InstanceConfigurationData {
   logger: IntegrationLogger;
 }
 
-function createInstanceConfiguration(
```
mknoedel, line 53: Moved to a shared utils file

packages/integration-sdk-runtime/src/execution/__tests__/executeIntegration.test.ts
```diff
@@ -32,13 +29,14 @@ import {
...
 import { LOCAL_INTEGRATION_INSTANCE } from '../instance';
+import { createInstanceConfiguration } from './utils/createIntegrationConfig';
 
 const brotliDecompress = promisify(zlib.brotliDecompress);
 
 jest.mock('fs');
 
 function sleep(ms: number) {
-  return new Promise((resolve) => {
+  return new Promise<void>((resolve) => {
```
mknoedel, line 39: TsLint error

austinkelleher (dismissed):
Awesome change!

aiwilliams (approved):
Excellent, thanks @mknoedel! Also appreciate the comments about why some changes were made.

aiwilliams:
Oh, almost forgot... the CHANGELOG.md update would be a great addition so that it travels with the commit(s)/PR.

ndowmon:
Curious, I haven't been following the conversation on this one... why did we choose to trim string values instead of having customers fix their configurations?

austinkelleher:
@ndowmon It's pretty common for customers to make mistakes like that. There isn't a great reason to _not_ trim values imo. We do this other places in the backend as well.

austinkelleher (approved):

PR #420, opened 2021-01-20
Allow duplicate key tracker to grow greater than V8 engine limit

austinkelleher:
Never hit this issue before...

The Qualys integration run on a machine that has >30GB of memory. We allow V8 to use the entire ~30GB memory section using the `--max-old-space-size` Node option when the we start the process. The Qualys integration exited with an out of memory error despite only consuming ~20% of the total memory. I learned that V8 has an internal 1GB limit allocated to each `Map`, or 2^24 keys. We are hitting the 1GB limit. See: https://stackoverflow.com/questions/54452896/maximum-number-of-entries-in-node-js-map/54466812#54466812

This change implements a basic `BigMap` that is an array of maps. We start with a map with a maximum key space of 2,000,000 keys and will add a map every time we reach that threshold. I realize this is not perfect because some integrations will produce smaller or larger keys, but to avoid too much complexity and potential slowness accrued from a lot of calculations at runtime, this is probably a good enough approach.

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/jobState.ts
+++ b/packages/integration-sdk-runtime/src/execution/jobState.ts
@@ -8,11 +8,14 @@ import {
 
 import { GraphObjectStore } from '../storage';
 import { StepGraphObjectDataUploader } from './uploader';
+import { BigMap } from './utils/bigMap';
 
 export interface DuplicateKeyTrackerGraphObjectMetadata {
   _type: string;
   _key: string;
 }
+
+const DUPLICATE_KEY_TRACKER_DEFAULT_MAP_KEY_SPACE = 2000000;
 
 /**
  * Contains a map of every graph object key to a specific set of metadata about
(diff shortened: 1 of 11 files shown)


packages/integration-sdk-runtime/src/execution/jobState.ts
```diff
@@ -22,15 +25,17 @@ export interface DuplicateKeyTrackerGraphObjectMetadata {
  * table.
  */
 export class DuplicateKeyTracker {
-  private readonly graphObjectKeyMap = new Map<
-    string,
+  private readonly graphObjectKeyMap: BigMap<
     DuplicateKeyTrackerGraphObjectMetadata
-  >();
-
+  >;
```
aiwilliams, line 30: Without taking the key type parameter, this seems more like a `BigStringMap`. 😄

aiwilliams (approved):
Need a CHANGELOG entry. Thanks for tackling this. 💪

aiwilliams (dismissed):

charlieduong94 (approved):

PR #421, opened 2021-02-09
Log successful calls to uploadData() to synchronizer

ndowmon:


mknoedel (approved):

ndowmon:
Also changed project to version together.

lerna.json
```diff
@@ -4,5 +4,5 @@
     "packages/integration-sdk-*",
     "packages/cli"
   ],
-  "version": "independent"
```
austinkelleher, line 7: Can you explain why this change was made?

ndowmon: Yeah. We don’t independently version this project (at the moment) and the cli command ‘yarn lerna version’ (without adding the version number) makes you go through and version each of the like 6 packages independently. It’s sort of an annoying tedious workflow for me so I would prefer versioning the project together unless/until we actually do choose to version independently.

austinkelleher (approved):

PR #430, opened 2021-03-04
Fixes #429 - "toMatchGraphObjectSchema" fails when multiple classes share same "enum" property

### LINKED ISSUE #429: "toMatchGraphObjectSchema" fails when multiple classes share same "enum" property
## Description

It's possible that two classes contain the same property name and both have "enum" values in the schema. For example, `Host` and `Device` both contain a `platform` property that contain enum values.
 
See here:
- https://github.com/JupiterOne/data-model/blob/master/src/schemas/Host.json#L75
- https://github.com/JupiterOne/data-model/blob/master/src/schemas/Device.json#L58

We should merge the two sets of enum values in this case.

## Sample error

```
schema is invalid: data.properties['platform'].enum should NOT have duplicate items (items ## 7 and 15 are identical)
```

austinkelleher:


mknoedel (dismissed):
Great fix, I actually ran into this in Intune as well, but I fixed it by just removing "Host". Looks like I should add it back in now.

aiwilliams (commented):
🔥

PR #442, opened 2021-03-26
Update the EntityRawData type to be in-line with the persister

### LINKED ISSUE #441: `setRawData()` causes the persister to error when setting an array of data
This issue causes the persister schema validator to not persist these entities.

Meaning a raw data of:
```
[
  {
    name: 'bad-data',
    rawData: [{ bad: 'data' }]
  }
]
```
Causes the persister to throw an error where:
```
[
  {
    name: 'bad-data',
    rawData: { bad: 'data' }
  }
]
```
will be properly persisted

mknoedel:
### Description
The persister does schema validation on the `_rawData` value of entities that is different from this TypeScript type, which causes confusing upload errors like:

```
Error uploading collected data (API response: code=ENTITY_UPLOAD_FAILED_VALIDATION, message="1 validation error(s):
- Error in entities[0]._rawData (reason=Expected { [_: string]: { body: string | { [_: string]: unknown }; contentType: string | undefined; } } | undefined, but was object in _rawData)
").
```

This change will let Typescript tell us if we put in an array instead an object, thus preventing these errors:
Screen Shot 2021-03-26 at 1 06 04 PM

Fixes #441

### DIFF
--- a/packages/integration-sdk-core/src/data/rawData.ts
+++ b/packages/integration-sdk-core/src/data/rawData.ts
@@ -8,10 +8,10 @@ import { EntityRawData, RawDataTracking } from '../types';
  * @param name name of raw data, unique within scope of entity; `'default'`
  * unless otherwise specified
  */
-export function getRawData<T>(
+export function getRawData(
   trackingEntity: RawDataTracking,
   name: string = 'default',
-): T | undefined {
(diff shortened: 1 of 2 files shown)


ndowmon (approved):

mknoedel:
didn't name the branch the right way

PR #444, opened 2021-03-29
AB#4717- Only upload entities and relationships in the step they were created in

mknoedel:
This is fixing an issue were entities from long-running steps were being flushed by later, un-related steps.

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
+++ b/packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
@@ -278,7 +278,6 @@ export function executeStepDependencyGraph<
 
       try {
         await step.executionHandler(context);
-        context.logger.stepSuccess(step);
 
         if (stepHasDependencyFailure(stepId)) {
           status = StepResultStatus.PARTIAL_SUCCESS_DUE_TO_DEPENDENCY_FAILURE;
@@ -297,7 +296,7 @@ export function executeStepDependencyGraph<
         status = StepResultStatus.FAILURE;
       }
 
-      await context.jobState.flush();
+      await context.jobState.flush(stepId);
 
       if (context.jobState.waitUntilUploadsComplete) {
         try {
@@ -305,6 +304,9 @@ export function executeStepDependencyGraph<
           // fatal failure. We just want to make this step as a partial success
           // and move on with our lives!
           await context.jobState.waitUntilUploadsComplete();
+          if (status != StepResultStatus.FAILURE) {
+            context.logger.stepSuccess(step);
+          }
         } catch (err) {
           context.logger.stepFailure(step, err);
           status = StepResultStatus.FAILURE;
--- a/packages/integration-sdk-runtime/src/execution/uploader.ts
+++ b/packages/integration-sdk-runtime/src/execution/uploader.ts
@@ -148,7 +148,7 @@ export function createPersisterApiStepGraphObjectDataUploader({
       const context = jobContextWithUploaderMetadataLogger({
         synchronizationJobContext,
         uploadId: uuid(),
-        stepId: stepId,
+        stepId,
       });
 
       try {
--- a/packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
+++ b/packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
@@ -210,19 +210,21 @@ export class FileSystemGraphObjectStore implements GraphObjectStore {
   async flush(
     onEntitiesFlushed?: (entities: Entity[]) => Promise<void>,
     onRelationshipsFlushed?: (relationships: Relationship[]) => Promise<void>,
+    stepId?: string,
   ) {
     await Promise.all([
-      this.flushEntitiesToDisk(onEntitiesFlushed),
-      this.flushRelationshipsToDisk(onRelationshipsFlushed),
+      this.flushEntitiesToDisk(onEntitiesFlushed, stepId),
+      this.flushRelationshipsToDisk(onRelationshipsFlushed, stepId),
     ]);
   }
 
   async flushEntitiesToDisk(
     onEntitiesFlushed?: (entities: Entity[]) => Promise<void>,
+    onlyThisStepId?: string,
   ) {
     await this.lockOperation(() =>
       pMap(
-        this.localGraphObjectStore.collectEntitiesByStep(),
+        this.localGraphObjectStore.collectEntitiesByStep(onlyThisStepId),
         async ([stepId, entities]) => {
           const indexable = entities.filter((e) => {
             const indexMetadata = this.getIndexMetadataForGraphObjectType({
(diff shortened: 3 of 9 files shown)


packages/integration-sdk-runtime/src/execution/uploader.ts
```diff
@@ -148,7 +148,7 @@ export function createPersisterApiStepGraphObjectDataUploader({
       const context = jobContextWithUploaderMetadataLogger({
         synchronizationJobContext,
         uploadId: uuid(),
-        stepId: stepId,
+        stepId,
```
mknoedel, line 151: Per @austinkelleher's request

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
```diff
@@ -129,6 +129,25 @@ describe('flushEntitiesToDisk', () => {
...
+    await store.addEntities(stepIdA, entitiesA);
+
+    const stepIdB = 'b';
+    const entityTypeB = uuid();
+    const entitiesB = times(3, () => createTestEntity({ _type: entityTypeB }));
+    await store.addEntities(stepIdB, entitiesB);
+
+    await store.flushEntitiesToDisk(
+      // eslint-disable-next-line @typescript-eslint/require-await
+      async (entities) => expect(entities).toEqual(entitiesA),
```
mknoedel, line 147: This errors without the change as it also uploads `entitiesB`

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
```diff
@@ -876,8 +918,10 @@ describe('flush callbacks', () => {
   });
 });
 
-function setupFileSystemObjectStore(params?: FileSystemGraphObjectStoreParams) {
-  const storageDirectoryPath = uuid();
+function setupFileSystemObjectStore(
+  params?: FileSystemGraphObjectStoreParams,
+  storageDirectoryPath = uuid(),
```
mknoedel, line 923: storagePath and step-id are the same

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -210,19 +210,21 @@ export class FileSystemGraphObjectStore implements GraphObjectStore {
...
-      this.flushEntitiesToDisk(onEntitiesFlushed),
-      this.flushRelationshipsToDisk(onRelationshipsFlushed),
+      this.flushEntitiesToDisk(onEntitiesFlushed, stepId),
+      this.flushRelationshipsToDisk(onRelationshipsFlushed, stepId),
     ]);
   }
 
   async flushEntitiesToDisk(
     onEntitiesFlushed?: (entities: Entity[]) => Promise<void>,
+    onlyThisStepId?: string,
```
austinkelleher, line 223: Thoughts on renaming this to just `stepId`? I think it naturally implies that it would be specific to a step. That would be the same verbiage we've used everywhere else.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
```diff
@@ -129,6 +129,25 @@ describe('flushEntitiesToDisk', () => {
...
+    const entitiesA = times(3, () => createTestEntity({ _type: entityTypeA }));
+    await store.addEntities(stepIdA, entitiesA);
+
+    const stepIdB = 'b';
+    const entityTypeB = uuid();
+    const entitiesB = times(3, () => createTestEntity({ _type: entityTypeB }));
+    await store.addEntities(stepIdB, entitiesB);
+
+    await store.flushEntitiesToDisk(
+      // eslint-disable-next-line @typescript-eslint/require-await
```
austinkelleher, line 146: How about something like this, so that the test will fail if the callback was never called:

```typescript
const onEntitiesFlushedFn = jest.fn().mockResolvedValue(Promise.resolve());
await store.flushEntitiesToDisk(onEntitiesFlushedFn);

expect(onEntitiesFlushedFn).toHaveBeenCalledTimes(1);
expect(onEntitiesFlushedFn).toHaveBeenCalledWith(entitiesA);
```

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -297,14 +296,17 @@ export function executeStepDependencyGraph<
...
-      await context.jobState.flush();
+      await context.jobState.flush(stepId);
 
       if (context.jobState.waitUntilUploadsComplete) {
         try {
           // Failing to upload all integration data should not be considered a
           // fatal failure. We just want to make this step as a partial success
           // and move on with our lives!
           await context.jobState.waitUntilUploadsComplete();
+          if (status != StepResultStatus.FAILURE) {
```
austinkelleher, line 307: `!==`?

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -278,7 +278,6 @@ export function executeStepDependencyGraph<
 
       try {
         await step.executionHandler(context);
-        context.logger.stepSuccess(step);
```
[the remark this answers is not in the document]
austinkelleher: 💯

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
mknoedel: Yeah, that's more explicit. Good catch

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
mknoedel: works for me

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -210,19 +210,21 @@ export class FileSystemGraphObjectStore implements GraphObjectStore {
...
   ) {
     await this.lockOperation(() =>
       pMap(
```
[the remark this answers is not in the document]
mknoedel: I'm good with either as I think there may eventually be a benefit to maintaining the ability for a full flush like on errors or at the end of a process or something like that.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
mknoedel: More explicit. Much like. Will change.

mknoedel:
After talking with @austinkelleher and @aiwilliams, we might not want to do these changes as in sdk v3 we are going to put a much stronger focus on streaming entities and anything tieing us to the existing step system might just add more to untangle later.

We determined that the extra debugging and joblog miscommunication is worth less than the ease of building sdk v3, so closing for now.

PR #454, opened 2021-04-13
Make jobState.findEntity() more efficient

ndowmon:
This is an incremental change. Notably, it doesn't add the entity index _within_ the file, which has been a topic of conversation.


packages/integration-sdk-core/src/types/storage.ts
```diff
@@ -30,8 +30,13 @@ export interface GraphObjectStore {
     onRelationshipsFlushed?: (relationships: Relationship[]) => Promise<void>,
   ): Promise<void>;
 
+  /**
+   * @deprecated Use findEntity
+   */
   getEntity({ _key, _type }: GraphObjectLookupKey): Promise<Entity>;
 
+  findEntity(_key: string): Promise<Entity | undefined>;
```
ndowmon, line 38: Officially added `findEntity` to the `GraphObjectStore` spec, and deprecated `getEntity`.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -98,6 +102,56 @@ function integrationStepsToGraphObjectIndexMetadataMap(
...
+ *   { location: 'inMemory' }
+ *
+ * Entities that have been flushed to disk have a location of
+ *   { location: 'onDisk', graphDataPath: '/graph/step-id/filename.json' }
+ *
+ * Finally, entities that have been flushed from the InMemoryGraphObjectStore, but that
+ * have `indexMetadata.enabled = false`, are explicitly not findable. These have a
+ * location of
+ *   { location: 'nonIndexable' }
+ */
```
ndowmon, line 149: I wanted to design this to be a holistic view into where entities exist in this system, including whether they are `inMemory` _or_ `nonIndexable`. However, this map will be larger than it needs to be for that purpose. An alternative is to _only_ maintain a map of entities on disk (since finding them in memory is trivial, and we may not care if they have not been indexed on disk).

Would love to know all your thoughts on this.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -107,6 +161,7 @@ export class FileSystemGraphObjectStore implements GraphObjectStore {
     string,
     GraphObjectIndexMetadataMap
   >;
+  private readonly entityLocationMap = new Map<string, GraphObjectLocation>();
```
austinkelleher, line 164: Change this to `BigMap`

mknoedel (dismissed):
Looks great! I had some implementation comments, but reguardless this will be a good help.

austinkelleher (approved):
Nice one!

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
aiwilliams: I agree that we do not need to maintain a mapping of things in memory or not indexed or anywhere else. I'd like to see the FileSystemGraphObjectStore encapsulate its management of an index of its data; nothing outside the object should know what it's doing to be fast as what it does. The code that uses an in memory store and flushes to disk when that gets full could be a `BufferedGraphObjectStore` that writes to an in memory store and flushes to a file system store, but provides a single interface for `findEntity`.

aiwilliams (approved):
Sweet. Looking forward to seeing how an integration goes from slow to fast thank to these changes!

PR #457, opened 2021-04-16
Add yarn clean:build script

ndowmon:
If packages in this project got out-of-sync, rebuilding was really confusing. Added these scripts to ease development.


austinkelleher (approved):
Looks good to me. I think eventually we will want to document these npm scripts.

PR #471, opened 2021-05-21
Fixes #470 - Introduce beforeAddEntity hook into IntegrationInvocationConfig

### LINKED ISSUE #470: Introduce `beforeAddEntity` hook into `IntegrationInvocationConfig`
Sometimes you want to be able to add a specific property to _all_ entities. This can add a significant amount of code churn to a project. It would be great if there is a hook that could be exposed and called before an entity is added to the job state via `jobState.addEntity` or `jobState.addEntities`.

In the https://github.com/jupiterone/graph-google-cloud project, we would like to add the `projectId` property to all entities, but there isn't an easy method of doing that without changing a lot of code in the project.

Example usage:

```typescript
export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> = {
  instanceConfigFields: {},
  integrationSteps: [],

  beforeAddEntity(context: IntegrationStepContext, entity: Entity): Entity {
    const projectId = context.instance.config.projectId;

    return {
      ...entity,
      projectId: entity.projectId || projectId
    };
  }
};
```

austinkelleher:
See #470

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to
 
 ## Unreleased
 
+- [#470](https://github.com/JupiterOne/sdk/issues/470) Introduce
+  `beforeAddEntity` hook into `IntegrationInvocationConfig`. See the development
+  documentation for more information on its usage.
+
 ## 6.1.0 - 2021-05-19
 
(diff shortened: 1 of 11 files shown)


mknoedel (approved):
Looks great!

PR #476, opened 2021-06-01
Add expect(mappedRelationships).toTargetEntities(entities) matcher

ndowmon:
Currently using this matcher in the `graph-azure` project.

### DIFF
--- a/packages/integration-sdk-testing/src/__tests__/jest.test.ts
+++ b/packages/integration-sdk-testing/src/__tests__/jest.test.ts
@@ -1,7 +1,14 @@
-import { Entity, ExplicitRelationship } from '@jupiterone/integration-sdk-core';
+import {
+  createIntegrationEntity,
+  createMappedRelationship,
+  Entity,
+  ExplicitRelationship,
+  RelationshipClass,
+} from '@jupiterone/integration-sdk-core';
 import {
   toMatchGraphObjectSchema,
   toMatchDirectRelationshipSchema,
+  toCreateValidRelationshipsToEntities,
   GraphObjectSchema,
   registerMatchers,
 } from '../jest';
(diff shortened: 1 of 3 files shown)


packages/integration-sdk-testing/src/jest.ts
```diff
@@ -224,6 +230,72 @@ export interface ToMatchGraphObjectSchemaParams {
...
+declare global {
+  // eslint-disable-next-line @typescript-eslint/no-namespace
+  namespace jest {
+    interface Matchers<R> {
+      toCreateValidRelationshipsToEntities(entities: Entity[]): R;
+    }
+  }
+}
+
+export function toCreateValidRelationshipsToEntities(
```
aiwilliams, line 242: I'm not feeling the name... 🤔 

1. Looking at [where it's used in Azure integration](https://github.com/JupiterOne/graph-azure/blob/79372d18098ed036105386a417c0c8331c3ea03f/src/steps/resource-manager/authorization/index.test.ts#L236), seems like it should be `expect(mappedRelationships).toTargetEntities([...])`. This would reveal the fact that this does not support other kinds of relationships, which took me some time to realize.

2. I'm unsure whether this matcher is meant to ensure that the `jobState` contains the expected entities, since in the Azure use case, it looks like this:

```ts
    const {
      userEntities,
      groupEntities,
      servicePrincipalEntities,
      roleAssignmentEntities,
    } = await getSetupEntities(configFromEnv);
    expect(mappedGroupRelationships).toCreateValidRelationshipsToEntities(
      groupEntities,
    );
```

3. Though the code is attempting to enforce that there is only one entity in the set to which `targetFilterKeys` would apply, IF the intention is to verify that the set of mapped relationships in the job state target something in the set of job state entities, it would seem reasonable to find multiple matches.

ndowmon: Great suggestions @aiwilliams. 

1. I agree with the naming convention, thanks for suggesting.
2. In the case of azure, the target entities do not exist in the jobState. They could, however.
3. Good suggestion. I'll change this to require _at least one_ relationship.

ndowmon: I ended up adding the option `enforceSingleTarget` to _allow_ the case for exactly one target.

packages/integration-sdk-testing/src/__tests__/jest.test.ts
```diff
@@ -583,6 +590,166 @@ Find out more about JupiterOne schemas: https://github.com/JupiterOne/data-model
   });
 });
 
+describe('#toCreateValidRelationshipsToEntities', () => {
```
aiwilliams, line 593:
```suggestion
describe('#toTargetEntities', () => {
```

packages/integration-sdk-testing/src/__tests__/jest.test.ts
```diff
@@ -583,6 +590,166 @@ Find out more about JupiterOne schemas: https://github.com/JupiterOne/data-model
...
+  test('should pass if mapped relationships match entities', () => {
```
aiwilliams, line 594:
```suggestion
  test('should pass if mapped relationship targets any entities', () => {
```

packages/integration-sdk-testing/src/__tests__/jest.test.ts
```diff
@@ -583,6 +590,166 @@ Find out more about JupiterOne schemas: https://github.com/JupiterOne/data-model
...
+      ],
+    );
+
+    expect(result).toEqual({
+      message: expect.any(Function),
+      pass: true,
+    });
+  });
+
+  test('should pass if mapped relationships match more than one entities', () => {
```
aiwilliams, line 632:
```suggestion
  test('should pass if mapped relationship targets multiple entities', () => {
```

packages/integration-sdk-testing/src/__tests__/jest.test.ts
```diff
@@ -583,6 +590,166 @@ Find out more about JupiterOne schemas: https://github.com/JupiterOne/data-model
...
+      [targetEntity, targetEntity],
+    );
+
+    expect(result).toEqual({
+      message: expect.any(Function),
+      pass: true,
+    });
+  });
+
+  test('should fail if mapped relationships match more than one entities and enforceSingleTarget = true', () => {
```
aiwilliams, line 670:
```suggestion
  test('should fail if mapped relationship targets more than one entity and enforceSingleTarget = true', () => {
```

packages/integration-sdk-testing/src/__tests__/jest.test.ts
```diff
@@ -583,6 +590,166 @@ Find out more about JupiterOne schemas: https://github.com/JupiterOne/data-model
...
+    expect(result).toEqual({
+      message: expect.any(Function),
+      pass: false,
+    });
+    expect(result.message()).toMatch(
+      'Multiple target entities found for mapped relationship, expected exactly one: {',
+    );
+  });
+
+  test('should fail if mapped relationships do not match entities', () => {
```
aiwilliams, line 712:
```suggestion
  test('should fail if mapped relationship targets unknown entities', () => {
```

aiwilliams (changes requested):
A rename not caught in the test description; made some suggestions for descriptions.

aiwilliams (approved):
💯  I think we'll need a testing guide soon that features things like this for those who will not read a CHANGELOG and/or are new to building integrations.

PR #478, opened 2021-06-02
add diffing functionality to the sdk

mknoedel:
Usage:

This is used when migrating integrations from the old sdk to the new sdk to ensure there are no unintended changes.

Helper text:

```
Usage: compare [options] <oldPath> <neoPath>

Compare the differences between two datasets that were downloaded from JupiterOne via the query: 

         find * as e that RELATES TO as r * return e.*, r.*

Example Dataset: 

{
    "type": "table",
    "data": [
        {
            "e._objectType": "entity",
            "e._key": "12345",
            "e.name": "Name"
        }
    ],
    "totalCount": 1
}

Arguments:

  oldPath                          relitive path to the dataset to compare against
  neoPath                          relitive path to the new dataset

Options:
  -logkey, --log-only-key-changes  Only log entities and relationships that have _key values that are not in the other dataset.
  -h, --help                       display help for command
```

Output:
```
oldPath: ./.old.json
neoPath: ./.neo.json
logOnlyKeyChanges: undefined
{ oldKeys: 1, newKeys: 1 }
differences:  [
  {
    "key": "jupiterone/token-service/pull-requests/14",
    "mappedRelationship": false,
    "diff": {
      "name": {
        "__old": "Name 2",
        "__new": "Name 1"
      }
    }
  }
]
```

This could be a lot better, but it is useful enough for us to get through the few remaining integrations we still need to port to the new sdk.

### DIFF
--- a/packages/integration-sdk-cli/src/commands/compare.ts
+++ b/packages/integration-sdk-cli/src/commands/compare.ts
@@ -29,5 +29,3 @@ export function compare() {
       findDifferences(oldPath, neoPath, options.logOnlyKeyChanges)
     });
 }
-
-compare().parse()
--- a/packages/integration-sdk-cli/src/commands/compare.ts
+++ b/packages/integration-sdk-cli/src/commands/compare.ts
@@ -1,0 +1,33 @@
+import { createCommand } from 'commander';
+import * as log from '../log';
+import { findDifferences } from '../utils/diff';
+
+export function compare() {
+  return createCommand('compare')
+    .storeOptionsAsProperties()
+    .arguments('<oldPath> <neoPath>')
+    .description('Compare the differences between two datasets that were downloaded from JupiterOne via the query: \n\n' +
+
+    '\t find * as e that RELATES TO as r * return e.*, r.*\n\n' +
+
+    'Example Dataset: \n\n' +
+
+    JSON.stringify(require('../utils/exampleDataset.json'), null, 4)
+    
+    , {
+      oldPath: 'relitive path to the dataset to compare against',
+      neoPath: 'relitive path to the new dataset'
+    })
+    .option(
+      '-logkey, --log-only-key-changes',
+      'Only log entities and relationships that have _key values that are not in the other dataset.'
+    )
+    .action((oldPath, neoPath, options) => {
+      log.info(`oldPath: ${oldPath}`);
+      log.info(`neoPath: ${neoPath}`);
+      log.info(`logOnlyKeyChanges: ${options.logOnlyKeyChanges}`)
+      findDifferences(oldPath, neoPath, options.logOnlyKeyChanges)
+    });
+}
+
+compare().parse()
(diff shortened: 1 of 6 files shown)


packages/integration-sdk-cli/src/commands/compare.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    '\t find * as e that RELATES TO as r * return e.*, r.*\n\n' +
+
+    'Example Dataset: \n\n' +
+
+    JSON.stringify(require('../utils/exampleDataset.json'), null, 4)
+    
+    , {
+      oldPath: 'relitive path to the dataset to compare against',
+      neoPath: 'relitive path to the new dataset'
```
ndowmon, line 19:
```suggestion
      oldPath: 'relative path to the dataset to compare against',
      neoPath: 'relative path to the new dataset'
```

packages/integration-sdk-cli/src/commands/compare.ts
```diff
@@ -0,0 +1,31 @@
...
+    })
+    .option(
+      '-logkey, --log-only-key-changes',
```
ndowmon, line 22: Short versions of CLI args are typically a single char (e.g. `-l`) but there's nothing technically wrong with `-logkey` as the short version.

packages/integration-sdk-cli/src/commands/compare.ts
```diff
@@ -0,0 +1,31 @@
...
+      'Only log entities and relationships that have _key values that are not in the other dataset.'
```
ndowmon, line 23: Is "other" the _old_ or _new_ dataset in this case?

packages/integration-sdk-cli/src/utils/diff.ts
```diff
@@ -0,0 +1,84 @@
+import fs from "fs";
+import { diff as diffJson } from 'json-diff'
+
+export function findDifferences(oldJsonDataPath, newJsonDataPath, logOnlyKeyChanges) {
+
+  const neo = JSON.parse(fs.readFileSync(newJsonDataPath, 'utf8'))
```
ndowmon, line 6: Just curious - why use `neo` and not `new`?

ndowmon (approved):

packages/integration-sdk-cli/src/utils/diff.ts
mknoedel: New is a keyword. Couldn't think of a better word at the time.

packages/integration-sdk-cli/src/commands/compare.ts
mknoedel: Another way to put it would be that it will only log if the key is only in one dataset or the other, but not in both.

packages/integration-sdk-cli/src/utils/diff.ts
ndowmon: 👍

ndowmon (approved):

PR #482, opened 2021-06-09
Allow `findEntity()` to accept `string | undefined`

ndowmon:
Very often I find myself using the non-null assertion operator on properties I'm passing into the `jobState.findEntity()` method. Using non-null assertion is an antipattern in typescript IMO and this method should _explicitly_ handle the `undefined` case for safety's sake.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ and this project adheres to
   ]);
   ```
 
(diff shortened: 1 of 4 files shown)


mknoedel (approved):
Nice 💯

kenanwarren (approved):

PR #484, opened 2021-06-13
Document step patterns

ndowmon:


docs/integrations/step-patterns.md
```diff
@@ -0,0 +1,326 @@
...
+  - `acme_device`
+  - `acme_device_application`
+- Relationships
+  - `acme_account_has_user`
+  - `acme_account_has_group`
+  - `acme_group_has_user`
+  - `acme_user_has_device`
+  - `acme_device_has_application`
+
+In order design the optimal mapping of steps to their entities & relationships,
```
aiwilliams, line 33: In order to

docs/integrations/step-patterns.md
```diff
@@ -0,0 +1,326 @@
...
+
+## "Fetch Relationships" step pattern
+
+Endpoints that list resources using path parameters, such as
+`/groups/{groupId}/users`, typically indicate either the **Fetch Relationships**
+pattern or the **Fetch Child Entities** pattern, described below. Target
+resources (in this case, `acme_user`) otherwise listed from a top-level **list**
+endpoint (like `/users`), signal that JupiterOne entities will be created
+elsewhere using a **Fetch Entities** pattern and that this endpoint should be
+handled with a **Fetch Relationships** pattern.
```
aiwilliams, line 173: This looks incomplete

aiwilliams (commented):
This is great, thanks for taking the time to get something like this going!

I'm unsure about the use of the term "child", but I don't have a better name to suggest. My main hangup is that it seems the patterns using the term "child" can apply when resources don't fit an ancestry metaphor, but a containment or association (many to many) relationship.

docs/integrations/step-patterns.md
```diff
@@ -0,0 +1,326 @@
...
+```typescript
+const fetchAccountStep: Step = {
+  id: 'fetch-account',
+  name: 'Fetch Account',
+  entities: [{ _type: 'acme_account', ... }],
+  relationships: [],
+  dependsOn: [],
+  executionHandler: (context) => {
+    const { jobState, instance } = context;
+    const client = new AcmeClient({ config: instance.config });
```
mknoedel, line 116: I think we might want to start using a getter pattern for clients instead of creating a new client for each step in order to better control rate and abuse limits.

docs/integrations/step-patterns.md
```diff
@@ -0,0 +1,326 @@
...
+```typescript
+const buildUserDeviceRelationships: Step = {
+  id: 'build-user-device-relationships',
+  name: 'Build User to Device Relationships',
+  entities: [],
+  relationships: [{ _type: 'acme_user_has_device', ... }],
+  dependsOn: ['fetch-users', 'fetch-devices'],
+  executionHandler: (context) => {
+    const { jobState, instance } = context;
+    const client = new AcmeClient({ config: instance.config });
```
mknoedel, line 299: Wouldn't we just make this relationship when fetching the device, as this client isn't being used in this example?

mknoedel (commented):
Thanks for making this!

docs/integrations/step-patterns.md
ndowmon: Yes, great catch!

docs/integrations/step-patterns.md
ndowmon: That's a good thought. I'll leave this as is, since the focus is more on everything below this, but I definitely agree with you.

docs/integrations/step-patterns.md
ndowmon: 👍

ndowmon:
>I'm unsure about the use of the term "child", but I don't have a better name to suggest. My main hangup is that it seems the patterns using the term "child" can apply when resources don't fit an ancestry metaphor, but a containment or association (many to many) relationship.
I am operating under the assumption that when _not_ appropriate for an ancestry metaphor, it's best to use the **Fetch Relationships** pattern described elsewhere. In this case, I'm actually operating under the assumption that the `acme_device_application` entities are all direct children of the `acme_device`... i.e. even if multiple devices have an application called **Google Chrome**, for example, they would be treated as direct descendants of `acme_device` and separate entities in the graph. Perhaps I've assumed too much here. 

Would love to talk more about this (and naming conventions in general) when you've got time.

@aiwilliams

docs/integrations/step-patterns.md
```diff
@@ -0,0 +1,326 @@
...
+pattern or the **Fetch Child Entities** pattern, described below. Target
+resources (in this case, `acme_user`) otherwise listed from a top-level **list**
+endpoint (like `/users`), signal that JupiterOne entities will be created
+elsewhere using a **Fetch Entities** pattern and that this endpoint should be
+handled with a **Fetch Relationships** pattern.
+
+This type of step ingests no new entities, and instead builds relationships
+between existing entities. This pattern uses the `jobState.iterateEntities` and
+`jobState.findEntity` methods to reference entities that have already been added
+to the job state in previous steps. It's important that both the `fetch-groups`
```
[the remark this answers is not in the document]
ndowmon: Let's chat more about this in the future. I think verifying that the `_key` exists in the job state before building the relationship is very useful. Maybe it's better to use `jobState.hasKey()` rather than `jobState.findEntity` in that case.

PR #492, opened 2021-06-23
Add j1-integration diff command for colorized diffs

ndowmon:
This builds on top of `j1-integration compare` but uses the exported `diffString()` function from `json-diff`, which gives pretty colorized diffs

```diff

 {
   tenable_user_2232707: {
     properties: {
-      id: 2232707
+      id: "2232707"
     }
   }
   tenable_scan_19: {
     properties: {
-      id: 19
+      id: "19"
     }
   }
   tenable_scan_13: {
     properties: {
-      id: 13
+      id: "13"
     }
   }
   tenable_scan_17: {
     properties: {
-      id: 17
+      id: "17"
     }
   }
   tenable_scan_15: {
     properties: {
-      id: 15
+      id: "15"
     }
   }
...
```

### DIFF
--- a/packages/integration-sdk-cli/package.json
+++ b/packages/integration-sdk-cli/package.json
@@ -36,6 +36,7 @@
     "@pollyjs/adapter-node-http": "^4.0.4",
     "@pollyjs/core": "^4.0.4",
     "@pollyjs/persister-fs": "^4.0.4",
+    "@types/json-diff": "^0.5.1",
     "@types/lodash": "^4.14.158",
     "@types/pollyjs__adapter-node-http": "^2.0.0",
     "@types/pollyjs__core": "^4.0.0",
--- a/packages/integration-sdk-cli/src/commands/diff.ts
+++ b/packages/integration-sdk-cli/src/commands/diff.ts
@@ -1,0 +1,197 @@
+import { createCommand } from 'commander';
+import fs from 'fs';
+import { diffString } from 'json-diff';
+
+declare module 'json-diff' {
+  /**
+   * The exported types from @types/json-diff exclude an optional fourth parameter, `diffOptions`,
+   * that allows the user to pass the `keysOnly` flag. Sadly, based on testing, it seems that
+   * `keysOnly` does not work under certain circumstances, particularly with diffed arrays.
+   *
+   * Method signature as exported from @types/json-diff:
+   *   function diffString(obj1: unknown, obj2: unknown, colorizeOptions?: ColorizeOptions): string;
+   */
+  function diffString(
+    obj1: unknown,
+    obj2: unknown,
+    colorizeOptions?: { color?: boolean },
+    diffOptions?: { keysOnly?: boolean },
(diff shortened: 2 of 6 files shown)


packages/integration-sdk-cli/src/commands/diff.ts
```diff
@@ -0,0 +1,197 @@
...
+  ): string;
+}
+
+export function diff() {
+  return createCommand('diff')
+    .storeOptionsAsProperties()
+    .arguments('<oldExportPath> <newExportPath>')
+    .description(
+      'Compare two datasets downloaded from JupiterOne using: \n\n' +
+        '\t find * with _integrationInstanceId="abc-123" that RELATES TO * return tree\n\n',
```
ndowmon, line 28: Keen observers will notice that this J1QL should _actually_ use an optional relationship, in order to capture any entities that have no edges:
```
FIND * with _integrationInstanceId='integration-instance-id' (THAT RELATES TO *)? RETURN TREE
```

However, I tested this in dev, and I'd say that the non-optional traversal query may take ~5 seconds, and the optional traversal query may take well over a minute. For that reason, I've chosen _not_ to use the optional traversal query in this description.

Anyone disagree? Think that this should include the optional traversal?

packages/integration-sdk-cli/package.json
```diff
@@ -36,6 +36,7 @@
     "@pollyjs/adapter-node-http": "^4.0.4",
     "@pollyjs/core": "^4.0.4",
     "@pollyjs/persister-fs": "^4.0.4",
+    "@types/json-diff": "^0.5.1",
```
kenanwarren, line 39: I know we've probably had this package in the sdk for awhile, but it looks unmaintained so it might be something worth migrating away from in the future.

kenanwarren (approved):

packages/integration-sdk-cli/package.json
ndowmon: Seems that maybe this package would be preferred (https://www.npmjs.com/package/diff)

`Diff.diffJson()`

kenanwarren: Yeah that might work. Sadly it doesn't look like there are any that either have 0 dependencies or keep their dependencies maintained.

packages/integration-sdk-cli/src/commands/diff.ts
aiwilliams: Very good find @ndowmon. I had originally assumed that all entities are related to something, essentially making the `Account` entity of the integration the root node. However, that's certainly not true all the time. Perhaps we could use two queries to make it faster and to separate entities from relationships?

1. `FIND * WITH _integrationInstanceId='integration-instance-id'`
2. `FIND * WITH _integrationInstanceId='integration-instance-id' THAT RELATES TO AS r * RETURN r.*`

Otherwise, perhaps it's just fine to wait a minute since we're looking to validate data, not run the query a lot of times and wait.

PR #500, opened 2021-06-30
Added deleteData function to jobstate

ceelias:


packages/integration-sdk-core/src/types/jobState.ts
```diff
@@ -40,6 +40,13 @@ export interface JobState {
    */
   getData: <T>(key: string) => Promise<T | undefined>;
 
+  /**
+   * Delete arbitrary data stored by parent steps.
+   *
```
ceelias, line 45: Wasn't too sure what should be included here. Suggestions welcome

ndowmon (dismissed):
Looks great, Chris!

kenanwarren (approved):

packages/integration-sdk-runtime/src/execution/jobState.ts
```diff
@@ -82,6 +82,10 @@ export class MemoryDataStore {
   get(key: string): unknown {
     return this.data[key];
   }
+
+  delete(key: string): void {
+    this.data[key] = undefined;
```
austinkelleher, line 87: I think we should actually delete this key from memory completely instead of assigning the value to `undefined`. 

```js
delete this.data[key];
```

austinkelleher (changes requested):
Thanks for working on this! I left a small request.

austinkelleher (approved):

PR #501, opened 2021-06-30
Known Vendors

erkangz:
Putting this in the SDK should allow us to more consistent map Vendor entities in different integrations, and pre-populate useful vendor information such as privacy policy. for example:

- in Google integration: https://github.com/JupiterOne/graph-google/blob/master/src/steps/tokens/index.ts#L51
- in Okta integration: https://github.com/JupiterOne/graph-okta/blob/master/src/util/knownVendors.ts#L3


erkangz:
going to put this in a new `vendor-stack` repo

PR #508, opened 2021-07-14
Release 6.11.0

mknoedel:
This change adds the ability to, in the integrationConfig, specify multiple dependency graphs that each step can be assigned to. These graphs will then be executed in the order specified.

This maintains backwards compatibility with existing integrations while allowing us to specify an integration step as needing to be last for the google-cloud role binding steps.

NOTE: I didn't need to change anything for local execution because of one side-effect of this change. If your step definition `dependsOn` a step that is no longer in your dependency graph, you will get a "Node does not exist" error. I thought this was a good way to handle this case so I left it as is, but if we want that to change and instead log a warning, then I would add back my changes to `prepareLocalStepCollection` that I removed after noticing they didn't do anything because of this behavior 🙃

### DIFF
--- a/packages/integration-sdk-core/src/types/step.ts
+++ b/packages/integration-sdk-core/src/types/step.ts
@@ -151,4 +151,20 @@ export type StepMetadata = StepGraphObjectMetadataProperties & {
    * before the current step can.
    */
   dependsOn?: string[];
+
+  /**
+   * An optional array of identifiers used to execute dependency
+   * graphs in a specific order. These values should match the
+   * IntegrationInvocationConfig `dependencyGraphOrder`
+   * prpoperty.
+   * 
+   * Steps that do not have a `dependencyGraphId` will be added to
+   * the default dependency graph which is executed first.
+   * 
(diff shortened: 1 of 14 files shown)


mknoedel:
relates to https://github.com/JupiterOne/sdk/pull/507

austinkelleher (approved):

packages/integration-sdk-core/src/types/step.ts
```diff
@@ -151,4 +151,20 @@ export type StepMetadata = StepGraphObjectMetadataProperties & {
    * before the current step can.
    */
   dependsOn?: string[];
+
+  /**
+   * An optional array of identifiers used to execute dependency
+   * graphs in a specific order. These values should match the
+   * IntegrationInvocationConfig `dependencyGraphOrder`
+   * prpoperty.
```
aiwilliams, line 159: This is a single `string`, not an `Array`.

aiwilliams (commented):
Just got to reading this, love the simplicity! Great work.

PR #509, opened 2021-07-16
Use J1 colors in j1-integration visualize-types

ndowmon:
While working on a blog post, I noticed that the output of `j1-integration visualize-types` provides good content, but I didn't think screenshots would not look great in J1 marketing content because of the color scheme. Updated to use J1 primary colors.

<h2>Before</h2>
Screen Shot 2021-07-16 at 10 09 39 AM

<h2>After</h2>
Screen Shot 2021-07-16 at 10 09 49 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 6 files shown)


austinkelleher (approved):
Lit

aiwilliams:
Niiice.

PR #515, opened 2021-08-02
Trim properties from placeholder entities created from j1-integration visualize

ndowmon:
Previously, the "placeholder entities" generated from j1-integration visualize included all of the properties in `targetEntity`, but in reality, target entities are unique only in their `targetFilterKeys`. 

Previous: 
Screen Shot 2021-08-02 at 4 34 32 PM

Now: 
Screen Shot 2021-08-02 at 4 29 22 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to
 
 ## Unreleased
 
+### Changed
+
+- Changed how `j1-integration visualize` displays placeholder entities. Now only
+  properties present in `targetFilterKeys` are displayed in the graph, making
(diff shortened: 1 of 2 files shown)


austinkelleher (approved):
You mind adding a test for this?

https://github.com/JupiterOne/sdk/blob/afaa8ffa61bd9e0356226924ec0b513981d288dd/packages/integration-sdk-cli/src/visualization/createMappedRelationshipNodesAndEdges.ts

ndowmon:
Added tests here: https://github.com/JupiterOne/sdk/pull/521

PR #533, opened 2021-08-20
Document command now documents mapped relationships from step metadata

ceelias:
Now that we have the ability to declare mapped relationships in the step meta data, this PR makes the document command generate docs for these mapped relationships

Example Output:

# Integration with JupiterOne

## Setup

In this section, please provide details about how to set up the integration with
JupiterOne. This may require provisioning some resources on the provider's side
(perhaps a role, app, or api key) and passing information over to JupiterOne.

### Entities

The following entities are created:

| Resources | Entity \`_type\` | Entity \`_class\` |
| --------- | -------------- | --------------- |
| The Group | \`my_group\`     | \`Group\`         |

### Mapped Relationships

The following mapped relationships are created:

| Source Entity \`_type\` | Relationship \`_class\` | Target Entity \`_type\` | Direction |
| --------------------- | --------------------- | --------------------- | --------- |
| \`my_group\`            | **HAS**               | \`*your_user*\`         | FORWARD   |

### DIFF
--- a/packages/integration-sdk-cli/src/commands/document.ts
+++ b/packages/integration-sdk-cli/src/commands/document.ts
@@ -5,6 +5,7 @@ import {
   StepGraphObjectMetadataProperties,
   StepEntityMetadata,
   StepRelationshipMetadata,
+  StepMappedRelationshipMetadata,
 } from '@jupiterone/integration-sdk-core';
 import { promises as fs } from 'fs';
 import {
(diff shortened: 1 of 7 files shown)


packages/integration-sdk-cli/src/commands/document.ts
```diff
@@ -146,11 +147,33 @@ function generateRelationshipTableFromAllStepEntityMetadata(
...
+      'Source Entity `_type`',
+      'Relationship `_class`',
+      'Target Entity `_type`',
+      'Direction',
+    ],
+    ...metadata.map((v) => [
+      `\`${v.sourceType}\``,
+      `**${v._class}**`,
+      `\`*${v.targetType}*\``,
+      `${v.direction}`,
```
ceelias, line 164: Thoughts on the order and how this looks?

ndowmon: I like it. The _italics_ are nice for the target.

ndowmon (approved):
Looks great! Thank you Chris!

austinkelleher (approved):
Nice change!

Issue #288, opened 2020-08-18
Remove `sourceEntityType` property from `RelationshipMapping` interface


ndowmon:
#62 was introduced a little while ago to allow a `sourceEntityType` property on the `RelationshipMapping` interface. Little did I know that this change did not adhere to the runtypes spec, causing the following error in some mapped relationships created using `v2.6.0` - `v2.10.0`.

```
(8:09:03 AM) - [job_abort] - Aborting integration job. (reason=Error uploading collected data (API response: code=RELATIONSHIP_UPLOAD_FAILED_VALIDATION, message="1 validation error(s): - Error in relationships[0]._mapping (reason=Following properties are not allowed: [sourceEntityType] in _mapping) (code=RELATIONSHIP_UPLOAD_FAILED_VALIDATION, statusCode=400, correlationId=4789d8dc-c64b-4bd0-80f8-989dfaf26a9e)").)
```

Issue #67, opened 2020-04-22
Add `addEntity`, `addRelationship` (singular forms of existing functions) to storage interface


aiwilliams:
I think we should consider adding `addEntity` and `addRelationship` to the storage interface. This could help to avoid a pattern where folks end up creating collector variables, which just end up keeping unnecessary references around longer than needed (memory bloat in the frame). That is, I can see folks collecting all the things and calling `addEntities/addRelationships` just once at the end of a step, thereby negating the benefits of the flushing the store does. Of course, we should also consider adding guidance that encourages working on data in small batches.

PR #536, opened 2021-08-24
Ensure arrays are not too large

mknoedel:
Created in response to - [Google Cloud #288](https://github.com/JupiterOne/graph-google-cloud/pull/288)

Neptune has an undocumented cap of how many array elements it can store in a single edge or vertex. Through some investigation, @softwarewright found this limit to be around 1,500 elements. 

The google cloud integration was sometimes attempting to store permission arrays that were more than 1,500 elements. Those were causing an error in the persister which was not bubbling back to the integration, so these objects were not getting stored.

In order to avoid the confusion, it seems better to add this check in the sdk to catch these issues early.

This check was only added to entities because the persister handles array properties differently between Entities and Relationships -- see [Persister #67](https://github.com/JupiterOne/jupiter-persister/pull/67). I am unsure why the persister handles these differently.


aiwilliams (commented):
Would be great to get a CHANGELOG entry alongside these changes.

Is there any chance that it matters how big the entries are in the array? One thing is sure: they [don't document any limits](https://docs.aws.amazon.com/neptune/latest/userguide/limits.html) beyond `55MB` 😆

ndowmon:
> around 1,500 elements

I don't see anything about array length, but I do see the following on AWS docs:

>There is a size limit of 55 MB on the size of an individual property or label. In RDF terms, this means that the value in any column (S, P, O or G) of an RDF quad cannot exceed 55 MB.

https://docs.aws.amazon.com/neptune/latest/userguide/limits.html

Should this PR instead be set up to check the property size?

mknoedel:
To both @ndowmon and @aiwilliams, this was a limit that @softwarewright was seeing in the persister logs. We will have to wait until he gets back from vacation to verify.

mknoedel:
> Would be great to get a CHANGELOG entry alongside these changes.

Oops, `git add .` acts weird when you working in a monorepo, my bad, adding that back in.

mknoedel:
No longer needed.

PR #556, opened 2021-10-01
Add cause.stack to IntegrationError stack trace

ndowmon:
In the azure integration, the way that we're making client API calls is causing almost every provider error, from every distinct step, to be grouped using the following trace:

```
at request (/opt/jupiterone/app/node_modules/@jupiterone/graph-azure/dist/azure/resource-manager/client.js:180:23)
at runMicrotasks (<anonymous>)
at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

This change should help to fingerprint every unique error in sentry or other similar services.

### DIFF
--- a/packages/integration-sdk-core/src/errors.ts
+++ b/packages/integration-sdk-core/src/errors.ts
@@ -84,7 +84,7 @@ export class IntegrationError extends Error {
     this.fatal = options.fatal;
     this._cause = options.cause;
     if (options.cause?.stack) {
-      this.stack = this.stack + '\n' + options.cause.stack;
+      this.stack += '\nCaused By: ' + options.cause.stack;
     }
   }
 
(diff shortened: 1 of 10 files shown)


packages/integration-sdk-core/src/errors.ts
```diff
@@ -83,6 +83,9 @@ export class IntegrationError extends Error {
     this.code = options.code;
     this.fatal = options.fatal;
     this._cause = options.cause;
+    if (options.cause?.stack) {
+      this.stack = this.stack + '\n' + options.cause.stack;
```
austinkelleher, line 87: Consider this package (created by Matt Lavin at LifeOmic): https://github.com/mdlavin/nested-error-stacks

austinkelleher (dismissed):
Approved with suggestion

packages/integration-sdk-core/src/errors.ts
ndowmon: I don't really feel that package adds a lot here, so I'm going to leave it out to keep a leaner dependency tree. However, I did lift `Caused by: ` which I liked from that package.

ceelias (approved):

PR #559, opened 2021-10-07
Change job log name from 'step_failure' to 'error_step_failure' so jobs marked as 'errorsOccurred: true'

ndowmon:
Just a proposal. This would cause the integrations to set `errorsOccurred: true` when a step fails, hence setting the job status chip to yellow rather than green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to
 
 ## Unreleased
 
+### Changed
+
+- Changed step failure job log name from `step_failure` to `error_step_failure`,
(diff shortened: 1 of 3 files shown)


rwaterman (commented):
Thanks @ndowmon. Just to clarify, this won't fail an entire integration? For example, if a scoped role was required and a step failed, this would show as yellow (not red) with this work? Thanks!

aiwilliams:
https://jupiterone.atlassian.net/browse/INT-524

austinkelleher:
Closing as this PR is stale.

PR #573, opened 2021-11-04
Allow undefined values in jobState.findEntity and jobState.hasKey

ndowmon:
### Added

- Updated the interfaces for `jobState.findEntity` and `jobState.hasKey` to allow `undefined`. Oftentimes, we use optional chaining with `jobState.findEntity` or `jobState.hasKey`, so having the ability to pass `undefined` into these methods can make our code easier to read.

  ```ts
  // without allowing `undefined`, we often need to assert values as `string`
  const virtualMachineId = await jobState.findEntity(
    nic.virtualMachine?.id as string,
  );

  // by allowing `undefined`, we can more safely use these methods without type assertions
  const virtualMachineId = await jobState.findEntity(nic.virtualMachine?.id);
  ```

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/jobState.ts
+++ b/packages/integration-sdk-runtime/src/execution/jobState.ts
@@ -252,7 +252,8 @@ export function createStepJobState({
     },
     addRelationships,
 
-    findEntity: async (_key: string) => {
+    findEntity: async (_key: string | undefined) => {
+      if (!_key) return null;
       const graphObjectMetadata = duplicateKeyTracker.getGraphObjectMetadata(
         _key,
       );
@@ -266,7 +267,8 @@ export function createStepJobState({
       );
     },
 
-    hasKey: (_key: string) => {
+    hasKey: (_key: string | undefined) => {
+      if (!_key) return false;
       return duplicateKeyTracker.hasKey(_key);
     },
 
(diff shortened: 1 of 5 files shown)


packages/integration-sdk-runtime/src/execution/jobState.ts
```diff
@@ -252,7 +252,8 @@ export function createStepJobState({
     },
     addRelationships,
 
-    findEntity: async (_key: string) => {
+    findEntity: async (_key: string | undefined) => {
+      if (!_key) return null;
```
mknoedel, line 256: Is `null` correct or should it be `undefined`?

mknoedel (approved):

aiwilliams (approved):
Love the CHANGELOG notes!

packages/integration-sdk-runtime/src/execution/jobState.ts
ndowmon: `null` is correct per the interface:
```diff
-  findEntity: (_key: string) => Promise<Entity | null>;
+  findEntity: (_key: string | undefined) => Promise<Entity | null>;
```
https://github.com/JupiterOne/sdk/pull/573/files#diff-3c56620e866ab752465a4ee3067c23ba05f05c7dab4bea3cc083b4ef2548d135R84

PR #574, opened 2021-11-05
Explicitly require a _key property when using createIntegrationEntity

ndowmon:
This is a quality-of-life improvement that was triggered by a recent failure in the `graph-azure` project, which could have been prevented if `_key` were a required property.

This is a breaking change because integrations that do not already pass a `_key` property will need to explicitly set one now.

```
### Changed

- **\*BREAKING\*** Explicitly require a `_key` property when using
  `createIntegrationEntity()`. Previously, the `createIntegrationEntity()`
  function allowed the `_key` property to be _optional_, and when not present,
  the function automatically uses either `id` or `providerId` as the entity
  `_key`.

  This caused (entirely preventable) runtime errors if the given `source` data
  did not have an `id` or `providerId` property available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ and this project adheres to
 
 ## Unreleased
 
+### Changed
+
+- **\*BREAKING\*** Explicitly require a `_key` property when using
+  `createIntegrationEntity()`. Previously, the `createIntegrationEntity()`
+  function allowed the `_key` property to be _optional_, and when not present,
+  the function automatically uses either `id` or `providerId` as the entity
+  `_key`.
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by aiwilliams — APPROVED** [review]
I'm 100% into this.

**Comment by ndowmon** [comment]
> triggered by a recent failure in the `graph-azure` project, which could have been prevented if `_key` were a required property

FWIW the `graph-azure` project actually did pass `_key`, the problem was just that we expected a `data.id` property to exist on `azure_kubernetes_cluster`, but it is undefined. We haven't actually encountered undefined `id` properties from the Azure API, even though every interface does define `id?: string`

```ts
  _key: data.id as string,
```

aiwilliams (approved):

PR #577, opened 2021-11-19
Add jobState reference to beforeAddEntity function

ndowmon:
### Added

- Added support for `jobState` in the `beforeAddEntity` hook. Also added support
  for async implementations of `beforeAddEntity`. Example usage:

  ```ts
  import { IntegrationInvocationConfig } from '@jupiterone/integration-sdk-core';
  import { integrationSteps } from './steps';
  import { createAccountVendorMappedRelationship } from './converters';
  export const invocationConfig: IntegrationInvocationConfig = {
    integrationSteps,
    beforeAddEntity: async (context, entity) => {
      if (entity._type === 'acme_account') {
        await context.jobState.addRelationship(
          createAccountVendorMappedRelationship(entity),
        );
      }
      return entity;
    },
  };
  ```


ndowmon:
A prime use case for this feature is custom mapped relationships. For example, imagine that we give our users access through the integration UI to define some custom mapping rules:

Screen Shot 2022-01-05 at 4 48 00 PM

We could then very easily convert any user-defined mapping rules into mapped relationships using the `beforeAddEntity` function. 

---

Take for example the following `userDefinedMappingRules`:

```ts

const userDefinedMappingRules = [
  {
    sourceFilter: {
      _type: "jira_account"
    },
    relationshipProperties: {
      _class: "HOSTS",
      _type: "jira_hosts_account"
    },
    relationshipDirection: "REVERSE",
    targetFilterKeys: [["_class", "name"]],
    propertyMappings: [
      {
        targetValue: "Jira",
        targetProperty: "name"
      },
      {
        targetValue: "Jira",
        targetProperty: "displayName"
      },
      {
        targetValue: "Vendor",
        targetProperty: "_class"
      },
      {
        targetValue: "jira",
        targetProperty: "_type"
      }
    ]
  },
]
```

We could very easily translate the mapping rule to a mapped relationship like so:

```ts
import { createMappedRelationship, IntegrationInvocationConfig, RelationshipDirection } from '@jupiterone/integration-sdk-core';
import { integrationSteps } from './steps';

export const invocationConfig: IntegrationInvocationConfig = {
  integrationSteps,
  beforeAddEntity: async (context, entity) => {
    /**
     * NOTE: this code is clearly specific to the mapping rule we've defined 
     * above, but it's easy to see how this could be made generic
     */
    if (entity._type === 'jira_account') {
      await context.jobState.addRelationship(
        createMappedRelationship({
          _class: 'HOSTS',
          _type: 'jira_hosts_account',
          _mapping: {
            relationshipDirection: RelationshipDirection.REVERSE,
            sourceEntityKey: entity._key,
            targetFilterKeys: [['_class', 'name']],
            targetEntity: {
              name: 'Jira',
              displayName: 'Jira',
              _class: 'Vendor',
              _type: 'jira',
            }
          }
        })
      )
    } 
    return entity;
  },
};
```

This offers a few benefits over creating global (or integration) mapping rules in J1 for a specific customer:

1. Doesn't clutter `mappings.json` with customer-specific logic.
2. Shifts much of the processing load off of the `jupiter-mapper` and into the integrations themselves.
3. Produces a mapped relationship, which is persisted to a database and therefore more easily traceable.
4. Allows users themselves to see and edit their own mapping rules.

austinkelleher:
@ndowmon Do we still need this?

ndowmon:
@austinkelleher yes please

if you're ready to review, I can rebase.

PR #579, opened 2021-11-23
Proposed fix for issue reading symlinks in Windows.

adam-in-ict:


packages/integration-sdk-runtime/src/fileSystem.ts
```diff
@@ -127,7 +127,15 @@ export async function symlink({ sourcePath, destinationPath }: SymlinkInput) {
...
   await ensurePathCanBeWrittenTo(fullDestinationPath);
-  await fs.symlink(fullSourcePath, fullDestinationPath, 'junction');
+  // On Windows, we need to perform hardlinks for files
+  if (
+    (await fs.lstat(fullSourcePath)).isFile() &&
+    process.platform === 'win32'
+  ) {
+    await fs.link(fullSourcePath, fullDestinationPath);
+  } else {
+    await fs.symlink(fullSourcePath, fullDestinationPath, 'junction');
```
ndowmon, line 137: What's the difference between `link` and `symlink`? FWIW, the `junction` arg was added to symlinks _specifically_ to handle windows file system, so if we're doing something differently for windows, we can remove the `junction` arg.

jomadu:
Is the aim here to address this issue? [JupiterOne/integration-template/issues/62](https://github.com/JupiterOne/integration-template/issues/62)

packages/integration-sdk-runtime/src/fileSystem.ts
ndowmon: Is there any reason not to use hardlinks in _any_ os? Would this cause issues in MacOS / linux?

How do hardlinks affect how these fs objects are stored? Will we be using more durable storage in our execution environment when using `link` vs `symlink`?

packages/integration-sdk-runtime/src/fileSystem.ts
```diff
@@ -127,7 +127,15 @@ export async function symlink({ sourcePath, destinationPath }: SymlinkInput) {
   const fullDestinationPath = path.resolve(directory, destinationPath);
 
   await ensurePathCanBeWrittenTo(fullDestinationPath);
-  await fs.symlink(fullSourcePath, fullDestinationPath, 'junction');
+  // On Windows, we need to perform hardlinks for files
+  if (
+    (await fs.lstat(fullSourcePath)).isFile() &&
+    process.platform === 'win32'
```
jomadu, line 133: Why is `win32` included in this condition?

packages/integration-sdk-runtime/src/fileSystem.ts
ndowmon: Hmm...

Maybe elsewhere, we can use `fs.readLink`, rather than `fs.read` to safely access the content of symbolic links on any os.

https://nodejs.org/docs/latest/api/fs.html#fspromisesreadlinkpath-options

adam-in-ict: Symlinks point to a file.  Hardlinks point to the inode value of a file.  TLDR: if you move a file to a new directory, hardlinks will follow the move and symlinks will break.  Here's an article that can do a better job than me of explaining if you're interested further:  https://medium.com/@307/hard-links-and-symbolic-links-a-comparison-7f2b56864cdd

I believe we still want the 'junction' argument in the case that we're symlinking a directory.  Otherwise, we'll potentially be back to the same issue as before where Windows will complain about insufficient admin credentials.  But to be fair, I haven't tested that to be 100% sure at this time.

I've typically only used symlinks when developing in Linux.  Symlinks are friendlier in some ways, particularly if you want the link to be from one partition to another.  Personally, my vote would be to do symlinks everywhere, but it seems like a reasonable tradeoff for Windows to avoid admin requirements.

packages/integration-sdk-runtime/src/fileSystem.ts
adam-in-ict: I included the `process.platform === 'win32'` conditional so that we could continue to use symlinks for files when developing in non-Windows platforms.

jomadu: Interesting that possible values for [process.platform](https://nodejs.org/api/process.html#processplatform) doesn't include `win64` 🤷

Apparently, [`win32` includes 64 bit platforms](https://stackoverflow.com/questions/8683895/how-do-i-determine-the-current-operating-system-with-node-js). TIL!

Thanks for answering my question 👍

adam-in-ict: :)  Yep!  I tested on a 64 bit Windows build and realized the same thing.  Doesn't feel great seeing 32 bit seemingly explicitly called out, but that's how process.platform is currently designed.

jomadu:
@austinkelleher - still needs review?

austinkelleher (approved):

packages/integration-sdk-runtime/src/fileSystem.ts
ndowmon: FWIW I think the `junction` was originally implemented in order to make writing work on windows machines... that maybe can be removed at this point.

PR #585, opened 2021-12-17
Simplify file system index iteration logic

austinkelleher:
No functional changes.

### DIFF
--- a/packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/indices.ts
+++ b/packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/indices.ts
@@ -9,21 +9,21 @@ import { readJsonFromPath, WalkDirectoryIterateeInput } from '../../fileSystem';
 import { buildIndexDirectoryPath } from './path';
 import { iterateParsedGraphFiles } from '../..';
 
-interface IterateIndexInput<GraphObject> {
+interface BaseIterateCollectionIndexParams<GraphObject> {
   type: string;
   iteratee: GraphObjectIteratee<GraphObject>;
 }
 
-interface BaseIterateIndexInput<GraphObject>
-  extends IterateIndexInput<GraphObject> {
+interface IterateCollectionIndexParams<GraphObject>
+  extends BaseIterateCollectionIndexParams<GraphObject> {
   collectionType: 'entities' | 'relationships';
 }
 
-async function iterateTypeIndex<T extends Entity | Relationship>({
+async function iterateCollectionTypeIndex<T extends Entity | Relationship>({
   type,
   collectionType,
   iteratee,
-}: BaseIterateIndexInput<T>) {
+}: IterateCollectionIndexParams<T>) {
   const path = buildIndexDirectoryPath({
     collectionType,
     type,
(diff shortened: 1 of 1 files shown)


packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/indices.ts
```diff
@@ -4,61 +4,49 @@ import {
...
   iteratee,
-}: IterateIndexInput<T>) {
+}: BaseIterateIndexInput<T>) {
   const path = buildIndexDirectoryPath({
-    collectionType: 'entities',
+    collectionType,
     type,
   });
 
-  await walkDirectory({
```
austinkelleher, line 30: The `iterateParsedGraphFiles` function was introduced in a recent PR and acts as a replacement for all of this code.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/indices.ts
```diff
@@ -75,13 +63,3 @@ export async function readGraphObjectFile<T>({
     });
   }
 }
-
-function isObjectFlushedEntityData(object: any): object is FlushedEntityData {
```
austinkelleher, line 79: `isObjectFlushedEntityData` and `isObjectFlushedRelationshipData` functions aren't needed. We fall back to an empty array when the property does not exist.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/indices.ts
adam-in-ict: 💯   Thanks!  I should've caught that when we introduced iterateParsedGraphFiles.

adam-in-ict (dismissed):

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/indices.ts
```diff
@@ -4,61 +4,49 @@ import {
...
 import { buildIndexDirectoryPath } from './path';
+import { iterateParsedGraphFiles } from '../..';
 
 interface IterateIndexInput<GraphObject> {
   type: string;
   iteratee: GraphObjectIteratee<GraphObject>;
 }
 
-export async function iterateEntityTypeIndex<T extends Entity = Entity>({
+interface BaseIterateIndexInput<GraphObject>
```
aiwilliams, line 17: 1. What do you think of `IterateCollectionIndexInput`? I suppose that `Base` to me that in a type hierarchy, this would be the type with the minimal number of properties and other types would build on this, but here we see that it is extending something with fewer properties.
2. What do you think about, in cases where function parameters are received as an object for destructuring, we use the name `IterateCollectionIndexParams` instead of `Input`? I prefer `Params` since it reflects the fact that we're creating a type to represent parameters to a function.

aiwilliams (dismissed):

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/indices.ts
ndowmon: sweeet

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/indices.ts
```diff
@@ -4,61 +4,49 @@ import {
...
-          await iteratee(entity);
-        }
-      }
-    },
-  });
+  await iterateParsedGraphFiles(async (data) => {
+    for (const graphObj of (data[collectionType] as T[]) || []) {
+      await iteratee(graphObj);
+    }
+  }, path);
```
ndowmon, line 36: FWIW I find it a bit more readable when the callback is the last argument, rather than the first, but not a big deal.

ndowmon (approved):

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/indices.ts
austinkelleher: 1. Thanks for the input. I meant to flip the naming the other way. 😬 
2. Definitely! Will change.

Thanks for the great suggestions.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/indices.ts
austinkelleher: Agreed! Good feedback. This code was here previously, so I didn't change the signature.

aiwilliams (approved):

PR #606, opened 2022-02-07
INT-2469 Add cache when using --step option

VDubber:
The --use-dependencies-cache was replaced by "always on" cache when the `--step` flag is being used. If desired, it can be turned off with `--no-cache`. A file path to a separate cache can be supplied with `--cache-path`.

> ## Updated
> 
> Added --use-dependencies-cache flag to the `collect` command to  increase development speed on long-running steps.
> 
> 
> A customer is looking for a way to "skip" a step(s) that is depended upon by the target step. This implementation uses existing data from the .j1-integration directory to populate the the .j1-cache if no filepath is specified.
> 
> Looking for final feedback.


VDubber:
Running these changes within graph-signal-sciences produces the following output:

image

image

ndowmon (commented):
Sam, this is an awesome effort! I did uncover some really tricky behavior that I think we should definitely discuss - getting this one done is definitely more complex than I had originally thought. Let's set up some time to chat!

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -220,6 +223,45 @@ export async function executeWithContext<
       createStepGraphObjectDataUploader,
     } = options;
 
+    // Load data for ignored step dependencies.
+    if (options.useStorageForIgnoredStepDependencies) {
```
ndowmon, line 227: Oops, forgot to add this comment to the review - it would be really nice to abstract this behavior behind a function.

```ts
if (options.useStorageForIgnoredStepDependencies) {
  await populateJobStateFromCache({ ... });
}
```

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -220,6 +223,45 @@ export async function executeWithContext<
...
+              await graphObjectStore.addEntities(stepId, entities);
+              stepStartStates[stepId].cacheLoaded = true;
+            }
+          },
+        });
+      }
+
+      if (loadedEntities.length > 0) {
+        logger.info(
+          { loadedEntities },
```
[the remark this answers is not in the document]
VDubber: Great thought. Conveying that something was done was the purpose so a count would be sufficient, I think.

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -220,6 +223,45 @@ export async function executeWithContext<
       createStepGraphObjectDataUploader,
     } = options;
 
+    // Load data for ignored step dependencies.
+    if (options.useStorageForIgnoredStepDependencies) {
+      const loadedEntities: string[] = [];
+      for (const stepId of config?.dependentSteps ?? []) {
+        const path = getRootStorageAbsolutePath(`graph/${stepId}/entities`);
```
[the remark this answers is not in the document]
VDubber: Probably a limit on my experience with large integrations. If a step (A) depends on a step (B) with relationships, does A ever query or rely on the relationships?
For the sake of speed I didn't ingest relationship from disk. 
tl;dr - the reason of speed

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -220,6 +223,45 @@ export async function executeWithContext<
...
+
+      if (loadedEntities.length > 0) {
+        logger.info(
+          { loadedEntities },
+          `Successfully loaded entities for ignored step dependencies from disk.`,
+        );
+      }
+    }
+
+    await removeStorageDirectory();
```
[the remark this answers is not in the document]
VDubber: No, this is because the data is exported to disk after each run. So a user flow would be: 1) `j1-integration collect -step fetch-users` and then 2) `j1-integration collection --step fetch-users -I` and 2) can be run as many times as desired.

-----

Thank you for thinking this through. I wasn't aware of how flushing worked and I think you are right.
Here are two thoughts:
1. we first copy the pertinent data to a .j1-step-data directory and then read the data in from there.
2. OR we allow the user to specify an arg via the CLI to a directory of data they want imported.

Or we implement both. If the directory arg is specified, we use it. If not, we copy the pertinent data.

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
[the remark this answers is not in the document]
VDubber: image

This is what the results say. We could introduce a new status: cache_miss?

docs/integrations/development.md
```diff
@@ -1051,7 +1051,7 @@ For convenience when developing locally, we will also look for a
 Initially, the CLI will support a limited interface consisting of only three
 commands: `collect`, `sync`, and `run`.
 
-#### `j1-integration collect`
+#### Command `j1-integration collect`
```
VDubber, line 1054: Added Command and Option to the front of these because they were hard to see

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -242,14 +251,74 @@ export function executeStepDependencyGraph<
...
         }
       });
     }
 
+    async function loadCacheForStep(step: Step<TStepExecutionContext>) {
+      let status = StepResultStatus.FAILURE;
+      const stepCacheFilePath = stepStartStates[step.id].stepCachePath;
+      const entitiesPath = `${stepCacheFilePath}/entities`;
+      const relationshipsPath = `${stepCacheFilePath}/relationships`;
+
```
ndowmon, line 274: I think we want to use the `jobState` interface to add entities and relationships rather than the `graphObjectsStore` to ensure some functionality is preserved (this includes use of the typeTracker, duplicateKeyTracker, and `beforeAddEntity` function, if it exists). 

```suggestion
      if (createStepGraphObjectDataUploader) {
        uploader = createStepGraphObjectDataUploader(step.id);
      }

      const { logger, jobState } = buildStepContext<
        TExecutionContext,
        TStepExecutionContext
      >({
        context: executionContext,
        duplicateKeyTracker,
        typeTracker,
        graphObjectStore,
        dataStore,
        stepId: step.id,
        beforeAddEntity,
        uploader,
      });
```

Then below: 

```diff
-           await graphObjectStore.addEntities(step.id, entities);
+           await jobState.addEntities(entities);
```

```diff
-           await graphObjectStore.addRelationships(step.id, relationships);
+           await jobState.addRelationships(relationships);
```

And finally after all have been ingested:

```diff
+ await jobState.flush();
```

ndowmon: You also get access to the step-specific logger 🥳

ndowmon: I'm a little bit wondering whether it's worth just injecting the `loadCacheForStep` into the `executeStep` function (rather than where it's at inside of `enqueueLeafSteps()`, so that we don't have to create the `context` in two places, and we can reuse all of the existing log messages etc. defined in `executeStep`. Probably not worth changing, but a thought exercise.

```diff
      try {
-       await step.executionHandler(context);
+       if (hasCachePath(stepId)) {
+         await loadCacheForStep(stepId, context);
+       } else {
+         await step.executionHandler(context);
+       }
        context.logger.stepSuccess(step);
```

ndowmon (dismissed):
So much better! Great work and thanks for the screenshots / tests. 

I've approved with some comments, would love to see what you have to say about them but overall I'm good with this.

packages/integration-sdk-cli/src/commands/collect.ts
```diff
@@ -60,10 +82,43 @@ export function collect() {
...
+  );
+}
+
+/**
+ * When no filepath is specified, the .j1-integration directory
+ * is copied to .j1-cache
+ */
+async function copyToCache() {
+  const graphDirectory = path.join(getRootStorageDirectory(), 'graph');
+  if (fs.ensureDir(graphDirectory)) {
```
austinkelleher, line 116: This seems like it is an async function? See: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/fs-extra/index.d.ts#L45

Can we install `@types/fs-extra` as a dev dep?

docs/integrations/development.md
```diff
@@ -1143,14 +1143,45 @@ For convenience, steps can allow be provided as a comma delimited list.
...
+ex: `j1-integration collect --step fetch-users --use-dependencies-cache` - Builds & uses cache from .j1-integration  
+ex: `j1-integration collect --step fetch-users --use-dependencies-cache ./` - Uses .j1-cache found in the root of the project  
+ex: `j1-integration collect --step fetch-users --use-dependencies-cache ./path-to-cache` - Uses .j1-cache found in the path specified 
+
+A common use pattern: 
+1. Execute collection command _without_ the `--use-dependencies-cache` option to gather data in .j1-integration
+2. Execute collection command with `--step` and `--use-dependencies-cache` option _without_ specifying a filepath.
+ This will cause the .j1-integration data to populate the .j1-cache.
+3. Execute collection command with `--step` and `--use-dependencies-cache` option specifying a 
+filepath to the previously created .j1-cache (most commonly `./`).
```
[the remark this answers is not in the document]
ceelias: I have the same question as Nick here. If this is true, I suppose running step 2 is essentially "refreshing the cache". I was going to suggest we add an option to clear and refresh the cache but sounds like maybe this would handle the refreshing part anyway. I'm just not sure how intuitive that is vs a separate flag needed to refresh the cache in here and otherwise it will default to using the cache that already lives in the directory. Thoughts?

packages/integration-sdk-cli/src/__tests__/cli.test.ts
```diff
@@ -48,6 +48,70 @@ describe('collect', () => {
...
+        },
+      ],
+      metadata: {
+        partialDatasets: {
+          types: [],
+        },
+      },
+    });
+  });
+
```
ceelias, line 114: I may have missed it but what happens in the case that the integration wasn't run yet and there is no data to copy over to the cache? Might be worth a test here.

packages/integration-sdk-cli/src/commands/collect.ts
```diff
@@ -30,15 +33,34 @@ export function collect() {
       collector,
       [],
     )
+    .option(
+      '-C, --use-dependencies-cache [filePath]',
+      'Loads cache for the dependencies required by the step(s) specified in --step option. Execution of these steps is skipped. Data found in .j1-integration is used if no filepath is provided.',
```
ceelias, line 38: More so for curiousity but we have the ability to create multiple dependency graphs. Have you tried to run that use case/does your code account for that? Maybe something we can address in a later PR if not.

ceelias (commented):
Really solid work! Thanks for working through this! I left a fix on a typo and a few thought provoking comments

packages/integration-sdk-cli/src/commands/collect.ts
```diff
@@ -60,10 +82,43 @@ export function collect() {
...
+
+/**
+ * When no filepath is specified, the .j1-integration directory
+ * is copied to .j1-cache
+ */
+async function copyToCache() {
+  const graphDirectory = path.join(getRootStorageDirectory(), 'graph');
+  if (fs.ensureDir(graphDirectory)) {
+    await fs.emptyDir(getRootCacheDirectory());
+    await fs.copy(graphDirectory, getRootCacheDirectory()).catch((error) => {
```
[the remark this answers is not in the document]
VDubber: Good thought especially since the .j1-integration dir is about to be deleted.

packages/integration-sdk-cli/src/commands/collect.ts
```diff
@@ -30,15 +33,34 @@ export function collect() {
...
       // Point `fileSystem.ts` functions to expected location relative to
       // integration project path.
       process.env.JUPITERONE_INTEGRATION_STORAGE_DIRECTORY = path.resolve(
         options.projectPath,
         '.j1-integration',
       );
 
+      if (
+        typeof options.useDependenciesCache === 'string' ||
+        options.useDependenciesCache instanceof String
```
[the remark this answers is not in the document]
VDubber: String or Boolean is correct. As long as the commander package always uses one or the other, then yes. This is a bit more defensive code. But not pretty. This type of check would actually be really nice to have in a util...

packages/integration-sdk-cli/src/commands/collect.ts
VDubber: I have not. I'd love to see that in action!

docs/integrations/development.md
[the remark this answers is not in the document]
VDubber: @ndowmon Yes. if you repeat step 2 it will copy over the .j1-integration data over and over and only execute the fetch-users step. If the .j1-integration data goes sideways, you'll end up with a cache that is busted. Hence the suggestion to specify the filepath - then it'll stop trying to create the cache.

@ceelias It isn't clear. I thought about a command as well but that seemed too much. I imagine caches being self-maintaining. Not sure if that's the correct approach. Right now the cache creation destroys the previous cache. We could do more of an update, maybe?.... That might be tricky to get right.

packages/integration-sdk-cli/src/__tests__/cli.test.ts
VDubber: If a step is tagged as using the cache but fails to load any entities/relationships, it marks that step as a failure. I'll see if I can get that in here.

VDubber: I added a test and it caused the other tests to fail. I may come back to this at another time.

packages/integration-sdk-cli/src/commands/collect.ts
ndowmon: Opened #619 to track

packages/integration-sdk-cli/src/commands/collect.ts
```diff
@@ -60,10 +82,43 @@ export function collect() {
         },
         {
           enableSchemaValidation,
+          useDependenciesCache: options.useDependenciesCache,
```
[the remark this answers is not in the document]
VDubber: Good call. Adjusting.

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
VDubber: Great suggestion here. Really make the logging beautiful instead of 2nd class.

docs/integrations/development.md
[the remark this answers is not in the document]
VDubber: @ndowmon @ceelias 
Crazy idea: what if we turn on caching by default whenever the `--step` flag is used? We could introduce a simple `--no-cache` flag and `-cache-path` flag. I think this would eliminate the confusion around how caching works currently.

The --step flag is already used to decrease developer wait time... Why not make it ever greater with the cache?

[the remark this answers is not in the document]
ndowmon: 🤔 hm, that is an interesting idea... although I do think you may be stuck in cacheland 😉 

My concern with defaulting to `--no-cache` is that differences in upstream dependencies _may_ (or probably _should_) cause the chosen `--step` to behave differently. I still think it's best to refresh the dependent steps before running the new step.

aiwilliams (commented):
> Crazy idea: what if we turn on caching by default whenever the --step flag is used? We could introduce a simple --no-cache flag and -cache-path flag. I think this make the cache helpful and out of the way until you need to deal with it.
>
> The --step flag is already used to decrease developer wait time... Why not make it even better with the cache?
>
> On the initial run, when there is no .j1-integration data, it doesn't utilize the cache...

I think this is an excellent idea and should be where we go in the next iteration of this excellent feature.

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -340,6 +355,67 @@ export function executeStepDependencyGraph<
...
+      context: TStepExecutionContext,
+    ) {
+      let status = StepResultStatus.FAILURE;
+      const entitiesPath = `${stepCacheFilePath}/entities`;
+      const relationshipsPath = `${stepCacheFilePath}/relationships`;
+
+      const { jobState, logger } = context;
+
+      let entitiesCount = 0;
+      await walkDirectory({
```
austinkelleher, line 374: We should be able to replace this with `iterateParsedGraphFiles` which is exported from `packages/integration-sdk-runtime/src/fileSystem.ts`.

Cleanup suggestion...

There is a function called `iterateParsedGraphFiles`, which is exported from `packages/integration-sdk-runtime/src/fileSystem.ts`. I think we could expose two new functions:

- `iteratedParsedEntityGraphFiles`
- `iteratedParsedRelationshipGraphFiles`

Example impl:

```ts
export function iteratedParsedEntityGraphFiles(
  iteratee: (entities: Entity[]) => Promise<void>,
  graphPath?: string
) {
  return iterateParsedGraphFiles(async (data) => {
    if (data.entities) await iteratee(data.entities);
  }, graphPath);
}
```

Example usage:

```ts
await iteratedParsedEntityGraphFiles(
  async (entities) => {
    entitiesCount += entities.length;
    await jobState.addEntities(entities);
    status = StepResultStatus.CACHED;
  },
  entitiesPath
);
```

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -340,6 +355,67 @@ export function executeStepDependencyGraph<
...
+            relationshipCount += relationships.length;
+            await jobState.addRelationships(relationships);
+            status = StepResultStatus.CACHED;
+          }
+        },
+      });
+
+      if (entitiesCount || relationshipCount) {
+        logger.info(
+          `Loaded ${entitiesCount} entities and  ${relationshipCount} relationship(s) from cache.`,
```
austinkelleher, line 408: Minor:

It is good practices to place all variables in the logger function's first arg instead of using string interpolation. String interpolated logs are slow to search.

Example of a fast log search term:

```
Loaded entities and relationship(s) from cache.
```

Example of a slow search term for which we will need to use the log search "contains" operator to find all instances of our log message. E.g. `~= "relationship(s) from cache."`

```
Loaded 10 entities and 10 relationship(s) from cache.
```

Suggestion:

```ts
logger.info({
  entitiesCount,
  relationshipCount
}, 'Loaded entities and relationship(s) from cache.');
```

packages/integration-sdk-runtime/src/execution/step.ts
```diff
@@ -163,12 +169,27 @@ export function prepareLocalStepCollection<
           ctx,
         ) ?? {});
         const enabledRecord: StepStartStates = {};
+
+        if (dependenciesCache?.enabled) {
+          ctx.logger.info(
+            `Using dependencies cache found at ${dependenciesCache.filepath}`,
```
austinkelleher, line 175: Same feedback about logging.

```suggestion
            { cacheFilePath: dependenciesCache.filePath }, 'Using dependencies cache',
```

packages/integration-sdk-runtime/src/execution/step.ts
```diff
@@ -163,12 +169,27 @@ export function prepareLocalStepCollection<
...
+          );
+        }
+
         for (const stepId of allStepIds) {
           const originalValue = originalEnabledRecord[stepId] ?? {};
           if (stepsToRun.includes(stepId)) {
             enabledRecord[stepId] = {
               ...originalValue,
               disabled: false,
+              ...(dependenciesCache?.enabled &&
```
austinkelleher, line 185: Would you mind extracting this out into a function? I think it will be a bit easier for readers to understand.

packages/integration-sdk-runtime/src/fileSystem.ts
```diff
@@ -17,12 +17,12 @@ import { readGraphObjectFile } from './storage/FileSystemGraphObjectStore/indice
 const brotliCompress = promisify(zlib.brotliCompress);
 const brotliDecompress = promisify(zlib.brotliDecompress);
 
-export const DEFAULT_CACHE_DIRECTORY_NAME = '.j1-integration';
+export const DEFAULT_STORAGE_DIRECTORY_NAME = '.j1-integration';
```
austinkelleher, line 20: 👍

docs/integrations/development.md
```diff
@@ -1143,14 +1144,54 @@ For convenience, steps can allow be provided as a comma delimited list.
...
+results from disk. The intent of this is to increase development speed for new
+integrations. When no filepath is specified, an attempt to create a cache is
+made by copying the contents of `./.j1-integrations/graph` directory to
+`./.j1-cache`. The structure of the cache follows a similar format as the
+.j1-integration data storage, as described [here](#data-collection).
+
+###### An example of the expected cache structure
+
+```
+.j1-cache/
```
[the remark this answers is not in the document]
austinkelleher: Agreed. @VDubber and I talked about this too.

austinkelleher (commented):
Nice work @VDubber! I left some comments. I know that you still have a few changes that you're working on, but I supplied my initial review.

docs/integrations/development.md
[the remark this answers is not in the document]
VDubber: Great minds with the same idea. I'll make it happen.

austinkelleher (approved):

PR #607, opened 2022-02-09
Add ability to disable checking class schema

edreyer1:


ndowmon (dismissed):

packages/integration-sdk-testing/src/jest.ts
```diff
@@ -319,19 +321,21 @@ export function toMatchGraphObjectSchema<T extends Entity>(
...
-          graphObjectClassToSchemaRef(classInst),
-        ),
-      );
-    } catch (err) {
-      return {
-        message: () => `Error loading schemas for class (err=${err.message})`,
-        pass: false,
-      };
+  if (!disableClassMatch) {
+    for (const classInst of _class) {
```
ndowmon, line 325: Ohhh I think I just noticed another thing that has been biting us recently... there's this annoying bit where we can't match if the `_class` is not an array... it's cause above, we are forcing the `_class` property to be an array.

Maybe you can also make this change:

```diff
- _class = Array.isArray(_class) ? _class : [_class];
+ classSchemasToMatch = Array.isArray(_class) ? _class : [_class];
```
and then update this:

```suggestion
    for (const classInst of classSchemasToMatch) {
```

ndowmon:
Can you update the `CHANGELOG.md` before merge?

ndowmon (approved):

PR #611, opened 2022-02-10
Add optional `schema` property to graph object metadata

ndowmon:
### Added

- Added optional `schema` property to `StepGraphObjectMetadata`. This allows
  developers to provide the property schema to expect on entities,
  relationships, and mapped relationships. This serves two uses:
  1. Schemas can be used at runtime or test-time to verify that an entity has
     the correct properties
  2. The `j1-integration document` command could automatically produce consumer
     documentation about the properties that an entity / relationship is
     expected to have

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,14 @@ and this project adheres to
 
 ### Added
 
+- Added optional `schema` property to `StepGraphObjectMetadata`. This allows
+  developers to provide the property schema to expect on entities,
+  relationships, and mapped relationships. This serves two uses:
+  1. Schemas can be used at runtime or test-time to verify that an entity has
+     the correct properties
+  2. The `j1-integration document` command could automatically produce consumer
+     documentation about the properties that an entity / relationship is
+     expected to have
 - Added `executeStepWithDependencies` utility to
   `@jupiterone/integration-sdk-testing` package. This allows developers to test
   specific integration steps in isolation, while assuring that all of its
--- a/packages/integration-sdk-core/src/types/step.ts
+++ b/packages/integration-sdk-core/src/types/step.ts
@@ -95,7 +95,14 @@ export interface StepGraphObjectMetadata {
   _type: string;
 
   /**
-   * The schema used to describe the properties assigned to this entity
+   * An optional extension to the JSON schemas defined by the `_class` of this entity
+   * in the @jupiterone/data-model.
+   *
+   * Extensions defined by this schema are in addition to the schemas referenced by the
+   * `_class` values of the entity. These should be considered unique properties for
+   * this integration. Ask yourself, should a property here be considered for promotion
+   * to the @jupiterone/data-model to benefit normalization of all integrations producing
+   * this class of entity?
    */
   schema?: GraphObjectSchema;
 
--- a/packages/integration-sdk-core/src/types/step.ts
+++ b/packages/integration-sdk-core/src/types/step.ts
@@ -1,4 +1,7 @@
-import { RelationshipClass } from '@jupiterone/data-model';
+import {
+  RelationshipClass,
+  IntegrationEntitySchema,
+} from '@jupiterone/data-model';
 
 import {
   ExecutionContext,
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,14 @@ and this project adheres to
 
 ### Added
 
+- Added optional `schema` property to `StepGraphObjectMetadata`. This allows
+  developers to provide the property schema to expect on entities,
+  relationships, and mapped relationships. This serves two uses:
+  1. Schemas can be used at runtime or test-time to verify that an entity has
+     the correct properties
+  2. The `j1-integration document` command could automatically produce consumer
```
austinkelleher, line 19: Would love to see a ticket in the SDK describing this!

austinkelleher (commented):
Awesome! Would you mind adding this to the development documentation please? I'd also love to see a task created to update the `integration-template`!

packages/integration-sdk-core/src/types/step.ts
```diff
@@ -81,9 +84,21 @@ export interface GraphObjectIndexMetadata {
   enabled: boolean;
 }
 
+export interface GraphObjectSchema extends IntegrationEntitySchema {
+  $schema?: string;
+  $id?: string;
+  description?: string;
+  additionalProperties?: boolean;
```
ndowmon, line 91: Do we want to make some of these `required`? It could prevent us from bugs like what we saw in `graph-aws`

ndowmon: e.g. `properties`, `required`, `additionalProperties`

ndowmon:
```diff
export interface GraphObjectSchema extends IntegrationEntitySchema {
  $schema?: string;
  $id?: string;
  description?: string;
- additionalProperties?: boolean;
+ additionalProperties: boolean;
+ properties: Required<IntegrationEntitySchema>['properties'];
+ required: Required<IntegrationEntitySchema>['required'];
}
```

aiwilliams: Considering the form of https://github.com/JupiterOne/data-model/blob/178d8838d6004a45ff97b229fd8a0dafd3b50e6f/src/index.ts#L6:

```ts
export type IntegrationEntitySchema = {
  $ref?: string;
  allOf?: IntegrationEntitySchema[];
  properties?: {
    [propertyName: string]: any;
  };
  required?: string[];
};
```

Nothing is required there, which makes sense, allowing for various combinations of those properties.

For this `GraphObjectSchema`, I think we could make `description` required. I'm confused about what `$schema` does here in the context of the metadata for an entity. Will `$id` ever be provided, perhaps we could remove that? We could probably use some documentation on these properties.

What do you think about naming this `EntitySchemaExtension`?

ndowmon: TBH I think it's bad that we're not using the published NPM types for JSON Schema. 

`yarn add --dev @types/json-schema`

```ts
import { JSONSchema7 } from 'json-schema'
```

Perhaps this type should _actually_ be something like the following?

```ts
type GraphObjectSchema = Required<Pick<JSONSchema7, 'additionalProperties' | 'required' | 'properties'>>;
```

ndowmon: Personally I don't see why we would want to document a `description` if we're just using this to describe step entity metadata.

ndowmon: Also I don't love `EntitySchemaExtension` because I would really like to see this used on relationships as well, in cases where relationships will have properties on them (example, permissions relationships or firewall rule relationships)

aiwilliams: Ohh, nice find on the `@types/json-schema`! I wonder if I totally overlooked that or if it's relatively new. I think it would be wise to move to use those types.

aiwilliams (approved):

packages/integration-sdk-core/src/types/step.ts
ndowmon: Tracking:
- https://github.com/JupiterOne/sdk/issues/620
- https://github.com/JupiterOne/data-model/issues/147

PR #639, opened 2022-02-22
Fix #638 - Support Neo4j entity ingestion with `false` property values

### LINKED ISSUE #638: Neo4j ingestion ignores properties with the value `false`
## Description

The Neo4j ingestion tool is ignoring properties with the value `false`.

## Test case

Example entity:

```json
{
  "_class": "User",
  "mfaEnabled": false,
  ...
}
```

The following Cypher query has 0 results:

```cy
MATCH (user:User {
  mfaEnabled: false
}) RETURN user
```

austinkelleher:
Fixes #638. The Neo4j ingestion tool was ignoring properties with the value `false`.

With this fix, the following query is supported:

```cy
MATCH (user:User {
  mfaEnabled: false
}) RETURN user
```

### DIFF
--- a/packages/integration-sdk-cli/src/neo4j/__tests__/neo4jUtilities.test.ts
+++ b/packages/integration-sdk-cli/src/neo4j/__tests__/neo4jUtilities.test.ts
@@ -60,7 +60,7 @@ describe('#buildPropertyParameters', () => {
     expect(
       buildPropertyParameters({
         test: '123',
-        ignore: undefined,
+        ignore: null,
       }),
     ).toEqual({
       test: '123',
--- a/packages/integration-sdk-cli/src/neo4j/__tests__/neo4jUtilities.test.ts
+++ b/packages/integration-sdk-cli/src/neo4j/__tests__/neo4jUtilities.test.ts
@@ -30,14 +30,52 @@ describe('#neo4jUtilities', () => {
       '1a!b@c#d$e%f^g&h*i(j)k-l=m+n\\o|p\'q\\"r;s:t/u?v.w,x>y<z`1~2\t3\n4[5]6{7}8 90',
     );
   });
+});
+
+describe('#buildPropertyParameters', () => {
   test('should build property string correctly including sanitization', () => {
-    const testPropResults: Object = buildPropertyParameters({
-      test: '123',
-      '1sanitize1hi&$abc d': '1h"i&$abc d',
-    });
-    expect(testPropResults).toEqual({
+    expect(
+      buildPropertyParameters({
+        test: '123',
+        '1sanitize1hi&$abc d': '1h"i&$abc d',
+      }),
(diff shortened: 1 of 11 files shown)


packages/integration-sdk-cli/src/neo4j/neo4jUtilities.ts
```diff
@@ -21,28 +21,29 @@ export function sanitizeValue(value: string): string {
...
+      // stringify JSON in rawData so we can store it.
+      propertyParameters[key] = `"${JSON.stringify(propVal)}"`;
     } else {
       // Sanitize out characters that aren't allowed in property names
       const propertyName = sanitizePropertyName(key);
 
-      //If we're dealing with a number or boolean, leave alone, otherwise
-      //wrap in single quotes to convert to a string and escape all
-      //other single quotes so they don't terminate strings prematurely.
-      if (propList[key]) {
```
austinkelleher, line 35: Previously, this was causing us to ignore `false`

packages/integration-sdk-cli/src/neo4j/neo4jUtilities.ts
```diff
@@ -21,28 +21,29 @@ export function sanitizeValue(value: string): string {
...
-          typeof propList[key] == 'number' ||
-          typeof propList[key] == 'boolean'
-        ) {
-          propertyParameters[propertyName] = propList[key];
-        } else {
-          propertyParameters[propertyName] = sanitizeValue(
-            propList[key].toString(),
-          );
-        }
+      if (typeof propVal === 'undefined' || propVal === null) {
```
austinkelleher, line 35: Explicitly check `undefined` and `null`

packages/integration-sdk-cli/src/neo4j/__tests__/neo4jUtilities.test.ts
```diff
@@ -30,14 +30,52 @@ describe('#neo4jUtilities', () => {
...
       test: '123',
-      n1sanitize1hi__abc_d: '1h\\"i&$abc d',
+    });
+  });
+
+  test('should ignore properties with the value "null"', () => {
+    expect(
+      buildPropertyParameters({
+        test: '123',
+        ignore: undefined,
```
ndowmon, line 63:
```suggestion
        ignore: null,
```

adam-in-ict (approved):

packages/integration-sdk-cli/src/neo4j/neo4jUtilities.ts
ndowmon: Just curious - why `typeof propVal === 'undefined'` rather than just `propVal === 'undefined'`?

ndowmon (approved):

ndowmon (approved):

PR #640, opened 2022-02-23
Add export to EntityCoreProperties

VDubber:
#why so that it could be successfully resolved in packages dependent on the sdk.


austinkelleher (approved):
💯

PR #645, opened 2022-03-02
Added `IngestionLimitEncountered` log event type

mdaum:


mdaum:
https://jupiterone.atlassian.net/browse/INT-2731

austinkelleher (approved):

austinkelleher:
@mdaum We'll need to update the SDK documentation describe when/why someone would leverage this logger event

ceelias (dismissed):
I agree with @austinkelleher that it would be nice to have some doc on when/how this is expected to be used

packages/integration-sdk-runtime/src/logger/__tests__/index.test.ts
```diff
@@ -669,17 +669,21 @@ describe('#publishEvent', () => {
 
     logger.on('event', onEmitEvent);
 
-    logger.publishWarnEvent({
-      name: IntegrationWarnEventName.MissingPermission,
-      description: 'the description',
-    });
+    let numEventsTypes = 0;
+    for (const key in IntegrationWarnEventName) {
```
ndowmon, line 673: Seems like jest's `test.each` method might be useful here!

packages/integration-sdk-core/src/types/logger.ts
```diff
@@ -44,6 +44,7 @@ export interface PublishInfoEventInput extends PublishEventInput {
 
 export enum IntegrationWarnEventName {
   MissingPermission = 'warn_missing_permission',
+  IngestionLimitEncountered = 'warn_ingestion_limit_encountered',
```
ndowmon, line 47: Docstrings can help developers understand the purpose of this event at development time. I've suggested a description, but I think you could come up with a better description than mine!

```suggestion
  /**
   * Some J1 integrations are configured to only ingest up to a set number 
   * of entities in a given integration job. For example, the Jira integration 
   * is configured to ingest only 2,000 issues per integration job.
   */
  IngestionLimitEncountered = 'warn_ingestion_limit_encountered',
```

ndowmon (approved):
Couple of suggestions, but looks good to me!

packages/integration-sdk-core/src/types/logger.ts
mdaum: sounds good! I will add some. I am probably going to omit the second sentence though. If we change the # issues per project ingested on Jira side, I feel like this is liable to get lost in the weeds and deviate from reality.

mdaum:
> @mdaum We'll need to update the SDK documentation describe when/why someone would leverage this logger event

https://github.com/JupiterOne/sdk/blob/main/docs/integrations/development.md#logger
Looks like this would be the spot for it. In it we say
```
Most information logged via the `logger` will _not_ be displayed to customers
via the integration job event log, but there are some special messages that need
to be displayed to customers to allow them to know if there are issues
preventing integrations from collecting data.

For these cases, we will provide specialized functions on the `logger` to assist
with displaying those kinds of messages.
```
and make no mention of `publishEvent` and it's Info, Warn, and Error variants. I will do this as a part of this PR.

ndowmon (approved):
Looks great. Thanks for the update to development.md!

austinkelleher (approved):
Thanks @mdaum!

Issue #646, opened 2022-03-03
`assignTags` does not support empty string values


austinkelleher:
Currently the `assignTags` function does not support empty string values. This is causing some integrations to not automatically assign legitimate tags onto entities.

Example input:

```ts
assignTags(entity, [{ key: 'something', value: '' }]);
```

Example output:

```ts
{
  ...,
  something: ''
}
```

PR #647, opened 2022-03-03
Add support for empty string values to assignTags

eXtremeX:
References #646

Probably a few different ways of handling this one, this approach uses JS coercion to let only string values that aren't `undefined` nor `null` pass through (our empty string lives there). What do you think of this? @austinkelleher


eXtremeX:
Great, will do!

PR #651, opened 2022-03-11
Adding properties to mapped entities for Neo4j uploads.

adam-in-ict:
Fixing issue causing mapped entities to have missing property data when uploaded to Neo4j.

### DIFF
--- a/packages/integration-sdk-cli/src/neo4j/neo4jGraphStore.ts
+++ b/packages/integration-sdk-cli/src/neo4j/neo4jGraphStore.ts
@@ -109,9 +109,11 @@ export class Neo4jGraphStore {
       if (relationship._mapping) {
         //Mapped Relationship
         if (relationship._mapping['skipTargetCreation'] === false) {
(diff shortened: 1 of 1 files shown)


packages/integration-sdk-cli/src/neo4j/neo4jGraphStore.ts
```diff
@@ -109,17 +109,19 @@ export class Neo4jGraphStore {
       if (relationship._mapping) {
         //Mapped Relationship
         if (relationship._mapping['skipTargetCreation'] === false) {
+          const targetEntity = relationship._mapping['targetEntity'];
           //Create target entity first
           const tempEntity: Entity = {
-            _class: relationship._mapping['targetEntity']._class,
+            ...targetEntity,
+            _class: targetEntity._class,
```
adam-in-ict, line 116: We don't need to worry about sanitizing _class or _type here, since they will be sanitized in the later addEntities() call.

austinkelleher (approved):

PR #653, opened 2022-03-17
Adding SDK_API_CLIENT_MAX_TIMEOUT environment variable

zemberdotnet:
# Description

This PR adds a new environment variable, `SDK_API_CLIENT_MAX_TIMEOUT`, to configure the `maxTimeout` for the SDK's `ApiClient`.

## Summary

The new `SDK_API_CLIENT_MAX_TIMEOUT` is an environment variable used to control the `retry.maxTimeout` property on the `Alpha` client the SDK uses to make calls to our public API routes. If no value is given for `SDK_API_CLIENT_MAX_TIMEOUT`, a default value of 10000 ms will be used. This is the same value the `Alpha` uses internally by default. `Alpha` also has other internal default values for `retry`. I have confirmed that if these other properties are `undefined` on the `retry` object we pass, `Alpha` will use its own defaults.

If the `SDK_API_CLIENT_MAX_TIMEOUT` is set but is not a number, then a new `IntegrationMaxTimeoutBadValue` error will be thrown. I would love feedback on this choice. I also thought about falling back to the default and logging a warning but felt being more explicit would be helpful. I would love to know what others think.

### DIFF
--- a/packages/integration-sdk-runtime/src/api/__tests__/index.test.ts
+++ b/packages/integration-sdk-runtime/src/api/__tests__/index.test.ts
@@ -7,7 +7,6 @@ import {
   createApiClient,
   getAccountFromEnvironment,
 } from '../index';
-import { IntegrationMaxTimeoutBadValueError } from '../error';
 
 jest.mock('@lifeomic/alpha');
 
@@ -91,6 +90,9 @@ describe('createApiClient', () => {
       apiBaseUrl,
       account: 'test-account',
       accessToken: 'test-key',
+      retryOptions: {
+        maxTimeout: 20000,
+      },
     });
 
     expect(client).toBeInstanceOf(AlphaMock);
@@ -107,22 +109,5 @@ describe('createApiClient', () => {
         maxTimeout: 20000,
       },
     });
-  });
-
-  test('bad SDK_API_CLIENT_MAX_TIMEOUT throws error', () => {
-    process.env.SDK_API_CLIENT_MAX_TIMEOUT = '10000x';
-    const apiBaseUrl = getApiBaseUrl();
-    try {
-      createApiClient({
-        apiBaseUrl,
-        account: 'test-account',
-        accessToken: 'test-key',
-      });
-    } catch (err) {
-      expect(err).toBeInstanceOf(IntegrationMaxTimeoutBadValueError);
-      expect(err.message).toBe(
-        'The SDK_API_CLIENT_MAX_TIMEOUT environment variable is not a number',
-      );
-    }
   });
 });
--- a/packages/integration-sdk-runtime/src/api/__tests__/index.test.ts
+++ b/packages/integration-sdk-runtime/src/api/__tests__/index.test.ts
@@ -77,10 +77,6 @@ describe('getApiKeyFromEnvironment', () => {
 });
 
 describe('createApiClient', () => {
-  afterEach(() => {
-    delete process.env.SDK_API_CLIENT_MAX_TIMEOUT;
-  });
-
   test('successfully creates apiClient', () => {
     process.env.SDK_API_CLIENT_MAX_TIMEOUT = '20000';
 
--- a/packages/integration-sdk-runtime/src/api/__tests__/index.test.ts
+++ b/packages/integration-sdk-runtime/src/api/__tests__/index.test.ts
@@ -7,6 +7,7 @@ import {
   createApiClient,
   getAccountFromEnvironment,
 } from '../index';
+import { IntegrationMaxTimeoutBadValueError } from '../error';
 
 jest.mock('@lifeomic/alpha');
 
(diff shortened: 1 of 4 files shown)


packages/integration-sdk-runtime/src/api/index.ts
```diff
@@ -41,9 +42,19 @@ export function createApiClient({
     headers.Authorization = `Bearer ${accessToken}`;
   }
 
+  const maxTimeout = process.env.SDK_API_CLIENT_MAX_TIMEOUT
```
VDubber, line 45: nit: Consider pulling out into a separate method. Maybe `determineMaxTimeout`?

VDubber:
Hey Matthew!
Code changes look great! I would like to understand the context a bit more. What caused the need for this change? Who will be the main user of this EnvVar? Why a EnvVar instead of a parameter to `createApiClient`?

Also, how can we successfully document that this EnvVar is available to the next developer?

packages/integration-sdk-runtime/src/api/index.ts
austinkelleher: I think we should avoid environment variables when we can. Environment variables are global state and require some clever testing strategies.

Instead of using an environment variable, I think `createApiClient` should accept a `retryConfig` argument. `createApiClient` is called from the managed integration SDK. I think we could allow passing configuration along to the managed SDK from the integration itself and have the managed SDK pass along the configuration to `createApiClient`.

```ts
createApiClient({
  ...,
  retryConfig: {
    maxTimeout: 10000
  }
});
```

If `retryConfig` is not supplied, then the value will be `undefined` and `Alpha` would just use the defaults. Once we are happy with the `maxTimeout` that we assign from the managed integration side, we should make a follow-up change to the SDK to use that value as the default.

Thoughts? @VDubber @zemberdotnet

VDubber: Agreed (see [here](https://github.com/JupiterOne/sdk/pull/653#issuecomment-1070986891)). Also, EnvVars don't provide an obvious interface.

zemberdotnet: In the latest PR the `createApiClient` function now looks like:
```ts
export function createApiClient({
  apiBaseUrl,
  account,
  accessToken,
  retryOptions,
}: CreateApiClientInput): ApiClient {
```
The `retryOptions` is optional. I went with `retryOptions` as that is what it's called in `Alpha`. 

Alternatively, we could pull in the full `AlphaOptions` from [Alpha](https://github.com/lifeomic/alpha/blob/9adf37ed32a030990e564d41cd09607fc7e6c9aa/src/Alpha.d.ts#L12). That would make `createApiClient` look more like:
```ts
export function createApiClient({
  account,
  accessToken,
  alphaOptions,
}): CreateApiClientInput): ApiClient {
```

`AlphaOptions` extends `AxiosOptions` so `apiBaseUrl` would be included on the `AlphaOptions` object. This is probably a bigger change than necessary and might break something - I'm not sure. I wanted to put it out there incase anyone had thoughts, but I'm happy with the existing change as well.

packages/integration-sdk-runtime/src/api/__tests__/index.test.ts
```diff
@@ -76,24 +76,38 @@ describe('getApiKeyFromEnvironment', () => {
...
-test('createApiClient', () => {
-  const apiBaseUrl = getApiBaseUrl();
-
-  const client = createApiClient({
-    apiBaseUrl,
-    account: 'test-account',
-    accessToken: 'test-key',
+describe('createApiClient', () => {
+  afterEach(() => {
+    delete process.env.SDK_API_CLIENT_MAX_TIMEOUT;
```
mdaum, line 81: I don't think you need this afterEach block anymore

austinkelleher (approved):
@zemberdotnet Nice work! After we are happy with this change, can you please update the development documentation to reflect the `createApiClient` API improvements?

zemberdotnet:
@VDubber 
Thanks for the review! These are all great questions:
> I would like to understand the context a bit more. What caused the need for this change?


There are two sets of retry logic going on. One part is visible in the SDK [here](https://github.com/JupiterOne/sdk/blob/1504c547765bd8636a8adcaf19098bc1b9f802e1/packages/integration-sdk-runtime/src/synchronization/index.ts#L267), and the other part is internal retry logic used by `Alpha` [here](https://github.com/lifeomic/alpha/blob/master/src/adapters/response-retry.js). The retry logic used by `Alpha` has a default `maxTimeout` of 10000 ms. When an upload is taking a particularly long time in the persister, this can cause `Alpha` to send another request. The duplicate request ends up creating duplicate keys which causes the whole job to close and many errors to follow.

> Who will be the main user of this EnvVar? Why a EnvVar instead of a parameter to createApiClient?

As I write this, it's been taken out as suggested by Austin and you. 😄 Having it in the configuration will make the code clearer and easier to test!

> Also, how can we successfully document that this EnvVar is available to the next developer?

While it has been taken out, it might be good to see if there are any other undocumented environment variables that could be worth documenting in a `Setup/Installation` section of a `README.md`

packages/integration-sdk-runtime/src/api/__tests__/index.test.ts
```diff
@@ -76,24 +76,34 @@ describe('getApiKeyFromEnvironment', () => {
...
-  expect(AlphaMock).toHaveBeenCalledWith({
-    baseURL: apiBaseUrl,
-    headers: {
-      Authorization: 'Bearer test-key',
-      'Content-Type': 'application/json',
-      'LifeOmic-Account': 'test-account',
-    },
+describe('createApiClient', () => {
+  test('successfully creates apiClient', () => {
+    process.env.SDK_API_CLIENT_MAX_TIMEOUT = '20000';
```
austinkelleher, line 81: Looks like this isn't used anymore

austinkelleher (approved):

mdaum (approved):

PR #657, opened 2022-03-17
Fixing an error where undefined retryOptions would not adopt defaults

zemberdotnet:
# Description

This PR fixes an error where if `retryOptions` is `undefined`, there would be no default retry behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to
 
 ## Unreleased
 
+### Fixed
+
+- Fixed an error where if `retryOptions` was `undefined` default options would
+  not be adopted
+
 ## [8.6.2] - 2022-03-17
 
 ### Added
--- a/packages/integration-sdk-runtime/src/api/index.ts
+++ b/packages/integration-sdk-runtime/src/api/index.ts
@@ -52,7 +52,7 @@ export function createApiClient({
   return new Alpha({
     baseURL: apiBaseUrl,
     headers,
-    retry: retryOptions,
+    retry: retryOptions ?? {},
   }) as ApiClient;
 }
 


packages/integration-sdk-runtime/src/api/index.ts
```diff
@@ -52,7 +52,7 @@ export function createApiClient({
   return new Alpha({
     baseURL: apiBaseUrl,
     headers,
-    retry: retryOptions,
+    retry: retryOptions ?? {},
```
VDubber, line 55: Is there an applicable test that would've caught this and we should consider updating now?

zemberdotnet: It would have been nice to catch this in a test. This is where `Alpha` sets defaults: [relevant code](https://github.com/lifeomic/alpha/blob/9adf37ed32a030990e564d41cd09607fc7e6c9aa/src/adapters/response-retry.js#L56). The client uses a response interceptor on the response and if there is an error it will retry. What actually is assigned to `retry` isn't accessible outside this scope. It might be indirectly testable by making a test where the client needs to retry?

VDubber: That makes sense. It's probably not worth the effort at this point is what I'm kinda coming to see now.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to
 
 ## Unreleased
```
VDubber, line 11: 8.6.3?

VDubber (approved):
Good catch!

CHANGELOG.md
zemberdotnet: Was going to add it in with the next `lerna version`. It's not a big thing, but if someone else merged changes then that might affect the version number.

VDubber: Nah, my bad!

PR #659, opened 2022-03-21
Issue 654 - Raw data reduction

adam-in-ict:
Attempting to shrink rawData when we receive a 413 error that the payload size is too large.

### DIFF
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -355,7 +355,7 @@ export function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
 
     // Find largest Entity
     for (const entity in data) {
-      const length = JSON.stringify(data[entity]).length
+      const length = JSON.stringify(data[entity]).length;
       if (length > largestEntitySize) {
         largestEntitySize = length;
         largestEntityKey = entity;
@@ -364,10 +364,10 @@ export function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
 
     // Find largest _rawData entry (typically 0, but check to be certain)
     for (const rawEntry in data[largestEntityKey]['_rawData']) {
-      const length = JSON.stringify(data[largestEntityKey]['_rawData'][rawEntry]).length;
-      if (
-        length > largestRawDataEntrySize
-      ) {
+      const length = JSON.stringify(
+        data[largestEntityKey]['_rawData'][rawEntry],
+      ).length;
+      if (length > largestRawDataEntrySize) {
         largestRawDataEntrySize = length;
         largestRawDataEntryKey = rawEntry;
       }
@@ -377,10 +377,15 @@ export function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
     for (const item in data[largestEntityKey]['_rawData'][
       largestRawDataEntryKey
     ]['rawData']) {
-      const length =
-        data[largestEntityKey]['_rawData'][largestRawDataEntryKey]['rawData'][item] ?
-        JSON.stringify(data[largestEntityKey]['_rawData'][largestRawDataEntryKey]['rawData'][item]).length :
-        0;
+      const length = data[largestEntityKey]['_rawData'][largestRawDataEntryKey][
+        'rawData'
+      ][item]
+        ? JSON.stringify(
+            data[largestEntityKey]['_rawData'][largestRawDataEntryKey][
+              'rawData'
+            ][item],
+          ).length
+        : 0;
       if (length > largestItemSize) {
         largestItemKey = item;
         largestItemSize = length;
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -335,7 +335,8 @@ function handleUploadDataChunkError({
 
   if (isRequestUploadTooLargeError(err)) {
     logger.info(`Attempting to shrink rawData`);
-    shrinkRawData(batch, logger);
+    const shrinkResults = shrinkRawData(batch);
+    logger.info(shrinkResults, 'Shrink raw data result');
   } else if (systemErrorResponseData?.code === 'JOB_NOT_AWAITING_UPLOADS') {
     throw new IntegrationError({
       code: 'INTEGRATION_UPLOAD_AFTER_JOB_ENDED',
@@ -442,6 +443,14 @@ interface keyAndSize {
   size: number;
 }
 
+// Interface for shrink run results
+interface shrinkRawDataResults {
+  initialSize: number;
+  totalSize: number;
+  itemsRemoved: number;
+  totalTime: number;
+}
+
 /**
  * Helper function to find the largest entry in an array and return its key
  * and approximate byte size.  We JSON.stringify as a method to try and have
@@ -471,11 +480,10 @@ function findLargestItemKeyAndByteSize(data: any[]): keyAndSize {
  *
  * @param data
  */
-export function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
-  data: T[K][],
-  logger,
+export function shrinkRawData(
+  data: UploadDataLookup[keyof UploadDataLookup][],
   maxSize = UPLOAD_SIZE_MAX,
-) {
+): shrinkRawDataResults {
   const startTimeInMilliseconds = Date.now();
   let totalSize = Buffer.byteLength(JSON.stringify(data));
   const initialSize = totalSize;
@@ -509,15 +517,12 @@ export function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
   }
 
   const endTimeInMilliseconds = Date.now();
-  logger.info(
-    {
-      initialSize,
-      totalSize,
-      itemsRemoved,
-      totalTime: endTimeInMilliseconds - startTimeInMilliseconds,
-    },
-    'Shrink raw data result',
-  );
+  return {
+    initialSize,
+    totalSize,
+    itemsRemoved,
+    totalTime: endTimeInMilliseconds - startTimeInMilliseconds,
+  };
 }
 
 interface AbortSynchronizationInput extends SynchronizationJobContext {
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -404,7 +404,7 @@ function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
       }
     }
 
-    console.log(
+    logger.info(
       `SHRINKING RAW DATA BY TRUNCATING ${largestItemKey} in rawData`,
     );
     data[largestEntryKey]['_rawData'][largestRawDataEntryKey]['rawData'][
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -429,6 +429,7 @@ export function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
   let totalSize = Buffer.byteLength(JSON.stringify(data));
   const initialSize = totalSize;
   let itemsRemoved = 0;
+  const sizeOfTruncated = Buffer.byteLength("'TRUNCATED'");
 
   while (totalSize > maxSize) {
     let largestEntityKey = '';
@@ -481,7 +482,7 @@ export function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
       largestItemKey
     ] = 'TRUNCATED';
     itemsRemoved += 1;
-    totalSize = Buffer.byteLength(JSON.stringify(data));
+    totalSize = totalSize - largestItemSize + sizeOfTruncated;
   }
   const endTimeInMilliseconds = Date.now();
 
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -480,7 +480,10 @@ function findLargestItemKeyAndByteSize(data: any[]): keyAndSize {
  *
  * @param data
  */
-export function shrinkRawData(data: UploadDataLookup[keyof UploadDataLookup][], maxSize = UPLOAD_SIZE_MAX,): shrinkRawDataResults {
+export function shrinkRawData(
+  data: UploadDataLookup[keyof UploadDataLookup][],
+  maxSize = UPLOAD_SIZE_MAX,
+): shrinkRawDataResults {
   const startTimeInMilliseconds = Date.now();
   let totalSize = Buffer.byteLength(JSON.stringify(data));
   const initialSize = totalSize;
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -272,14 +272,10 @@ interface UploadDataChunkParams<T extends UploadDataLookup, K extends keyof T> {
 }
 
 function isRequestUploadTooLargeError(err): boolean {
-  if (
+  return (
     err.code === 'RequestEntityTooLargeException' ||
     err.response?.status === 413
-  ) {
-    return true;
-  } else {
-    return false;
-  }
+  );
 }
 
 type SystemErrorResponseData = {
@@ -514,9 +510,13 @@ export function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
 
   const endTimeInMilliseconds = Date.now();
   logger.info(
-    `shrinkRawData: raw data reduced from ${initialSize} to ${totalSize} truncating ${itemsRemoved} rawData items in ${
-      endTimeInMilliseconds - startTimeInMilliseconds
-    } ms`,
+    {
+      initialSize,
+      totalSize,
+      itemsRemoved,
+      totalTime: endTimeInMilliseconds - startTimeInMilliseconds,
+    },
+    'Shrink raw data result',
   );
 }
 
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -404,9 +404,6 @@ function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
       }
     }
 
-    logger.info(
-      `SHRINKING RAW DATA BY TRUNCATING ${largestItemKey} in rawData`,
-    );
     data[largestEntryKey]['_rawData'][largestRawDataEntryKey]['rawData'][
       largestItemKey
     ] = 'TRUNCATED';
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -355,8 +355,9 @@ export function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
 
     // Find largest Entity
     for (const entity in data) {
-      if (JSON.stringify(data[entity]).length > largestEntitySize) {
-        largestEntitySize = JSON.stringify(data[entity]).length;
+      const length = JSON.stringify(data[entity]).length
+      if (length > largestEntitySize) {
+        largestEntitySize = length;
         largestEntityKey = entity;
       }
     }
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -480,10 +480,7 @@ function findLargestItemKeyAndByteSize(data: any[]): keyAndSize {
  *
  * @param data
  */
-export function shrinkRawData(
-  data: UploadDataLookup[keyof UploadDataLookup][],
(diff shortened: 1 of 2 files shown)


ndowmon:
Would you mind adding some tests?

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -305,22 +314,103 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+        },
+        { concurrency: UPLOAD_CONCURRENCY },
+      );
+    } catch (err) {
+      if (
+        err.code === 'RequestEntityTooLargeException' ||
+        err.response?.status === 413
+      ) {
+        logger.info(`Attempting to shrink raw_data`);
+        shrinkRawData(data);
```
ceelias, line 344: I wonder how much more efficient this would be if we only pass the actual batch that caused the failure rather than all off the data.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -275,11 +278,16 @@ export async function uploadDataChunk<
       delay: 200,
       factor: 1.05,
       handleError(err, attemptContext) {
-        if (err.code === 'RequestEntityTooLargeException') {
-          // No reason to retry these errors as the request size ain't gonna change.
-          throw err;
```
VDubber, line 280: Is `throw err` any different than using `attemptContext.abort()`

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -323,6 +331,69 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+    for (const item in data[largestEntryKey]['_rawData'][
+      largestRawDataEntryKey
+    ]['rawData']) {
+      if (
+        data[largestEntryKey]['_rawData'][largestRawDataEntryKey]['rawData'][
+          item
+        ] &&
+        data[largestEntryKey]['_rawData'][largestRawDataEntryKey]['rawData'][
+          item
+        ].length > largestItemSize
```
VDubber, line 378: Is the `item` guaranteed to have the `.length` property? 
What if the property is actually an object, is the intent to find the largest nested property or just cut out the entire object? 
We may want to add a few test cases with different data types.
Also, isn't `default` the most common root property within rawData? I could be totally wrong on this one.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -323,6 +331,69 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+        largestRawDataEntrySize
+      ) {
+        largestRawDataEntrySize = JSON.stringify(
+          data[largestEntryKey]['_rawData'][rawEntry],
+        ).length;
+        largestRawDataEntryKey = rawEntry;
+      }
+    }
+
+    // Find largest item within rawData
```
VDubber, line 368: It'd be interesting to know the benefit of find the largest N items. If there happens to be 2-3 items that are very similar and thus similar in size, it'd be better to remove those items all at the same time. Maybe worth considering now or in the future. Hard to know how to optimize without more data examples.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -275,11 +278,16 @@ export async function uploadDataChunk<
       delay: 200,
       factor: 1.05,
       handleError(err, attemptContext) {
-        if (err.code === 'RequestEntityTooLargeException') {
-          // No reason to retry these errors as the request size ain't gonna change.
-          throw err;
-        }
+        // Did err.code ever work?  I see it as undefined if I console.log its
```
VDubber, line 281: Were you able to get a solid handle on this?

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -323,6 +331,69 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+    let largestRawDataEntryKey = '';
+    let largestRawDataEntrySize = 0;
+    let largestItemKey = '';
+    let largestItemSize = 0;
+
+    // Find largest Entity
+    for (const entry in data) {
+      if (JSON.stringify(data[entry]).length > largestEntrySize) {
+        largestEntrySize = JSON.stringify(data[entry]).length;
+        largestEntryKey = entry;
```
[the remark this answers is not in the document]
adam-in-ict: I thought about and even started to perform the sorting, but decided to bank on the assumption that most of the time we'd only be removing one very large value and not typically need to go through the loop again.  In the case that we only need to perform one removal, simply finding the largest is quicker than sorting the entire thing.  I can certainly switch to the method of sorting that you'd mentioned though.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -29,6 +29,9 @@ export { createEventPublishingQueue } from './events';
 const UPLOAD_BATCH_SIZE = 250;
 const UPLOAD_CONCURRENCY = 6;
 
+// Let's max at 5 meg even though 6 will technically work
+const UPLOAD_SIZE_MAX = 5000000;
```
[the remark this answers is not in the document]
adam-in-ict: Fair point.  I don't have any solid reason for that much wiggle room.  I think all we'd really need space for is header information.  I guess my main thought was that even a 5 MB upload size was pretty generous.  But I take your point about the cost of finding the extra 1 MB of data to remove.

packages/integration-sdk-runtime/src/synchronization/index.ts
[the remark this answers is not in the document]
zemberdotnet: I don't want to derail this in a different direction if it requires too much work / adds to much complexity, given @VDubber comment here and below, I was wondering why we don't try to split the batch into more pieces?

I'm not sure if data integrity is high priority for `rawData`, but if we scan the data with `JSON.stringyify(data[entry]).length` and find that every object is less than 6MB each, we could retain all the data by creating several batches of size < `MAX_SIZE`.

Otherwise we could send what will fit, and then apply a truncating strategy to the remaining objects >6MB. I'm not sure if this was discussed or if retaining all `_rawData` is a high priority, so would love to know your thoughts.

packages/integration-sdk-runtime/src/synchronization/index.ts
adam-in-ict: Good catch!  I'd actually intended to do a JSON.stringify() call on that item so that it would be the closest thing to an apples-to-apples comparison for finding the largest overall thing no matter what the data type.  I'll get that updated and you can have another look and see if you're still concerned with it.

Regarding `default` being the most common root property within rawData, yes I believe you're correct that `default` is the most common value for the `name` property, i.e.

"_rawData": [
        {
          "name": "default",
          "rawData": {
            "admin_id": "...",
            "admin_units": [],
            ...
          }
        }
      ]

packages/integration-sdk-runtime/src/synchronization/index.ts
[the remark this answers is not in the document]
adam-in-ict: I updated the max size to 6 MB - 16 KB.  I believe that most servers will reject a header larger than 16 KB, so that should be more than plenty of space for it.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -323,6 +331,69 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
   );
 }
 
+export function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
```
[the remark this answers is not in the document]
adam-in-ict: Added in latest commit.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -323,6 +331,69 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+  while (totalSize > UPLOAD_SIZE_MAX) {
+    let largestEntryKey = '';
+    let largestEntrySize = 0;
+    let largestRawDataEntryKey = '';
+    let largestRawDataEntrySize = 0;
+    let largestItemKey = '';
+    let largestItemSize = 0;
+
+    // Find largest Entity
+    for (const entry in data) {
```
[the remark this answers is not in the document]
adam-in-ict: Updated in latest commit

packages/integration-sdk-runtime/src/synchronization/index.ts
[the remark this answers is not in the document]
adam-in-ict: I like the idea of making it robust enough to just work with any sized entity, but I also don't know that we'd be happy in the long run if we let our integrations potentially start uploading 10+ MB entities.  That could get fairly painful from a database/storage/cost perspective.  Truncating at 6 MB feels like we've still given entities plenty of storage.

I think I'd be more inclined to go the route of reducing batch sizes to succeed with all sizes if we were running into entities of 6+ MB in non-raw data, but even then that feels like a case where the integration likely needs to break that out into multiple entities to really gain the most value out of storing it in a graph database.

packages/integration-sdk-runtime/src/synchronization/index.ts
adam-in-ict: Got some solid feedback from @austinkelleher on this one.  The difference is due to my testing being running `yarn j1-integration sync` locally versus this running in an ECS container in AWS and making direct calls.  I added in a helper function that checks both so that it will work in both instances but also be a bit more readable.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -399,6 +414,84 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+        largestItemKey = item;
+        largestItemSize = length;
+      }
+    }
+
+    data[largestEntityKey]['_rawData'][largestRawDataEntryKey]['rawData'][
+      largestItemKey
+    ] = 'TRUNCATED';
+    itemsRemoved += 1;
+    totalSize = Buffer.byteLength(JSON.stringify(data));
```
zemberdotnet, line 484: Would the following be a constant equivalent:
```ts
    totalSize = totalSize - largestItemSize + 9
```
where 9 is the length of `TRUNCATED`

adam-in-ict: Agreed with the overall idea for sure. Saving that many JSON.stringify calls for the entire data object should be some worthwhile savings.  Pushed an update to this in the latest commit.

zemberdotnet: I think in the case of plain ASCII text the `.length` and the `.byteLength` will be the same.
```ts 
console.log("TRUNCATED".length === Buffer.byteLength("TRUNCATED")) // true
```
However UTF-8 characters like `🦄` will not have `.length` and `.byteLength` of the same size.
```ts
console.log("🦄".length === Buffer.byteLength("🦄")) // 2 === 4 -- false
console.log("ø".legnth === Buffer.byteLength("ø")) // 1 === 2 -- false
```
Currently, `largestItemSize` comes from the `.length` property of the string, but the `byteLength` of the request is what matters.

I think with the 16KB cushion built-in it may be safe 99% of the time, but occasionally, this could still produce a request larger than the 6MB limit if it contains a large number of UTF-8 characters. I'm not sure if `byteLength` is a constant time operation, but if it is then switching 
```ts
const length = data[largestEntityKey]['_rawData'][largestRawDataEntryKey][
        'rawData'
      ][item]
        ? JSON.stringify(
            data[largestEntityKey]['_rawData'][largestRawDataEntryKey][
              'rawData'
            ][item],
          ).length
        : 0;
```
to
```ts
const length = data[largestEntityKey]['_rawData'][largestRawDataEntryKey][
        'rawData'
      ][item]
        ? Buffer.byteLength(JSON.stringify(
            data[largestEntityKey]['_rawData'][largestRawDataEntryKey][
              'rawData'
            ][item],
          ))
        : 0;


```

adam-in-ict: Thanks and good call out!  Agreed that it's probably safe, but I'm going to update it based on your recommendation all the same.  Overall still executing faster than before.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -426,6 +440,87 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+  while (totalSize > maxSize) {
+    let largestEntityKey = '';
+    let largestEntitySize = 0;
+    let largestRawDataEntryKey = '';
+    let largestRawDataEntrySize = 0;
+    let largestItemKey = '';
+    let largestItemSize = 0;
+
+    // Find largest Entity
+    for (const entity in data) {
```
austinkelleher, line 469: This is interesting. Is there a reason why you chose to use `in` instead of `of`? 🤔 

I think we should switch this to `of` and just reference the `entity` directly instead of `data.entity`.

austinkelleher: Fwiw, it might be nice if this was another function too. `largestEntitySize` is _only_ used inside of this block. Would be nice to see something like this:

```ts
const largestEntity = getLargestEntityFromBatch(data);
```

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -426,6 +440,87 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+      const length = JSON.stringify(data[entity]).length;
+      if (length > largestEntitySize) {
+        largestEntitySize = length;
+        largestEntityKey = entity;
+      }
+    }
+
+    // Find largest _rawData entry (typically 0, but check to be certain)
+    for (const rawEntry in data[largestEntityKey]['_rawData']) {
```
austinkelleher, line 478:
```suggestion
    for (const rawDataEntryKey in largestEntity._rawData) {
```

austinkelleher: This would also make for a nice function.

```ts
const largestRawDataEntryKey = getLargestRawDataEntryKeyFromEntity(largestEntity);
```

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -426,6 +440,87 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+        'rawData'
+      ][item]
+        ? Buffer.byteLength(
+            JSON.stringify(
+              data[largestEntityKey]['_rawData'][largestRawDataEntryKey][
+                'rawData'
+              ][item],
+            ),
+          )
+        : 0;
```
austinkelleher, line 502: There's a lot going on here too and I'm having a bit of difficultly parsing all of it in my head 😅 

How about we split this into two functions:

```ts
const entityRawDataEntryByteLength = getByteLengthForEntityRawDataEntry(...)
```

```ts
const largestEntityItemSize = getLargestEntityItemSize(...)
```

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -426,6 +440,87 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
 }
 
+/**
+ * Removes data from the rawData of the largest entity until the overall size
+ * of the data object is less than maxSize (defaulted to UPLOAD_SIZE_MAX).
+ *
+ * @param data
+ */
+export function shrinkRawData<T extends UploadDataLookup, K extends keyof T>(
+  data: T[K][],
```
austinkelleher, line 450: What are your thoughts on making this _not_ generic, so that we can get better typings for `Entity`?

austinkelleher (commented):
Thanks for all of the work! Left some comments that I think will help make this easier to read.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -266,6 +271,17 @@ interface UploadDataChunkParams<T extends UploadDataLookup, K extends keyof T> {
   batch: T[K][];
 }
 
+function isRequestUploadTooLargeError(err): boolean {
+  if (
+    err.code === 'RequestEntityTooLargeException' ||
+    err.response?.status === 413
```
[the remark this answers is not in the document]
adam-in-ict: Updated in latest commit

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -426,6 +440,87 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+      largestItemKey
+    ] = 'TRUNCATED';
+    itemsRemoved += 1;
+    totalSize = totalSize - largestItemSize + sizeOfTruncated;
+  }
+  const endTimeInMilliseconds = Date.now();
+
+  logger.info(
+    `shrinkRawData: raw data reduced from ${initialSize} to ${totalSize} truncating ${itemsRemoved} rawData items in ${
+      endTimeInMilliseconds - startTimeInMilliseconds
```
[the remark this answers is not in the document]
adam-in-ict: Updated to pull out of the utility function and remove string interpolation in latest commit.

packages/integration-sdk-runtime/src/synchronization/index.ts
adam-in-ict: Sounds good to me.  I think I just defaulted to that to match the calling function.  In the latest commit it now expects `UploadDataLookup` arrays to allow for either Entities or Relationships (even though I know we're only likely to ever need to shrink Entities)

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -426,6 +437,94 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
   );
 }
 
+// Interface for storing both the key value and total size of a given array entry
+interface keyAndSize {
```
austinkelleher, line 441: I haven't seen camel case used for interface naming. Have you seen this? I've only seen pascal case used for `interface` names and camel case used for `interface` members.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -426,6 +437,94 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+
+/**
+ * Helper function to find the largest entry in an array and return its key
+ * and approximate byte size.  We JSON.stringify as a method to try and have
+ * an apples to apples comparison no matter what the data type of the value is.
+ *
+ * @param data
+ * @returns
+ */
+function findLargestItemKeyAndByteSize(data: any[]): keyAndSize {
```
austinkelleher, line 462: Thanks for pulling into a separate function.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -426,6 +437,94 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
...
+    let largestEntity: keyAndSize = { key: '', size: 0 };
+    let largestRawDataEntry: keyAndSize = { key: '', size: 0 };
+    let largestItem: keyAndSize = { key: '', size: 0 };
+
+    // Find largest Entity
+    largestEntity = findLargestItemKeyAndByteSize(data);
+
+    // Find largest _rawData entry (typically 0, but check to be certain)
+    largestRawDataEntry = findLargestItemKeyAndByteSize(
+      data[largestEntity.key]['_rawData'],
```
austinkelleher, line 503: Do we need to access this property in this way anymore? Are we able to true TypeScript typings to just use:

```ts
data[largestEntity.key]._rawData
```

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -319,14 +333,10 @@ function handleUploadDataChunkError({
...
 
-  if (err.code === 'RequestEntityTooLargeException') {
-    // No reason to retry these errors as the request size ain't gonna change.
-    throw new IntegrationError({
-      code: 'INTEGRATION_UPLOAD_FAILED',
-      cause: err,
-      fatal: false,
-      message: 'Failed to upload integration data because payload is too large',
-    });
+  if (isRequestUploadTooLargeError(err)) {
```
austinkelleher, line 336: If the upload batch is a batch of relationships, we should bubble `isRequestUploadTooLargeError`.

packages/integration-sdk-runtime/src/synchronization/index.ts
austinkelleher: We should try to avoid using `any` if possible.

adam-in-ict: Updated in latest commit.

packages/integration-sdk-runtime/src/synchronization/index.ts
adam-in-ict: I added a check that if we don't have any Entities and/or we don't have any _rawData entries in those Entities that we can truncate it throws.

packages/integration-sdk-runtime/src/synchronization/index.ts
adam-in-ict: Thanks for the catch!  Updated in latest commit.

austinkelleher (approved):

PR #661, opened 2022-03-22
Fix #660 - Trigger fatal error when system responds with `JOB_NOT_AWAITING_UPLOADS`

### LINKED ISSUE #660: Trigger fatal error when JupiterOne system responds with `JOB_NOT_AWAITING_UPLOADS`
When the JupiterOne system responds with `JOB_NOT_AWAITING_UPLOADS`, we should _not_ retry the request and we should trigger a `fatal` error. The JupiterOne system has force closed the job.

austinkelleher:
- Log error `code` when an integration logs an error
- Throw an `IntegrationError` with code `INTEGRATION_UPLOAD_FAILED` when the
  JupiterOne system responds with `RequestEntityTooLargeException`
- [#660] - Trigger fatal error when JupiterOne system responds with
  `JOB_NOT_AWAITING_UPLOADS`

### DIFF
--- a/packages/integration-sdk-runtime/src/logger/index.ts
+++ b/packages/integration-sdk-runtime/src/logger/index.ts
@@ -335,11 +335,17 @@ export class IntegrationLogger
     description: string;
   }) {
     const { eventName, errorId, err, description } = options;
+
+    // If there is a `code` property on the `Error`, we should include this
+    // in our log. This is helpful for when we receive an HTTP response error
+    // and the service includes specific codes (e.g. the JupiterOne system).
+    const code = (err as any).code;
+
     if (shouldReportErrorToOperator(err)) {
-      this.error({ errorId, err }, description);
+      this.error({ errorId, err, code }, description);
       this.onFailure({ err });
     } else {
-      this.warn({ errorId, err }, description);
+      this.warn({ errorId, err, code }, description);
     }
 
     this.publishEvent({
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -261,6 +261,35 @@ interface UploadDataChunkParams<T extends UploadDataLookup, K extends keyof T> {
   batch: T[K][];
 }
 
+type SystemErrorResponseData = {
+  /**
+   * The specific system-level error code (e.g. `ENTITY_IS_NOT_ARRAY`)
+   */
+  code: string;
+  /**
+   * The specific system-level error message
+   * (e.g. `"\"entities\" should be an array"`)
+   */
+  message: string;
+};
+
+/**
+ * The JupiterOne system will encapsulate error details in the response in
+ * some situations. For example:
+ *
+ * {
+ *   "error": {
+ *     "code": "ENTITY_IS_NOT_ARRAY",
+ *     "message": "\"entities\" should be an array"
+ *   }
+ * }
+ */
+function getSystemErrorResponseData(
+  err: any,
+): SystemErrorResponseData | undefined {
+  return err.response?.data?.error;
+}
+
 type HandleUploadDataChunkErrorParams = {
   err: any;
   attemptContext: AttemptContext;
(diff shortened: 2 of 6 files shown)


packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -335,11 +335,17 @@ export class IntegrationLogger
...
     const { eventName, errorId, err, description } = options;
+
+    // If there is a `code` property on the `Error`, we should include this
+    // in our log. This is helpful for when we receive an HTTP response error
+    // and the service includes specific codes (e.g. the JupiterOne system).
+    const code = (err as any).code;
+
     if (shouldReportErrorToOperator(err)) {
-      this.error({ errorId, err }, description);
+      this.error({ errorId, err, code }, description);
```
austinkelleher, line 345: The original error `code` is not showing up in the logs.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -260,6 +261,53 @@ interface UploadDataChunkParams<T extends UploadDataLookup, K extends keyof T> {
   batch: T[K][];
 }
 
+type HandleUploadDataChunkErrorParams = {
+  err: any;
+  attemptContext: AttemptContext;
+  logger: IntegrationLogger;
+};
+
+function handleUploadDataChunkError({
```
austinkelleher, line 270: Extracted this function as it has become fairly large.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -260,6 +261,53 @@ interface UploadDataChunkParams<T extends UploadDataLookup, K extends keyof T> {
...
+  err,
+  attemptContext,
+  logger,
+}: HandleUploadDataChunkErrorParams): void {
+  if (err.code === 'RequestEntityTooLargeException') {
+    // No reason to retry these errors as the request size ain't gonna change.
+    throw new IntegrationError({
+      code: 'INTEGRATION_UPLOAD_FAILED',
+      cause: err,
+      fatal: false,
```
austinkelleher, line 280: This is non-fatal because other steps should still run if one step fails to upload

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -260,6 +261,53 @@ interface UploadDataChunkParams<T extends UploadDataLookup, K extends keyof T> {
...
+      message: 'Failed to upload integration data because payload is too large',
+    });
+  } else if (err.code === 'JOB_NOT_AWAITING_UPLOADS') {
+    throw new IntegrationError({
+      code: 'INTEGRATION_UPLOAD_AFTER_JOB_ENDED',
+      cause: err,
+      fatal: true,
```
austinkelleher, line 287: This integration cannot proceed in this case. We should mark it as `fatal`.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -260,6 +261,53 @@ interface UploadDataChunkParams<T extends UploadDataLookup, K extends keyof T> {
...
+    attemptContext.attemptsRemaining &&
+    // There are sometimes intermittent credentials errors when running
+    // a managed integration on AWS Fargate. They consistently succeed
+    // with retry logic, so we don't want to log a warn.
+    err.code !== 'CredentialsError'
+  ) {
+    logger.warn(
+      {
+        err,
+        code: err.code,
```
adam-in-ict, line 303: Do you actually ever see a value for `err.code`?  I'm working in the same area for the work on shrinking rawData for `RequestEntityTooLargeException` error codes, but I never saw it as anything other than undefined.  I was looking at adding in checking for `err.response?.status` for the code, just because that's what I was used to using for Axios responses.  Would welcome any knowledge share on how this `err.code` is currently functioning.

austinkelleher: Good comms. This should be fixed in the latest commit.

adam-in-ict (approved):

ndowmon (commented):
Great change!

PR #663, opened 2022-03-24
Support executeStepWithDependencies for steps with no dependencyGraphId

ndowmon:
### Changed

- Support `executeStepWithDependencies` for steps with no `dependencyGraphId`, even if `invocationConfig.dependencyGraphOrder` is present

### DIFF
--- a/packages/integration-sdk-testing/src/executeStepWithDependencies.ts
+++ b/packages/integration-sdk-testing/src/executeStepWithDependencies.ts
@@ -9,15 +9,16 @@ import {
 export async function executeStepWithDependencies(params: StepTestConfig) {
   const { stepId, invocationConfig, instanceConfig } = params;
 
-  if (invocationConfig.dependencyGraphOrder) {
-    throw new Error(
-      'executeStepWithDependencies does not currently support dependencyGraphOrder',
-    );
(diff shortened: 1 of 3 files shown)


packages/integration-sdk-testing/src/executeStepWithDependencies.ts
```diff
@@ -9,16 +9,17 @@ import {
 export async function executeStepWithDependencies(params: StepTestConfig) {
   const { stepId, invocationConfig, instanceConfig } = params;
 
-  if (invocationConfig.dependencyGraphOrder) {
```
ndowmon, line 12: The Google Cloud integration has a limited number of steps that are set with `dependencyGraphId: 'last'`, but the remaining steps do _not_ have a `dependencyGraphId` and inherently execute first.

While we cannot presently use the `executeStepWithDependencies` function on any of the steps with `dependencyGraphId: 'last'`, there's no reason we should be excluded from using it with any of the other steps in that integration.

adam-in-ict (dismissed):

VDubber (approved):
Nice improvement

adam-in-ict (approved):

PR #664, opened 2022-03-25
Allow relationships with undefined properties

ndowmon:
### Changed

- Allow relationships to have `undefined` top-level properties


packages/integration-sdk-testing/src/jest.ts
```diff
@@ -536,7 +536,24 @@ export function toMatchDirectRelationshipSchema<T extends ExplicitRelationship>(
     newRelationshipSchema.properties._type = { const: params._type };
   }
 
-  return toMatchSchema(received, newRelationshipSchema);
+  return toMatchSchema(
+    removeUndefinedProperties(received),
```
ndowmon, line 540: AJV has some strange behavior in that it doesn't treat `undefined ` as `null`. 

Therefore, using the `directRelationshipSchema`, which enforces that all properties must be primitive (no arrays or objects) doesn't allow one to produce relationships with `undefined` properties:

```ts
      {
        additionalProperties: {
          anyOf: [
            { type: 'boolean' },
            { type: 'integer' },
            { type: 'null' },
            { type: 'number' },
            { type: 'string' },
          ],
        },
```

adam-in-ict (approved):

VDubber (approved):
Looks good!

packages/integration-sdk-testing/src/jest.ts
```diff
@@ -536,7 +536,24 @@ export function toMatchDirectRelationshipSchema<T extends ExplicitRelationship>(
...
+
+function removeUndefinedProperties<T extends Entity | ExplicitRelationship>(
+  received: T | T[],
+) {
+  if (!Array.isArray(received)) {
+    received = [received];
+  }
+  return received.map((r) => {
+    for (const [key, value] of Object.entries(r)) {
+      if (value === undefined) delete r[key];
```
austinkelleher, line 553: Should we be mutating the original object or perhaps we should create a new object and copy over _defined_ values?

austinkelleher (approved):
Looks good. Left a minor comment.

VDubber (approved):

PR #666, opened 2022-03-25
INT-3119 - Pass client-generated correlation ID along with sync job uploads

austinkelleher:
FYI @aiwilliams

### DIFF
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -20,6 +20,7 @@ import { ApiClient } from '../api';
 import { timeOperation } from '../metrics';
 import { FlushedGraphObjectData } from '../storage/types';
 import { AttemptContext, retry } from '@lifeomic/attempt';
+import { v4 as uuid } from 'uuid';
 
 export { synchronizationApiError };
 import { createEventPublishingQueue } from './events';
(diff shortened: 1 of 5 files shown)


packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -354,11 +359,33 @@ export async function uploadDataChunk<
...
+      await apiClient.post(
+        `/persister/synchronization/jobs/${jobId}/${type}`,
+        {
+          [type]: batch,
+        },
+        {
+          headers: {
+            // NOTE: Other headers that were applied when the client was created,
+            // are still maintained
+            [RequestHeaders.CorrelationId]: uploadCorrelationId,
```
austinkelleher, line 385: Here is the main change

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -354,11 +359,33 @@ export async function uploadDataChunk<
...
-      });
+    async (ctx) => {
+      logger.info(
+        {
+          uploadCorrelationId,
+          uploadType: type,
+          attemptNum: ctx.attemptNum,
+          batchSize: batch.length,
+        },
+        'Uploading data...',
```
austinkelleher, line 373: Log some data before the upload occurs...

adam-in-ict (approved):

zemberdotnet (approved):

PR #675, opened 2022-03-31
INT-3163 - Add --api-base-url for sync, run, and validate-question CLI commands

adam-in-ict:
Add --api-base-url for sync, run, and validate-question CLI commands.

### DIFF
--- a/docs/integrations/development.md
+++ b/docs/integrations/development.md
@@ -1305,6 +1305,14 @@ the job run. The polling will stop once the job was marked as complete.
 
 ex: `j1-integration sync --instance <integration instance id> --tail`
 
+###### Option `--api-base-url`
+
+If provided this option specifies which base URL to use for synchronization.
+Cannot be used with the --development flag.
+
+ex:
+`yarn j1-integration sync --integrationInstanceId <integration instance id> --api-base-url <api base url>`
+
 #### Command `j1-integration run`
 
 The `j1-integration run` command combines the functionality of the `collect` and
@@ -1417,6 +1425,7 @@ Options:
   -a, --jupiterone-account-id <jupiteroneAccountId>  J1 account ID used to validate J1QL queries
   -k, --jupiterone-api-key <jupiteroneApiKey>        J1 API key used to validate J1QL queries
   -d, --dry-run                                      skip making HTTP requests to validate J1QL queries
+  --api-base-url                                     API base URL used to validate question file.
   -h, --help                                         display help for command
 ```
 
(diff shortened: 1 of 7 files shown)


packages/integration-sdk-cli/src/commands/run.ts
```diff
@@ -55,7 +56,19 @@ export function run() {
       log.debug('Loading account from JUPITERONE_ACCOUNT environment variable');
       const account = getAccountFromEnvironment();
 
-      const apiBaseUrl = getApiBaseUrl({ dev: options.development });
+      let apiBaseUrl: string;
```
adam-in-ict, line 59: This overall logic is used in all three places with the only difference being what `development` mechanism is checked and what error is thrown.  We might look to try and standardize into a helper function when we come back through to add this functionality to additional commands.

docs/integrations/development.md
```diff
@@ -1305,6 +1305,14 @@ the job run. The polling will stop once the job was marked as complete.
 
 ex: `j1-integration sync --instance <integration instance id> --tail`
 
+###### Option `--api-base-url`
+
+If provided this option specifies which base URL to use for synchronization.
+Cannot be used with the --development flag.
```
austinkelleher, line 1311: Nit:

```suggestion
Cannot be used with the `--development` flag.
```

packages/integration-sdk-cli/src/commands/run.ts
```diff
@@ -55,7 +56,19 @@ export function run() {
...
+      if (options.apiBaseUrl) {
+        if (options.development) {
```
austinkelleher, line 61: Super minor nit, but we could reduce the nesting by slightly refactoring to something like:

```ts
if (options.apiBaseUrl && options.development) {
  throw new Error(
    'Invalid configuration supplied.  Cannot specify both --api-base-url and --development(-d) flags.',
  );
}

const apiBaseUrl = options.apiBaseUrl || getApiBaseUrl({
  dev: options.development,
});
```

We could also create a function:

(NOTE: would be nice if `options` was typed, but we could address that later)

```ts
function getApiBaseUrlFromCliArgs(options: any): string {
  if (options.apiBaseUrl && options.development) {
    throw new Error(
      'Invalid configuration supplied.  Cannot specify both --api-base-url and --development(-d) flags.',
    );
  }

  return options.apiBaseUrl || getApiBaseUrl({
    dev: options.development,
  });
}
```

austinkelleher (approved):
Approved with minor suggestions

PR #677, opened 2022-04-04
Adding --api-base-url for j1 export and import commands

adam-in-ict:


packages/cli/src/commands/j1Export.ts
```diff
@@ -45,6 +46,10 @@ export function j1Export() {
       '--no-include-deleted',
       'Exclude deleted entities/relationships in export',
     )
+    .option(
+      '--api-base-url <url>',
+      'The base URL used to initiate api calls with your instance of JupiterOne',
```
austinkelleher, line 51: Would you mind updating this with an example default?

```suggestion
      'The base URL used to initiate api calls with your instance of JupiterOne (defaults to `https://api.us.jupiterone.io`)',
```

austinkelleher (dismissed):
Looks good. Left a minor comment. Please update the docs (separate PR is fine)!

packages/cli/src/commands/j1Export.ts
adam-in-ict: Updated both import and export with above suggestion.  Thanks!

austinkelleher (approved):

PR #688, opened 2022-04-16
INT-3349 - Upgrade Polly.JS

austinkelleher:
The following packages have been upgraded:

- `@pollyjs/adapter-node-http`
- `@pollyjs/core`
- `@pollyjs/persister-fs`

The new Polly.JS packages ship with TypeScript definition files,
so the old `@type/pollyjs__` packages have been removed!

Additionally, `fs-extra` has been pinned to semver range `^9.0.0`
because there a change was released in `^10.0.0` that caused
a regression. See: https://github.com/jprichardson/node-fs-extra/pull/953.

**NOTE**: I also linked the `@jupiterone/integration-sdk-testing` package into an integration and manually tested the following scenarios:

- Run tests with _existing_ recordings
- Delete a recording, generate a _new_ recording, and test against the _new_ recording


package.json
```diff
@@ -35,5 +35,8 @@
     "ts-jest": "^27.1.0",
     "typescript": "^4.2.4"
   },
+  "resolutions": {
+    "fs-extra": "^9.0.0"
```
austinkelleher, line 39: See description for the reason for this

packages/integration-sdk-cli/package.json
```diff
@@ -38,16 +38,13 @@
   },
   "devDependencies": {
     "@jupiterone/integration-sdk-private-test-utils": "^8.11.0",
-    "@pollyjs/adapter-node-http": "^5.1.1",
-    "@pollyjs/core": "^5.1.1",
-    "@pollyjs/persister-fs": "^5.1.1",
+    "@pollyjs/adapter-node-http": "^6.0.5",
```
austinkelleher, line 41: Major version changes were considered and various testing utilities have been updated. CHANGELOG can be found here: https://github.com/Netflix/pollyjs/blob/master/CHANGELOG.md#600-2021-11-30

packages/integration-sdk-cli/package.json
```diff
@@ -38,16 +38,13 @@
...
+    "@pollyjs/core": "^6.0.5",
+    "@pollyjs/persister-fs": "^6.0.5",
     "@types/fs-extra": "^9.0.13",
     "@types/js-yaml": "^4.0.3",
     "@types/json-diff": "^0.5.1",
     "@types/lodash": "^4.14.158",
-    "@types/pollyjs__adapter-node-http": "^2.0.1",
```
austinkelleher, line 48: See description. These are no longer needed! 🥳

packages/integration-sdk-cli/src/__tests__/cli-run-failure.test.ts
```diff
@@ -37,14 +38,7 @@ beforeEach(() => {
   process.env.JUPITERONE_ACCOUNT = 'mochi';
   loadProjectStructure('validationFailure');
 
-  polly = new Polly('run-cli-failure', {
```
austinkelleher, line 40: Small bonus change. Extracted this functionality into `createTestPolly` because it is used in 3 tests.

packages/integration-sdk-cli/src/__tests__/util/recording.ts
```diff
@@ -0,0 +1,16 @@
+import { Polly, PollyConfig } from '@pollyjs/core';
+
+export function createTestPolly(
+  recordingName: string,
+  pollyConfigOverrides?: PollyConfig,
+) {
+  return new Polly(recordingName, {
+    adapters: ['node-http'],
+    persister: 'fs',
+    logLevel: 'silent',
```
austinkelleher, line 10: The only change made when creating new `Polly` instances is here. There used to be a `logging` boolean, which was assigned `false`. In Polly.JS v6, `logging` was removed and replaced with `logLevel`. I kept the existing behavior by assigning `silent` as the value.

packages/integration-sdk-cli/src/__tests__/util/synchronization.ts
```diff
@@ -22,7 +22,7 @@ export function setupSynchronizerApi({ polly, job, baseUrl }: SetupOptions) {
     .post(`${baseUrl}/persister/synchronization/jobs/${job.id}/entities`)
     .intercept((req, res) => {
       allowCrossOrigin(req, res);
-      job.numEntitiesUploaded += JSON.parse(req.body).entities.length;
+      job.numEntitiesUploaded += JSON.parse(req.body!).entities.length;
```
austinkelleher, line 25: The new internal Polly.JS TypeScript definitions allow `body` to be `undefined`. For this test to pass, `body` would need to be defined.

packages/integration-sdk-testing/src/recording.ts
```diff
@@ -78,7 +78,7 @@ export function setupRecording({
       return 'JupiterOneIntegationFSPersister';
     }
 
-    saveRecording(recordingId: number, data: Har): void {
+    async onSaveRecording(recordingId: string, data: Har): Promise<void> {
```
austinkelleher, line 81: Breaking change in Polly.JS v6. See the Polly.JS CHANGELOG.

packages/integration-sdk-testing/package.json
```diff
@@ -25,13 +25,9 @@
   "dependencies": {
     "@jupiterone/integration-sdk-core": "^8.11.0",
     "@jupiterone/integration-sdk-runtime": "^8.11.0",
-    "@pollyjs/adapter-node-http": "^5.1.1",
-    "@pollyjs/core": "^5.1.1",
-    "@pollyjs/persister-fs": "^5.1.1",
-    "@types/har-format": "^1.2.4",
```
austinkelleher, line 31: No longer needed since this data is exported from `@pollyjs/persister`! 🥳

package.json
peteriman: `fs-extra@10.1.0` has been released to address https://github.com/jprichardson/node-fs-extra/pull/953.

peteriman: `fs-extra@10.1.0` has been released to address https://github.com/jprichardson/node-fs-extra/pull/953.
Might want to bump it up instead, maybe?

peteriman, line 39: Related to `fs.realpath.native` being `undefined`, causing import of `fs-extra@10.0.0` to throw error.

Defensively patched in https://github.com/jprichardson/node-fs-extra/pull/953 and released in [`fs-extra@10.1.0`](https://github.com/jprichardson/node-fs-extra/blob/10.1.0/CHANGELOG.md).

Might want to bump it up instead, maybe?

But still, why is `fs.realpath.native` `undefined`?  That is a root cause to be investigated.

peteriman (changes requested):

package.json
austinkelleher: @peteriman Thanks a lot for putting a fix in for this issue **and** for following up here!

It's not clear to me why `fs.realpath.native` is `undefined` either. I may spend a little time trying to figure that out and get back to you if I find anything.

Thanks again!

adam-in-ict (approved):

package.json
peteriman: While `fs-extra` no longer throws error on require/import, if someone still uses `fs. realpath.native` (be it from `fs-extra` or from native `fs`) they would still face errors. 

@austinkelleher You might want to `console.log(typeof fs.realpath.native)` at the first line of code as a pre-requisite, then the subsequently before and after every require/import to identify the offending dependency that "deleted" it.

Issue #643, opened 2022-02-25
Create benchmark for `jobState` functions


austinkelleher:
## Problem

We'd like to understand the impact of `iterateEntities` under certain entity count circumstances. More specifically, we've recently been adopting a pattern where we have a single step that collects entities and a later step that iterates entities and loads raw data. It's possible that is costing us in performance.

Create a new benchmark suite that creates a `jobState` by leveraging the SDK's `createStepJobState` function. We'll need to supply a `FileSystemGraphObjectStore` to the `graphObjectStore` and an `uploader` that consists of a mocked `StepGraphObjectDataUploader`.

Example of a mocked `StepGraphObjectDataUploader`:

```ts
const uploader: StepGraphObjectDataUploader = {
  stepId: 'abc123',
  async enqueue(graphObjectData) {
    return Promise.resolve();
  },

  async waitUntilUploadsComplete() {
    return Promise.resolve();
  },
}
```

All params that you'll need to supply to `createStepJobState`:

```ts
export interface CreateStepJobStateParams {
  stepId: string;
  duplicateKeyTracker: DuplicateKeyTracker;
  typeTracker: TypeTracker;
  graphObjectStore: GraphObjectStore;
  dataStore: MemoryDataStore;
  uploader?: StepGraphObjectDataUploader;
  beforeAddEntity?: (entity: Entity) => Entity;
}
```

Additionally, we'll need to create instances of `DuplicateKeyTracker`, `TypeTracker`, and `MemoryDataStore`.

## Example Tests

Partial logic for the first test would look something like this:

```ts
const e1 = await jobState.addEntity(createMockEntity());
const e2 = await jobState.addEntity(createMockEntity());

await jobState.addRelationship(
  createDirectRelationship({
    from: e1,
    to: e2,
    _class: RelationshipClass.HAS,
  });
});
```

Partial logic for the second test would look something like this:

```ts
await jobState.addEntity(createMockEntity());

await jobState.iterateEntities({ 
  _type: Entities.TEST_TYPE._type 
}, async (e1) => {
  // Get the raw data property...we won't use it though.
  getRawData(e1);
  const e2 = await jobState.addEntity(createMockEntity());

  await jobState.addRelationship(
    createDirectRelationship({
      from: e1,
      to: e2,
      _class: RelationshipClass.HAS,
    });
  });
});
```

PR #691, opened 2022-04-20
[INT-3353] Updating and adding new benchmarks

zemberdotnet:
# Description

This PR adds two new benchmarks, updates an existing benchmark, and adds an `eventCollector` utility to make capturing and displaying benchmark information easier.

This partially addresses issue #643, but there is still more that could be done. This only benchmarks the `jobState.addEntities` function.

## Results
Benchmarks done on a 2019 MacBook Pro | 2.3 GHz 8-Core Intel Core i9 | 16 GB 2667 MHz DDR4

### FileSystemGraphObjectStore#addEntities
FileSystemGraphObjectStore#addEntities 100 Entities        46426.504 ops/sec
FileSystemGraphObjectStore#addEntities 1000 Entities       2479.691 ops/sec
FileSystemGraphObjectStore#addEntities 10000 Entities      128.303 ops/sec
FileSystemGraphObjectStore#addEntities 100000 Entities     9.504 ops/sec
### InMemoryGraphObjectStore#addEntities
InMemoryGraphObjectStore#addEntities 100   66478.234 ops/sec
InMemoryGraphObjectStore#addEntities 1000          3250.880 ops/sec
InMemoryGraphObjectStore#addEntities 10000         148.766 ops/sec
InMemoryGraphObjectStore#addEntities 100000        10.066 ops/sec
### JobState.addEntities
JobState#addEntities 100         25205.276 ops/sec
JobState#addEntities 1000        1600.863 ops/sec
JobState#addEntities 10000       71.557 ops/sec
JobState#addEntities 100000      4.855 ops/sec

### DIFF
--- a/packages/integration-sdk-benchmark/package.json
+++ b/packages/integration-sdk-benchmark/package.json
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "prebenchmark": "rm -rf .j1-integration",
-    "benchmark": "node ./src/index.js"
+    "benchmark": "for file in ./src/benchmarks/*; do yarn prebenchmark && node $file; done"
   },
   "dependencies": {
     "@jupiterone/integration-sdk-core": "^8.12.0",
--- a/packages/integration-sdk-benchmark/src/index.js
+++ b/packages/integration-sdk-benchmark/src/index.js
@@ -1,1 +1,3 @@
 require('./benchmarks/FileSystemGraphObjectStore.addEntities');
+require('./benchmarks/InMemoryGraphObjectStore.addEntities');
+require('./benchmarks/JobState.addEntities');
--- a/packages/integration-sdk-benchmark/src/index.js
+++ b/packages/integration-sdk-benchmark/src/index.js
@@ -1,3 +1,0 @@
-require('./benchmarks/FileSystemGraphObjectStore.addEntities');
-require('./benchmarks/InMemoryGraphObjectStore.addEntities');
-require('./benchmarks/JobState.addEntities');
--- a/packages/integration-sdk-benchmark/src/util/eventCollector.js
+++ b/packages/integration-sdk-benchmark/src/util/eventCollector.js
@@ -1,0 +1,26 @@
+function createEventCollector() {
+  return {
+    events: [],
+    addEvent: function (event) {
+      this.events.push(event);
+    },
+    publishEvents: function () {
+      this.events.sort((a, b) => {
+        return a.target.name > b.target.name;
+      });
+
+      for (const event of this.events) {
+        console.log(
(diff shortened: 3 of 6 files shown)


packages/integration-sdk-benchmark/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "scripts": {
     "prebenchmark": "rm -rf .j1-integration",
-    "benchmark": "node ./src/index.js"
+    "benchmark": "for file in ./src/benchmarks/*; do yarn prebenchmark && node $file; done"
```
zemberdotnet, line 15: Running all the benchmarks through `require`s in `index.js` leads to some performance impacts when running `async` benchmarks. It also can result in the `v8` engine crashing with out-of-memory errors for large tests.

packages/integration-sdk-benchmark/src/index.js
```diff
@@ -1 +1,3 @@
 require('./benchmarks/FileSystemGraphObjectStore.addEntities');
+require('./benchmarks/InMemoryGraphObjectStore.addEntities');
+require('./benchmarks/JobState.addEntities');
```
zemberdotnet, line 3: I'm not sure if removing the `index.js` file is the right thing to do, but it isn't a good way to run all the benchmarks.

packages/integration-sdk-benchmark/src/util/eventCollector.js
```diff
@@ -0,0 +1,26 @@
+function createEventCollector() {
```
VDubber, line 1: This is an awesome add, @zemberdotnet! 

One suggestion to consider is moving this to a `class` for a few reasons:
1. Classes help show the state of the object
2. Classes are easy to reason about and will make it easier for others devs to dive into.

I do link having the `createEventCollector` instead of using the new operator.
Maybe wrap the constructor in the existing `createEventCollector` method.

packages/integration-sdk-benchmark/package.json
VDubber: Looks like there is more for us to learn about the benchmarking package. I think your changes are a great for now!

packages/integration-sdk-benchmark/src/util/eventCollector.js
VDubber: Some jsdocs would be a nice touch.

packages/integration-sdk-benchmark/src/index.js
VDubber: I'd support the removal. A different approach seems to be needed.

VDubber (approved):

VDubber (approved):
Sweet!

PR #695, opened 2022-04-21
INT-3355 - Create initial http client

adam-in-ict:
This PR adds an initial HTTP client for eventual addition to the integration-template.  The intent of this client is to have an interface that is implementation agnostic so that we can switch between HTTP request libraries without impacting downstream users.

Initially this client is using node-fetch, but we would like to eventually switch to undici.


packages/integration-sdk-http-client/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  async executeAPIRequest(request: APIRequest): Promise<APIResponse> {
+    let attempts = 0;
+
+    do {
+      const response = await fetch(request.url, request);
+
+      if (isNonRetryableError(response.status)) {
+        const err = new Error(
+          `API request error for ${request.url}: ${response.statusText}`,
```
VDubber, line 33: Might be beneficial to do 
`API request error for ${request.url}: ${response.code} ${response.statusText}`,

packages/integration-sdk-http-client/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+        );
+        Object.assign(err, { code: response.status });
```
VDubber, line 35: Thoughts around extending Error and allowing `code` as a param?

packages/integration-sdk-http-client/src/util.ts
```diff
@@ -0,0 +1,19 @@
...
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+/**
+ * Returns true when we have a status code that is an error but not retryable.
+ * @param statusCode
+ * @returns boolean
+ */
+export function isNonRetryableError(statusCode: number): boolean {
+  if (statusCode !== 429 && statusCode >= 400) {
```
VDubber, line 15: It depends on the API, but I believe there are a number of status codes > 400 that would be worth retrying.

packages/integration-sdk-http-client/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+      retryAfter:
+        headers.get('X-RateLimit-RetryAfter') &&
+        Number(headers.get('X-RateLimit-RetryAfter')),
+    };
+
+    const unixTimeNow = getUnixTimeNow();
+    const timeToSleepInSeconds = this.rateLimitState.retryAfter
+      ? this.rateLimitState.retryAfter +
+        this.rateLimitConfig.sleepAdditionalSeconds -
+        unixTimeNow
```
VDubber, line 76: Is there a chance that this results in a negative number? Might be worth using Math.max(...., 0)

packages/integration-sdk-http-client/src/index.ts
```diff
@@ -0,0 +1,86 @@
...
+  cooldownPeriod: 1000,
+  sleepAdditionalSeconds: 0,
+};
+
+export type APIResourceIterationCallback<T> = (
+  resources: T[],
+) => boolean | void | Promise<boolean | void>;
+
+export class APIClient {
+  private rateLimitConfig: RateLimitConfig = DEFAULT_RATE_LIMIT_CONFIG;
```
VDubber, line 22: Consider adding a TODO to allow for this config to be customized.

VDubber:
I'm starting to imagine a client that either is configured via 1) passing an object on construction or 2) reads a j1client.config.js file.
It could allow for: 
- Configuration of rate limit details to be found in the headers or in the response body
- Configuration of retryable error status codes
- Configuration of the client. We want to end with Undici as the default but we should consider a way to wrap a providers client (e.g. Octokit from GitHub) as well.
- Configuration of pagination

VDubber (commented):
I like this initial round - simple and straightforward.

packages/integration-sdk-http-client/src/util.ts
adam-in-ict: Yeah, and some APIs use 503 instead of 429 for rate limit errors.  I pulled it out into a function so we could more easily abstract it later on, but didn't put a todo.  I think the range of retryable as well as the rate limit error need to be configurable, possibly as an optional parameter into the constructor?

For now, I'll add a TODO like I should have.

packages/integration-sdk-http-client/src/index.ts
adam-in-ict: I'd thought about that.  I'd also thought about using the errors like `IntegrationError` provided in the integration-sdk-core package, but wasn't sure I liked making it dependent on core right out of the gate.

packages/integration-sdk-http-client/src/__tests__/index.test.ts
```diff
@@ -37,17 +37,15 @@ test('it throws a forbidden error code', async () => {
...
       method: 'GET',
       headers: {
         accept: 'application/json',
       },
-    });
-  } catch (err) {
-    expect(err.code).toEqual(403);
-  }
+    }),
+  ).rejects.toThrowError('API request error for http://testUrl: 403 Forbidden');
```
VDubber, line 48: 🔥

packages/integration-sdk-http-client/src/errors.ts
```diff
@@ -0,0 +1,17 @@
+export interface APIErrorConfig {
```
VDubber, line 1: nit: rename APIErrorConfig to APIErrorParams

packages/integration-sdk-http-client/src/index.ts
```diff
@@ -29,13 +31,14 @@ export class APIClient {
       const response = await fetch(request.url, request);
 
       if (isNonRetryableError(response.status)) {
-        const err = new Error(
-          `API request error for ${request.url}: ${response.statusText}`,
-        );
-        Object.assign(err, { code: response.status });
+        const err = new APIError({
```
VDubber, line 34: 👍

packages/integration-sdk-http-client/src/index.ts
```diff
@@ -19,6 +20,7 @@ export type APIResourceIterationCallback<T> = (
 ) => boolean | void | Promise<boolean | void>;
 
 export class APIClient {
+  // TODO make rateLimitConfig configurable and only use the default values when none are provided
```
VDubber, line 23: nit: Austin recently asked that we start adding our GitHub username and a ticket number into the TODO comments..

VDubber (approved):
Looks nice! 😎

packages/integration-sdk-http-client/src/index.ts
adam-in-ict: Yeah, I saw that and thought it was a good idea but hadn't thought to come back and update these. Good looking out!  Thanks!

VDubber (approved):

PR #704, opened 2022-05-02
INT-3502 Neo4j Write Performance Tweaks (Starbase)

adam-in-ict:
Made two changes to help improve Starbase write performances into Neo4j:
- When available, use the entity _type to help find the start and end nodes for relationship creation
- Store async promises in an array and await at the end of entity and relationship creation

Additional later update:
- Store `_fromType` and `_toType` for all relationships.


packages/integration-sdk-cli/src/neo4j/neo4jGraphStore.ts
```diff
@@ -142,17 +154,20 @@ export class Neo4jGraphStore {
       );
 
       const buildCommand = `
-      MERGE (start {_key: $startEntityKey, _integrationInstanceID: $integrationInstanceID})
-      MERGE (end {_key: $endEntityKey, _integrationInstanceID: $integrationInstanceID})
+      MERGE (start${startEntityType} {_key: $startEntityKey, _integrationInstanceID: $integrationInstanceID})
```
VDubber, line 157: Can you teach me here? This look different

packages/integration-sdk-cli/src/neo4j/neo4jUtilities.ts
```diff
@@ -50,3 +55,30 @@ export function buildPropertyParameters(propList: Object) {
...
+  if (relationship.toType) {
+    return ':' + relationship.toType.toString();
+  } else if (
+    (relationship._mapping as RelationshipMapping)?.targetEntity?._type
+  ) {
+    return (
+      ':' + (relationship._mapping as RelationshipMapping).targetEntity._type
+    );
+  }
+  return '';
```
VDubber, line 83: Empty string over null to keep typing the same?

packages/integration-sdk-cli/src/neo4j/neo4jUtilities.ts
```diff
@@ -50,3 +55,30 @@ export function buildPropertyParameters(propList: Object) {
 
   return propertyParameters;
 }
+
+// Start and end type helper functions.  Prepends a : to any nonempty results for
+// immediate use in a Neo4j command.
+export function getFromType(relationship: Relationship): String {
```
VDubber, line 61: Unit tests?

VDubber (dismissed):
Looks good to my eyes. Some unit tests would be health.

packages/integration-sdk-cli/src/neo4j/neo4jGraphStore.ts
adam-in-ict: So if you are able to add a label to the node in a MERGE statement it lets Neo4j use indexes to find it more quickly (MERGE effectively does a MATCH statement to find it and then either updates or creates accordingly).  The ${startEntityType} and ${endEntityType} are two such potential labels, depending on if the collected data has type information collected (which so far looks like an optional thing to me, but I'm still investigating a bit further).  We have to do string interpolation because (last I checked) Neo4j doesn't allow labels to be parameterized like the other items we have in this Cypher statement.

packages/integration-sdk-cli/src/neo4j/neo4jUtilities.ts
adam-in-ict: Good call out!  I'll plan to add and see about getting another approval.   Thanks!

adam-in-ict: Added in one of the latest commits.

packages/integration-sdk-cli/src/neo4j/neo4jUtilities.ts
```diff
@@ -50,3 +55,25 @@ export function buildPropertyParameters(propList: Object) {
...
+  if (relationship._fromType) {
+    return ':' + relationship._fromType.toString();
+  }
+  return '';
+}
+
+export function getToType(relationship: Relationship): String {
```
VDubber, line 68: The name `getToType` doesn't included that fact that a `:` will be prepended.
Maybe `getToTypeLabel`?

packages/integration-sdk-cli/src/neo4j/neo4jGraphStore.ts
```diff
@@ -82,21 +85,26 @@ export class Neo4jGraphStore {
...
-      });
+      promiseArray.push(
+        this.runCypherCommand(buildCommand, {
+          propertyParameters: propertyParameters,
+          finalKeyValue: finalKeyValue,
+          integrationInstanceID: this.integrationInstanceID,
+        }),
+      );
     }
+    await Promise.all(promiseArray);
```
VDubber, line 96: Are we okay with a single failed promise rejecting the main promise?

packages/integration-sdk-cli/src/neo4j/neo4jGraphStore.ts
```diff
@@ -106,6 +114,10 @@ export class Neo4jGraphStore {
         endEntityKey = sanitizeValue(relationship._toEntityKey.toString());
       }
 
+      //Attempt to get start and end types
+      const startEntityType = getFromType(relationship);
```
VDubber, line 118: nit: startEntityTypeLabel?

packages/integration-sdk-cli/src/neo4j/neo4jGraphStore.ts
adam-in-ict: I'd wondered that too.  Technically we should only get a failed promise if the Neo4j bolt driver fails to respond (which should only happen if we lose connection to the Neo4j database, in which case we're in trouble anyway), but it does give me a bit of concern to be honest.

packages/integration-sdk-cli/src/neo4j/neo4jGraphStore.ts
adam-in-ict: Updated in latest commit.

packages/integration-sdk-cli/src/neo4j/neo4jUtilities.ts
```diff
@@ -50,3 +55,25 @@ export function buildPropertyParameters(propList: Object) {
...
+}
+
+export function getToTypeLabel(relationship: Relationship): String {
+  if (relationship._toType) {
+    return ':' + relationship._toType.toString();
+  } else if (
+    (relationship._mapping as RelationshipMapping)?.targetEntity?._type
+  ) {
+    return (
+      ':' + (relationship._mapping as RelationshipMapping).targetEntity._type
```
ndowmon, line 75: Do we actually support creating mapped relationships in Neo4j?

packages/integration-sdk-core/src/data/createIntegrationRelationship.ts
```diff
@@ -265,7 +265,9 @@ function createRelationship({
     _type,
     _class: relationshipClass,
     _fromEntityKey: fromKey,
+    _fromType: fromType,
```
ndowmon, line 268: Unfortunately, this will cause persister errors, which enforces that only known properties with leading underscores (`_<prop>`) are accepted. Any unknown properties are rejected.

```
22:10:24.820Z ERROR local: Error occurred while synchronizing collected data (integrationInstanceId=58d07123-2eef-4cd6-940c-2974d482222d, synchronizationJobId=7c63088c-87cf-4431-b7ef-663a4a5e3d02, integrationJobId=7c63088c-87cf-4431-b7ef-663a4a5e3d02)
  Error: Error uploading collected data (API response: code=RELATIONSHIP_UPLOAD_FAILED_VALIDATION, message="2 validation error(s):
  - Error in relationships[0]._fromType (reason=NOT_ALLOWED)
  - Error in relationships[0]._toType (reason=NOT_ALLOWED)
  ").
```

We'll need to rethink the approach here. Maybe you could load entity `_key` and `_type` into a map as you load entities, and fetch `_type` by `_key` as you load relationships?

```diff
+ const entityKeyToTypeMap = new Map<string, string>();
  async function handleGraphObjectFile(parsedData: FlushedGraphObjectData) {
    if (parsedData.entities) {
+     for (const { _key, _type } of parsedData.entities) entityKeyToTypeMap.set(_key, _type);
      await store.addEntities(parsedData.entities);
    }
    if (parsedData.relationships) {
      await store.addRelationships(
-       parsedData.relationships
+       parsedData.relationships.map((r) => ({
+         ...r,
+         _toType: entityKeyToTypeMap.get(r._toEntityKey),
+         _fromType: entityKeyToTypeMap.get(r._fromEntityKey),
+       }))
      );
    }
  }
```

This would work decently if you enforce that both entities are always sent to Neo4j before the relationship, but I'm not sure exactly the sequence of things here.

adam-in-ict: Thanks for the catch Nick! 

So is it just that I chose names that were preceded by an underscore?  Because I don't believe there's any firm requirement for the Neo4j upload functionality that they be named that way.  Would simply removing the underscore and using `fromType` and `toType` and treating them like normal optional properties be an acceptable alternative?

adam-in-ict: Also, was the above error you listed from running a sync command?  I can be sure to run that test against any future changes to this code (and maybe add a unit test for it too).

packages/integration-sdk-cli/src/neo4j/neo4jUtilities.ts
adam-in-ict: Yep! Still a few questions around it (for example, we currently respect the `skipTargetCreation` flag, but I'm concerned it will cause odd looking graphs if we haven't already run the job that is responsible for the target's creation).

packages/integration-sdk-core/src/data/createIntegrationRelationship.ts
adam-in-ict: After a bit more testing, it seems that it's probably not ideal to remove the underscore and have it flagged as something other than metadata.  I've removed collection of _fromType and _toType in the latest comit.

ndowmon (approved):

PR #707, opened 2022-05-09
throw upload error when shrunk payload too large

mdaum:
In time it would be great to have the sdk attempt to re-shuffle batches in an attempt to get larger entities uploaded. However, this change will prevent infinite loops when shrinking rawData on entities whose **mapped properties** alone are enough to cause a batch to fail due to upload size > 6MB.

I've seen real-world examples in the Tenable integration where the properties on a **single vulnerability entity* was between 1-2MB. Seeing as batches can be up to 250 entities, just a few of these was enough to cause an upload failure. However, it is important to note that if these entities were in batches of size `1`, the upload would have succeeded.

### DIFF
--- a/packages/integration-sdk-runtime/src/synchronization/__tests__/index.test.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/__tests__/index.test.ts
@@ -34,7 +34,6 @@ import { getExpectedRequestHeaders } from '../../../test/util/request';
 afterEach(() => {
   delete process.env.INTEGRATION_FILE_COMPRESSION_ENABLED;
   restoreProjectStructure();
-  jest.clearAllMocks();
 });
 
 describe('initiateSynchronization', () => {
@@ -492,12 +491,22 @@ describe('uploadDataChunk', () => {
   });
 });
 
-describe('shrinkLargeRawDataUpload', () => {
+describe('shrinkBatchRawData', () => {
+  const logger = createIntegrationLogger({
+    name: 'test',
+  });
+  logger.error = jest.fn();
+  logger.publishErrorEvent = jest.fn();
+  logger.info = jest.fn();
+
+  afterEach(() => {
+    delete process.env.INTEGRATION_FILE_COMPRESSION_ENABLED;
+    restoreProjectStructure();
+    jest.clearAllMocks();
+  });
+
   it('should shrink rawData until batch size is < 6 million bytes', () => {
     const largeData = new Array(500000).join('aaaaaaaaaa');
-    const logger = createIntegrationLogger({
-      name: 'test',
-    });
     const data = [
       {
         _class: 'test',
@@ -548,6 +557,7 @@ describe('shrinkLargeRawDataUpload', () => {
         ],
       },
     ];
+    const startingSize = Buffer.byteLength(JSON.stringify(data));
     const finalData = [
       {
         _class: 'test',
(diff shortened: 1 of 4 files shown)


packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -523,6 +522,26 @@ function getLargestRawDataEntryFromEntity(
...
+ * Intended to be used when logging upload errors due to large payload size.
+ *
+ * @param data
+ * @returns
+ */
+function getPropSizeMapFromEntity(data: Entity): any {
+  const propSizeMap = {};
+
+  for (const [key, value] of Object.entries({
+    ...data,
```
VDubber, line 537: is deconstructing `data` necessary?

mdaum: not at all, holdover from when I was putting other stuff in here beyond data. Great catch!

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -531,6 +550,7 @@ function getLargestRawDataEntryFromEntity(
  */
 export function shrinkRawData(
   data: UploadDataLookup[keyof UploadDataLookup][],
+  logger: IntegrationLogger,
```
adam-in-ict, line 553: I don't necessarily disagree with including a Logger into this function for all the logging we're wanting to do, but we'd originally planned on trying to keep this more as a pure utility function that would just return the results of the shrink process.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -506,9 +507,7 @@ function getLargestEntityFromBatch(
  * @param data
  * @returns
  */
-function getLargestRawDataEntryFromEntity(
-  data: EntityRawData[],
-): EntityRawData {
+function getLargestKeyFromData(data: EntityRawData[]): EntityRawData {
```
adam-in-ict, line 510: Technically, this helper function is returning the entire item, not just the key, but I otherwise like the name change.

packages/integration-sdk-runtime/src/synchronization/index.ts
mdaum: I think `shrinkRawData` is a fairly meaty function that could fail in cases that we are detecting here. In those cases, it is important that we quickly determine the cause of the issue by dumping size data surrounding the batch and its largest entity. I think it is reasonable to include the logger in this function (and let the subsequent functions `shrinkRawData` calls be purely utility) in light of this.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -524,42 +550,92 @@ function getLargestRawDataEntryFromEntity(
...
-      const largestRawDataEntry = getLargestRawDataEntryFromEntity(
-        largestEntity._rawData,
+    if (entityWithLargestRawData?._rawData) {
+      // Find largest element of the _rawData array (typically at index 0, but check to be certain)
+      const largestRawDataEntry = getLargestRawDataEntry(
+        entityWithLargestRawData._rawData,
       );
-
-      // Find largest item within rawData
+      // Find largest item within largest that element
```
adam-in-ict, line 599: nit:  this comment doesn't quite read the way I think you meant it to

packages/integration-sdk-runtime/src/synchronization/index.ts
adam-in-ict: In that case, we can simplify the return value of this function and just log the results at the end, since we'll have access to a logger in here now.

packages/integration-sdk-runtime/src/synchronization/index.ts
mdaum: no it does not :)

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -524,42 +550,92 @@ function getLargestRawDataEntryFromEntity(
...
+  }
+
+  return propSizeMap;
+}
+
+/**
+ * Removes data from the rawData of the largest entity until the overall size
+ * of the data object is less than maxSize (defaulted to MAX_BATCH_SIZE).
+ *
+ * @param batchData
```
VDubber, line 574: nit: update jsdoc with logger and maxBatchSize

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -524,42 +550,92 @@ function getLargestRawDataEntryFromEntity(
...
       );
-
-      // Find largest item within rawData
+      // Find largest item within largest that element
       const largestItemLookup = getLargestItemKeyAndByteSize(
         largestRawDataEntry.rawData,
       );
 
+      // if we can no longer truncate, log and error out
+      if (largestItemLookup.size === sizeOfTruncated) {
```
VDubber, line 605: nit: move this functionality to a separate method for readability.

packages/integration-sdk-runtime/src/synchronization/__tests__/index.test.ts
```diff
@@ -581,28 +591,115 @@ describe('shrinkLargeUpload', () => {
...
-    const shrinkResults = shrinkRawData(data);
-
-    expect(shrinkResults.itemsRemoved).toEqual(3);
+    const shrinkResults = shrinkBatchRawData(data, logger);
+    expect(shrinkResults.itemsRemoved).toEqual(2);
     expect(data).toEqual(finalData);
   });
 });
 
+describe('shrinkLargeUploadDueToProps', () => {
```
VDubber, line 608: I wanted to verify the bug so I ran this test with your fix commented out. Yeah, it didn't stop 😎

VDubber (dismissed):
These changes appear to address the concern. 💯

adam-in-ict (approved):

adam-in-ict (approved):

PR #712, opened 2022-05-12
Fix Starbase Duplicate Nodes

adam-in-ict:
Fix Starbase Duplicate Nodes

Forcing entity creation prior to relationship creation for Neo4j uploads to fix intermittent duplicate node error.

I provided the following detailed info in an inline comment regarding why we need to make this change:

_We have to traverse all graph files twice so that we can first create all entities followed by all relationships.  The current Neo4j data architecture only uses indexes and not constraints due to the fact that we would need a node key constraint consisting of both the _key value as well as the _integrationInstanceID value and that would require users to have an enterprise Neo4j instance instead of just a community version.  Additionally, MERGE commands within relationship creation to create the start and end nodes if they do not yet exist do not always have access to the Type label during creation, and are therefore not always found by the MERGE command within entity creation.  The current workaround is to make sure all entities are created before relationship creation occurs._

### DIFF
--- a/packages/integration-sdk-cli/src/neo4j/uploadToNeo4j.ts
+++ b/packages/integration-sdk-cli/src/neo4j/uploadToNeo4j.ts
@@ -52,8 +52,8 @@ export async function uploadToNeo4j({
     // key constraint consisting of both the _key value as well as the _integrationInstanceID
     // value and that would require users to have an enterprise Neo4j instance instead
     // of just a community version.  Additionally, MERGE commands within relationship 
-    // creation to create the start and end nodes if they do not yet exist are not able 
-    // to set the Type label correctly always, and are therefore not always found by 
+    // creation to create the start and end nodes if they do not yet exist do not always
+    // have access to the Type label during creation, and are therefore not always found by 
     // the MERGE command within entity creation.  The current workaround is to make 
     // sure all entities are created before relationship creation occurs.
     await iterateParsedGraphFiles(handleGraphObjectEntityFiles, pathToData);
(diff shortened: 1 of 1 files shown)


packages/integration-sdk-cli/src/neo4j/uploadToNeo4j.ts
```diff
@@ -36,14 +36,35 @@ export async function uploadToNeo4j({
...
+    // entities followed by all relationships.  The current Neo4j data architecture
+    // only uses indexes and not constraints due to the fact that we would need a node
+    // key constraint consisting of both the _key value as well as the _integrationInstanceID
+    // value and that would require users to have an enterprise Neo4j instance instead
+    // of just a community version.  Additionally, MERGE commands within relationship
+    // creation to create the start and end nodes if they do not yet exist do not always
+    // have access to the Type label during creation, and are therefore not always found by
+    // the MERGE command within entity creation.  The current workaround is to make
+    // sure all entities are created before relationship creation occurs.
+    await iterateParsedGraphFiles(handleGraphObjectEntityFiles, pathToData);
```
VDubber, line 63: Would it be possible to enhance `iterateParsedGraphFiles` to walk the directory only once, caching relationships along the way and then processing them after all entities have been iterated? To help alleviate walking the directory again?

VDubber: Or that change could be here in this file. Collecting the relationships via  the handler and then post processing them?

adam-in-ict: It could for sure, but my vote would be the simple solution for now and a TODO to remove the extra graph directory traversal in the future.  Just to try and keep the change and the time spent fairly small on this effort.

Plus, if we eventually do add in Type information for start and end entities in relationship data, we wouldn't have this constraint any more.

VDubber (approved):

PR #723, opened 2022-05-26
[INT-3542] Add Error Handling To SDK HTTP Client

zemberdotnet:
# Description

This PR adds extensible error handling to the SDK HTTP Client.

### DIFF
--- a/packages/integration-sdk-http-client/src/errors.ts
+++ b/packages/integration-sdk-http-client/src/errors.ts
@@ -3,6 +3,7 @@ export interface APIErrorParams {
   status: number;
   statusText: string;
   endpoint: string;
+  response?: Response;
 }
 
 export class APIError extends Error {
@@ -19,10 +20,13 @@ export class APIError extends Error {
    */
   readonly endpoint: string;
 
+  readonly response?: Response;
+
   constructor(config: APIErrorParams) {
     super(config.message);
     this.status = config.status;
     this.statusText = config.statusText;
     this.endpoint = config.endpoint;
+    this.response = config.response;
   }
 }
--- a/packages/integration-sdk-http-client/src/errors.ts
+++ b/packages/integration-sdk-http-client/src/errors.ts
@@ -1,17 +1,28 @@
 export interface APIErrorParams {
   message: string;
-
-  code: number;
+  status: number;
+  statusText: string;
+  endpoint: string;
 }
 
 export class APIError extends Error {
   /**
-   * Code associated with the error.
+   * Status code associated with the error.
    */
-  readonly code: number;
+  readonly status: number;
+  /**
+   *  Status Text (e.g. "Not Found")
+   */
+  readonly statusText: string;
+  /**
+   * The requested endpoint associated with the error.
+   */
+  readonly endpoint: string;
 
   constructor(config: APIErrorParams) {
     super(config.message);
-    this.code = config.code;
+    this.status = config.status;
+    this.statusText = config.statusText;
+    this.endpoint = config.endpoint;
   }
 }
--- a/packages/integration-sdk-http-client/src/index.ts
+++ b/packages/integration-sdk-http-client/src/index.ts
@@ -1,38 +1,11 @@
 import fetch from 'node-fetch';
-import {
-  APIResponse,
-  APIRequest,
-  APIRequestOptions,
-  RetryConfig,
-  RateLimitConfig,
-} from './types';
+import { APIResponse, APIRequest, APIRequestOptions } from './types';
 import { APIError } from './errors';
 export type APIResourceIterationCallback<T> = (
   resources: T[],
 ) => boolean | void | Promise<boolean | void>;
-import { retiresAvailable } from './util';
-import { defaultIsRetryable } from './defaults';
-
-type ResolvedAPIRequestOptions = {
-  retryConfig: Required<RetryConfig>;
-  rateLimitConfig: Required<RateLimitConfig>;
-};
-
-export const DEFAULT_RATE_LIMIT_CONFIG: RateLimitConfig = {
-  maxAttempts: 5,
-  reserveLimit: 30,
-  cooldownPeriod: 1000,
-  sleepAdditionalSeconds: 0,
-};
-
-export const DEFAULT_REQUEST_OPTIONS: ResolvedAPIRequestOptions = {
-  rateLimitConfig: DEFAULT_RATE_LIMIT_CONFIG,
-  retryConfig: {
-    isRetryable: defaultIsRetryable,
-    maxAttempts: 5 as const,
-    currentAttempt: 0,
-  },
-};
+import { resolveOptions, retiresAvailable } from './util';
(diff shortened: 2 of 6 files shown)


packages/integration-sdk-http-client/src/index.ts
```diff
@@ -19,77 +14,86 @@ export type APIResourceIterationCallback<T> = (
...
+    response.status === 501
+  ) {
+    throw new APIError({
+      message: `${request.method} ${request.url}: ${response.status} ${response.statusText}`,
+      status: response.status,
+      statusText: response.statusText,
+      endpoint: request.url,
+    });
+  }
+};
```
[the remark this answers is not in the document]
zemberdotnet: Here `isRetryable` would likely already have determined that a `404` is not retryable. 

In that case, should we:
- go to the error handler to try to handle it
- not go to the error handler to handle it.

Or should every non-200 go to the error handler regardless of retryablitiy?

packages/integration-sdk-http-client/src/index.ts
```diff
@@ -19,77 +14,86 @@ export type APIResourceIterationCallback<T> = (
...
 export class APIClient {
   // TODO (adam-in-ict) make rateLimitConfig configurable and only use the default values when none are provided
-  private rateLimitConfig: RateLimitConfig = DEFAULT_RATE_LIMIT_CONFIG;
-  private rateLimitState: RateLimitState;
 
-  async executeAPIRequest(request: APIRequest): Promise<APIResponse> {
+  async executeAPIRequest(
+    request: APIRequest,
+    errorHandler?: APIErrorHandlerFunction,
+    isRetryable?: isRetryableFunction,
```
zemberdotnet, line 65: This should probably be request and then an optional `requestOptions` object.

packages/integration-sdk-http-client/src/index.ts
```diff
@@ -19,77 +14,86 @@ export type APIResourceIterationCallback<T> = (
...
           status: response.status,
           statusText: response.statusText,
         };
+      } else {
+        await errorHandler(request, response);
       }
-
       attempts += 1;
-    } while (attempts < this.rateLimitConfig.maxAttempts);
+    } while (isRetryable(attempts, request, response));
```
zemberdotnet, line 90: This is probably not the right place for this, but I'm working out how to separate the `isRetryable` function from `APIErrroHandlerFunction`.

packages/integration-sdk-http-client/src/index.ts
```diff
@@ -19,77 +14,86 @@ export type APIResourceIterationCallback<T> = (
...
-        )
-      : 0;
-    await sleep(timeToSleepInSeconds * 1000);
-    if (
-      this.rateLimitState.limitRemaining &&
-      this.rateLimitState.limitRemaining <= this.rateLimitConfig.reserveLimit
-    ) {
-      await sleep(this.rateLimitConfig.cooldownPeriod);
-    }
-  }
```
zemberdotnet, line 94: This is moved to `./utils` to support error handler function composability.

packages/integration-sdk-http-client/src/util.ts
```diff
@@ -18,3 +18,44 @@ export function isNonRetryableError(statusCode: number): boolean {
...
+    statusCode === 429 ||
+    statusCode === 408 ||
+    (statusCode >= 500 && statusCode !== 501)
+  ) {
+    return true;
+  }
+  return false;
+}
+
+export async function handleRateLimitError(headers, rateLimitConfig) {
```
zemberdotnet, line 33: Not sure about the interface here. I think maybe there should be a unified interface for the sub error handler functions for composability.

packages/integration-sdk-http-client/src/util.ts
```diff
@@ -18,3 +18,44 @@ export function isNonRetryableError(statusCode: number): boolean {
...
+export function isRetryableStatusCode(statusCode: number): boolean {
+  if (
+    statusCode === 429 ||
+    statusCode === 408 ||
+    (statusCode >= 500 && statusCode !== 501)
+  ) {
+    return true;
+  }
+  return false;
+}
```
zemberdotnet, line 31: I like framing logic in a positive sense, but I know we have thought about just retrying everything. @VDubber I would love to talk to you more about how we can do this better.

adam-in-ict: I'd be interested in discussing this as well.  I think as long as we can treat this as a default and let users of the client override it, we're fine.

What you have currently for the retryable/non-retryable codes seems reasonable to me.

packages/integration-sdk-http-client/src/index.ts
```diff
@@ -19,77 +14,86 @@ export type APIResourceIterationCallback<T> = (
...
-      if (isNonRetryableError(response.status)) {
-        const err = new APIError({
-          message: `API request error for ${request.url}: ${response.status} ${response.statusText}`,
-          code: response.status,
-        });
-        throw err;
-      }
+    isRetryable = isRetryable === undefined ? defaultIsRetryable : isRetryable;
 
-      // TODO (adam-in-ict) make rate limit error configurable for instances where an API isn't using the default of 429
```
adam-in-ict, line 41: I'd be interested in leaving this TODO in the code.  We have at least one existing integration that returns 503 for timeouts instead of 429.

packages/integration-sdk-http-client/src/index.ts
```diff
@@ -19,77 +14,86 @@ export type APIResourceIterationCallback<T> = (
...
     throw new APIError({
       message: `Could not complete request within ${attempts} attempts!`,
-      code: 429,
+      status: response.status,
+      statusText: response.statusText,
+      endpoint: request.url,
     });
   }
-
-  //TODO (adam-in-ict) abstract out which Header values we're grabbing for rate limiting.
```
adam-in-ict, line 63: Would we be able to leave this TODO in the code as well?  I'm not sure that the headers we're looking for will always be the case (I'm not actually even totally confident I selected the ones that are the most common).

packages/integration-sdk-http-client/src/errors.ts
```diff
@@ -1,17 +1,28 @@
 export interface APIErrorParams {
   message: string;
-
-  code: number;
+  status: number;
+  statusText: string;
+  endpoint: string;
```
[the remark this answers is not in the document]
adam-in-ict: Like the change here.  Makes sense to align the two, but I'm in favor of not including any additional dependencies if we don't feel like we have to.

packages/integration-sdk-http-client/src/index.ts
zemberdotnet: Yep. I should have brought it over. I'll add it back in!

packages/integration-sdk-http-client/src/index.ts
zemberdotnet: I'll add it back in 😄

zemberdotnet:
Closing in favor of directly using `gaxios`

Issue #626, opened 2022-02-18
Support `--disable-schema-validation` flag in `j1-integration run` command


ndowmon:
We are able to disable schema validation in the collect command with `yarn j1-integration collect --disable-schema-validation`. However, when schema validation is needed for the `run` command, users need to use a workaround:

```
yarn j1-integration collect --disable-schema-validation
yarn j1-integration sync --integrationInstanceId INSTANCE_ID
```

The `run` command should support this flag natively:

```
yarn j1-integration run --disable-schema-validation --integrationInstanceId INSTANCE_ID
```

PR #724, opened 2022-05-27
INT-3984 - Support --disable-schema-validation option in run command

adam-in-ict:
Adding support to be able to pass --disable-schema-validation to the j1-integration run command (fix for Issue #626 )

### DIFF
--- a/packages/integration-sdk-cli/src/__tests__/cli-run.test.ts
+++ b/packages/integration-sdk-cli/src/__tests__/cli-run.test.ts
@@ -80,6 +80,83 @@ test('disables graph object schema validation', async () => {
   expect(process.env.ENABLE_GRAPH_OBJECT_SCHEMA_VALIDATION).toBeUndefined();
 });
 
+test('step should fail if enableSchemaValidation = true', async () => {
+  loadProjectStructure('instanceWithNonValidatingSteps');
+  const job = generateSynchronizationJob();
+
+  setupSynchronizerApi({
+    polly,
+    job,
+    baseUrl: 'https://api.us.jupiterone.io',
+  });
+
+  await createCli().parseAsync([
+    'node',
+    'j1-integration',
+    'run',
+    '--integrationInstanceId',
+    'test',
+  ]);
+
+  expect(log.displaySynchronizationResults).toHaveBeenCalledTimes(1);
+  expect(log.displayExecutionResults).toHaveBeenCalledWith({
+    integrationStepResults: [
+      {
+        id: 'fetch-users',
+        name: 'Fetch Users',
+        declaredTypes: ['my_user'],
+        partialTypes: [],
+        encounteredTypes: [],
+        status: StepResultStatus.FAILURE,
+      },
+    ],
+    metadata: {
+      partialDatasets: {
+        types: ['my_user'],
+      },
+    },
+  });
+});
+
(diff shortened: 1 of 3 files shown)


packages/integration-sdk-cli/src/commands/run.ts
```diff
@@ -41,6 +41,7 @@ export function run() {
       !!process.env.JUPITERONE_DEV,
     )
     .option('--api-base-url <url>', 'API base URL used during run operation.')
+    .option('-V, --disable-schema-validation', 'disable schema validation')
```
zemberdotnet, line 44: Do we use `-V` as the short hand for this in other places?

zemberdotnet (dismissed):

packages/integration-sdk-cli/src/__tests__/cli-run.test.ts
```diff
@@ -61,6 +61,25 @@ test('enables graph object schema validation', async () => {
...
+  await createCli().parseAsync([
+    'node',
+    'j1-integration',
+    'run',
+    '--integrationInstanceId',
+    'test',
+    '--disable-schema-validation',
+  ]);
+
+  expect(process.env.ENABLE_GRAPH_OBJECT_SCHEMA_VALIDATION).toBeUndefined();
```
austinkelleher, line 80: I'd really like if the assertions were against real runtime behavior instead of the value of an environment variable value. Tbh, I don't like that we mutate the value of `ENABLE_GRAPH_OBJECT_SCHEMA_VALIDATION` in the first place. Ideally, we would remove `ENABLE_GRAPH_OBJECT_SCHEMA_VALIDATION` (breaking) completely and only support the CLI arg.

packages/integration-sdk-cli/src/commands/run.ts
adam-in-ict: We do.  I had it match the option used in `j1-integration collect`

packages/integration-sdk-cli/src/__tests__/cli-run.test.ts
adam-in-ict: What would your preference be for this effort?  Go ahead and close this task out and capture a new task to remove `ENABLE_GRAPH_OBJECT_SCHEMA_VALIDATION` completely or go ahead and make those changes in this PR?

austinkelleher: My preference is that we'd make the suggested changes separately. I do think that we should consider improving the test added in this PR to actually assert that schema validation is not being considered _instead_ of just checking the value of an environment variable. 

Take a look at how this test asserts that the step `status` is `SUCCESS` because schema validation is disabled: https://github.com/JupiterOne/sdk/blob/1280faf454cefa6aff1aabd90b4890c1920b88c9/packages/integration-sdk-cli/src/__tests__/cli.test.ts#L215

adam-in-ict: I left the existing `ENABLE_GRAPH_OBJECT_SCHEMA_VALIDATION` tests in place, but added two additional tests that actually test for SUCCESS/FAILURE as outlined in the linked tests.  Thanks!

austinkelleher (approved):

Issue #678, opened 2022-04-04
Add documentation for `j1` CLI


adam-in-ict:
It appears that all of our documentation is against the `j1-integration` CLI.  We need to add documentation on the use of the `j1` CLI as well.

PR #727, opened 2022-05-27
Adding j1 CLI documentation.

adam-in-ict:
Adding documentation for `j1` CLI.

Fix for issue #678

### DIFF
--- a/docs/integrations/development.md
+++ b/docs/integrations/development.md
@@ -1526,3 +1526,63 @@ immediately reflect that in the graph.
 We hope to provide a good interface via this SDK for providing an interface for
 handling events via this SDK and providing commands/utilities for testing events
 to get an understanding of how they may affect the JupiterOne graph.
+
+## j1 CLI
+
+In addition to the `j1-integration` CLI, this project exposes a `j1` CLI for
+`import` and `export` processes.
+
+### Authentication
+
+For commands that require interaction with JupiterOne's API, the CLI will
+provide ways of inputing credentials. To support that, all commands that
+interact with an API will accept an `--api-key` option.
+
+For convenience when developing locally, we will also look for a
+`JUPITERONE_API_KEY` environment variable for an API key to use.
+
+### Supported commands
+
+Currently, the CLI supports a limited interface consisting of only two commands:
+`export` and `import`.
+
+#### Command `j1 export`
+
+```
+Usage: j1 export [options]
+
+Exports account's entities/relationships into csv files
+
+Options:
+  -d, --data-dir <relative_directory>  The directory where entities/relationships will be downloaded (default: ".j1/export")
+  --account <account>                  The JupiterOne account you are exporting entities/relationships from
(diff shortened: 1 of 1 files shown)


docs/integrations/development.md
```diff
@@ -1526,3 +1526,63 @@ immediately reflect that in the graph.
...
+interact with an API will accept an `--api-key` option.
+
+For convenience when developing locally, we will also look for a
+`JUPITERONE_API_KEY` environment variable for an API key to use.
+
+### Supported commands
+
+Currently, the CLI supports a limited interface consisting of only two commands:
+`export` and `import`.
+
```
VDubber, line 1548: Might be nice to include the required options?

zemberdotnet (approved):

docs/integrations/development.md
```diff
@@ -1526,3 +1526,63 @@ immediately reflect that in the graph.
 We hope to provide a good interface via this SDK for providing an interface for
 handling events via this SDK and providing commands/utilities for testing events
 to get an understanding of how they may affect the JupiterOne graph.
+
+## j1 CLI
```
zemberdotnet, line 1530: Might be better to start on line 1490 with the other j1 cli info. Up to you.

adam-in-ict: To be honest, I went back and forth with putting it higher up starting at line 1490.  My main reason for leaving it at the bottom is it felt like it was interrupting the `j1-integration` documentation too much when placed up there.

I'll leave it where it is for now, but we can move it if we start to see its location confusing people.

docs/integrations/development.md
```diff
@@ -1526,3 +1526,63 @@ immediately reflect that in the graph.
...
+
+In addition to the `j1-integration` CLI, this project exposes a `j1` CLI for
+`import` and `export` processes.
+
+### Authentication
+
+For commands that require interaction with JupiterOne's API, the CLI will
```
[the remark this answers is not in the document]
adam-in-ict: I actually copied this from the Authentication of `j1-integration` since they work the exact same way, but I agree that your wording is better.

docs/integrations/development.md
adam-in-ict: Required is a bit odd for these, since so many of them either have a default or will use .env values if provided that way.  

I think the only option truly flagged as a required one is the scope for the import command.  I can get that added in a later PR

docs/integrations/development.md
```diff
@@ -1526,3 +1526,63 @@ immediately reflect that in the graph.
...
+#### Command `j1 import`
+
+```
+Usage: j1 import [options]
+
+Imports exported account entities/relationships into JupiterOne account
+
+Options:
+  -d --data-dir <relative_directory>  The directory where entities and relationships can be found (default: ".j1/export")
+  --scope <scope>                     A unique id that identifies the synchronization job that will be importing your assets, use any id of your choosing.
```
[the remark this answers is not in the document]
adam-in-ict: No default.  It will error if one isn't provided.

PR #730, opened 2022-06-07
INT-3015 Add integrationJobId as optional param when initializing sync job

JakeFerrero:
When initializing the synchronization job via the persister api, allow an
integration job ID to be passed in, and if one is, pass it along to the persister
api client so a new integration job is not created.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ and this project adheres to
 
 ## Unreleased
 
+## [8.13.12] - 2022-06-08
+
+### Changed
+
+- Allow an integration job id to be passed in when initializing syncronization.
+
+## Unreleased
+
 ## [8.13.11] - 2022-05-27
 
 ### Changed


packages/integration-sdk-runtime/src/synchronization/__tests__/index.test.ts
```diff
@@ -59,6 +59,30 @@ describe('initiateSynchronization', () => {
     });
   });
 
+  test('starts a synchronization job with integration job id if provided', async () => {
+    const mockIntegrationJobId = 'test-integration-job-id';
+    const mockSyncJob = generateSynchronizationJob();
+    mockSyncJob.integrationJobId = mockIntegrationJobId;
+
+    const context = createTestContext();
+    context['integrationJobId'] = mockIntegrationJobId;
```
ndowmon, line 68: NBD but just curious - any reason you're not using dot notation here?

ndowmon (approved):
looks good! can you please update the CHANGELOG.md to reflect this contribution?

packages/integration-sdk-runtime/src/synchronization/__tests__/index.test.ts
```diff
@@ -59,6 +59,30 @@ describe('initiateSynchronization', () => {
...
+
+    const postSpy = jest.spyOn(apiClient, 'post').mockResolvedValue({
+      data: { job: mockSyncJob },
+    });
+
+    const synchronizationContext = await initiateSynchronization(context);
+
+    expect(synchronizationContext.job).toEqual(mockSyncJob);
+    expect(postSpy).toHaveBeenCalledTimes(1);
+    expect(postSpy).toHaveBeenCalledWith('/persister/synchronization/jobs', {
```
VDubber, line 79: nit: might be interesting to also add an assertion on the parameters on logger.child()
We want the integrationJobId passed into the logger to be the same as the one passed into the post method, right?

packages/integration-sdk-runtime/src/synchronization/__tests__/index.test.ts
JakeFerrero: I can't use dot notation here because context comes from `createTestContext()`. That function does not return an actual `SynchronizationJobContext`, but an object that looks like one. Specifically, an object with three fields, one of which is _not_ integrationJobId.

That being said, the helper function for generating a sync job context should probably return an actual `SynchronizationJobContext`. I can update it so it does.

austinkelleher: > That being said, the helper function for generating a sync job context should probably return an actual SynchronizationJobContext. I can update it so it does.

That makes sense to me @JakeFerrero.

austinkelleher (approved):
Approving but I think the comments here are valid. Would support if you wanted to tackle comments in a separate PR though. Up to you.

packages/integration-sdk-runtime/src/synchronization/__tests__/index.test.ts
JakeFerrero: Done.

packages/integration-sdk-runtime/src/synchronization/__tests__/index.test.ts
```diff
@@ -504,5 +538,5 @@ function createTestContext() {
     name: 'test',
   });
 
-  return { apiClient, logger, integrationInstanceId: 'test-instance' };
+  return { apiClient, logger, integrationInstanceId: 'test-instance-id' };
```
JakeFerrero, line 541: @ndowmon @austinkelleher @VDubber This helper function and the other helper function that's used in these unit tests - `generateSynchronizationJob` - were out of sync and returning different `integrationInstanceId`s.

I figure I'll leave my 2 cents here. I'm interested to see what other people think. I've been bitten in the past multiple times by helper functions in tests, similar to this instance. So I've started to move away from helpers in unit tests, or at least am much more conservative when I do use them, for two main reasons:
a) I think enumerating mock objects in the tests themselves adds visibility into exactly what and why that object looks the way it does for that particular test. Often times in unit tests, you're going to be altering a field or two of a mock object to cover a different code flow for a specific test.
b) I've found an over abundance of helpers can lead to this exact scenario - where they become disconnected and start returning different things for the same field, despite them should being the same. 

I've also seen arguments for helpers in unit tests. I'm not totally against them, I've just been more careful when making/using them. Just figured I'd throw this in and see what others' opinions are on the topic.

JakeFerrero:
@ndowmon Updated CHANGELOG. I will make a new release in a follow up PR.

PR #731, opened 2022-06-09
Release 8.13.12

JakeFerrero:


CHANGELOG.md
```diff
@@ -15,8 +15,6 @@ and this project adheres to
 
 - Allow an integration job id to be passed in when initializing syncronization.
 
-## Unreleased
```
JakeFerrero, line 18: I think I made a mistake when I updated the changelog in my previous PR. I thought the entry before the one I added (in this case the one below, 8.13.11) was "unreleased," so I also added "unreleased" before my entry as well. But I think that "unreleased" tag at the top of the file is just serving as a category or folder of sorts, since it was added 2 years ago and not removed. So I got rid of the "unreleased" tag I added last PR. Let me know if I'm misunderstanding.

PR #734, opened 2022-06-10
INT-3995 - Dependency visualization

adam-in-ict:
Adding in an additional CLI command to generate a visualization of the step dependencies.

`yarn j1-integration visualize-dependencies`

will generate an HTML file somewhat like this example from graph-google:
image


zemberdotnet:
Look awesome to me!

zemberdotnet (approved):

PR #736, opened 2022-06-12
INT-4145 - Step failure events do not expose original auth errors

### LINKED ISSUE #735: Integration step failure events do not expose original authorization or authentication errors
## Problem

If an integration step throws an `IntegrationProviderAuthorizationError` or an `IntegrationProviderAuthenticationError`, the integration reports a job event that does not include the original details. Instead, the user will see an `UNEXPECTED_ERROR` in their job log.

## Expected behavior

The original authorization or authentication details are exposed to the integration job log instead of an `UNEXPECTED_ERROR`

## Example

Sample test case:

```ts
export async function fetchUsers(
  context: IntegrationStepContext,
): Promise<void> {
  const cause = new Error('api error');

  throw new IntegrationProviderAuthorizationError({
    cause,
    endpoint: 'https://api.myfakeservice.io',
    status: 403,
    statusText: 'status text',
  });
}
```

Example output in job log:

```
Step "Fetch Users" failed to complete due to error. (errorCode="UNEXPECTED_ERROR", errorId="...", reason="Unexpected error occurred executing integration! Please contact us in Slack, or alternatively email us at support@jupiterone.com if the problem continues to occur.") 
```

austinkelleher:
If an integration step threw an `IntegrationProviderAuthorizationError` or an
`IntegrationProviderAuthenticationError`, the integration wasn't reporting a job
event that includes the original details. Instead, the user was seeing an
`UNEXPECTED_ERROR` in their job log.

Fixes #735 - See issue for more details.

### DIFF
--- a/packages/integration-sdk-runtime/src/logger/index.ts
+++ b/packages/integration-sdk-runtime/src/logger/index.ts
@@ -413,18 +413,18 @@ export function createErrorEventDescription(
   let errorCode: string;
   let errorReason: string;
 
-  if (err instanceof IntegrationError) {
+  if (isProviderAuthError(err)) {
+    errorCode = err.code;
+    errorReason = err.message;
+    // add additional instructions to the displayed message
+    // if we know that this is an auth error
+    message += PROVIDER_AUTH_ERROR_HELP;
+  } else if (err instanceof IntegrationError) {
     errorCode = err.code;
     errorReason = err.message;
   } else {
     errorCode = UNEXPECTED_ERROR_CODE;
     errorReason = UNEXPECTED_ERROR_REASON;
-  }
-
-  if (isProviderAuthError(err)) {
-    // add additional instructions to the displayed message
-    // if we know that this is an auth error
-    message += PROVIDER_AUTH_ERROR_HELP;
   }
 
   const nameValuePairs: NameValuePair[] = [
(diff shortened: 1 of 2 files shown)


packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -413,18 +413,18 @@ export function createErrorEventDescription(
   let errorCode: string;
   let errorReason: string;
 
-  if (err instanceof IntegrationError) {
```
ndowmon, line 416: I don't understand what I'm looking at. You're saying that this is returning `false` for `IntegrationProviderAuthenticationError`, and thus using the default unexpected `errorCode` and `errorReason` properties?

This test suggests that the condition here should be returning true, so I'm not sure why this _wouldn't_ be setting the correct error message... 

```ts
test.only('auth', () => {
  const err = new IntegrationProviderAuthenticationError({
    endpoint: '',
    status: '', 
    statusText: '',
  });

  expect(err instanceof IntegrationError).toBe(true);
});
```

Can you try explaining it to me like I'm five? 🙃

austinkelleher: Whoops I got a few things conflated when tackling another bug. I think you're right. I'll close soon.

PR #738, opened 2022-06-17
Add --buffer-size option for collect and run commands.

adam-in-ict:
Add --buffer-size option for collect and run commands.

### DIFF
--- a/packages/integration-sdk-cli/src/commands/collect.ts
+++ b/packages/integration-sdk-cli/src/commands/collect.ts
@@ -43,7 +43,7 @@ export function collect() {
     .option('-V, --disable-schema-validation', 'disable schema validation')
     .option(
       '-b, --buffer-size <number>',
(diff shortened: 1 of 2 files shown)


packages/integration-sdk-cli/src/commands/collect.ts
```diff
@@ -41,6 +41,10 @@ export function collect() {
       'Can be used with the `--step` to specify a path to a non-default cache location.',
     )
     .option('-V, --disable-schema-validation', 'disable schema validation')
+    .option(
+      '-b, --buffer-size <number>',
+      'specify buffer size for use in data collection (default 500)',
```
zemberdotnet, line 46: Could we add units? I assume this means 500 objects?

adam-in-ict: Updated in latest commit.

zemberdotnet (approved):

PR #741, opened 2022-06-17
Add skipped steps to event log

VDubber:
## Purpose
The desired result of this change is to provide users an understanding of all possible steps available in an integration

## How?
Before a step is executed, the stepStartState is reference to see if the step is disabled. Prior to this change, if the step was disabled nothing would happen. Now, we add a job event log for the step indicating to the user that the step was disabled. Within the StepStartState, each step can be provided a reason for why it is disabled. It is the responsibility of the graph-* project to indicate why it was disabled. 

I believe this a positive change to exposing more information to the user about their integrations.

### DIFF
--- a/packages/integration-sdk-core/src/types/step.ts
+++ b/packages/integration-sdk-core/src/types/step.ts
@@ -11,6 +11,13 @@ import {
 import { IntegrationInstanceConfig } from './instance';
 import { RelationshipDirection } from './relationship';
 
+export enum DisabledStepReason {
+  NONE = 'none', // No reason was provided
+  PERMISSION = 'permission', // Missing permission disabled this step
+  BETA = 'beta', // Step is in beta and only enabled on request
+  CONFIG = 'config', // Step was disabled via config
+}
+
 export interface StepStartState {
   /**
    * Indicates the step is disabled and should not be
@@ -19,10 +26,10 @@ export interface StepStartState {
   disabled: boolean;
 
   /**
-   * Message to user describing why the step was disabled.
-   * Should be user-friendly!
+   * Allows for a user-friendly message to explain
+   * why a step is disabled.
    */
-  disabledReason?: string;
+  disabledReason?: DisabledStepReason;
 
   /**
    * Provides a filepath to a cache for the given step.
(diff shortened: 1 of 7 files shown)


packages/integration-sdk-core/src/types/step.ts
```diff
@@ -18,6 +18,12 @@ export interface StepStartState {
    */
   disabled: boolean;
 
+  /**
+   * Message to user describing why the step was disabled.
+   * Should be user-friendly!
+   */
+  disabledReason?: string;
```
VDubber, line 25: Instead of free string, move to a enum and then build off of that.

packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -286,6 +286,16 @@ export class IntegrationLogger
...
   }
 
+  stepSkip(step: StepMetadata, reason?: string) {
+    const name = 'step_skip';
+    let description = `Skipped step "${step.name}".`;
+    if (reason) {
+      description += ` ${reason}`;
+    }
+    this.info(description);
+    this.publishEvent({ name, description });
```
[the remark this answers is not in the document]
VDubber: Thank you for bring this up! I have adjusted the logic to only publish this event if a `disabledReason` is provided. So it is now opt-in feature.

packages/integration-sdk-core/src/types/step.ts
```diff
@@ -11,13 +11,26 @@ import {
 import { IntegrationInstanceConfig } from './instance';
 import { RelationshipDirection } from './relationship';
 
+export enum DisabledStepReason {
+  NONE = 'none', // No reason was provided
+  PERMISSION = 'permission', // Missing permission disabled this step
+  BETA = 'beta', // Step is in beta and only enabled on request
+  CONFIG = 'config', // Step was disabled via config
```
adam-in-ict, line 18: Is there any interest in differentiating between a step that was explicitly skipped due to one of the listed reasons and a step that needs to be skipped due to a dependency being skipped?

adam-in-ict (approved):

PR #747, opened 2022-06-28
Adding Relationship Upload Batch Size option

zemberdotnet:
# Description

Relationships are often much smaller and much more certain in size than entities. Allowing a separate batch size for relationships would allow us to upload much faster.


zemberdotnet:
> I like it!
> 
> Would we be able to add this flag to the `sync` action as well?

We could definitely do that 👍

austinkelleher (approved):

adam-in-ict (approved):

PR #752, opened 2022-07-05
Add j1-integration diff options

ndowmon:
### Added

- Added the following arguments to `j1-integration diff` command:
  ```
  -i, --ignore-types <type-1>,<type-2>  ignore graph objects with _type: <type> (default: [])
  -S, --ignore-system-mapper            ignore graph objects with _source: 'system-mapper'
  -I, --ignore-system-internal          ignore graph objects with _source: 'system-internal'
  ```

### DIFF
--- a/packages/integration-sdk-cli/src/commands/diff.ts
+++ b/packages/integration-sdk-cli/src/commands/diff.ts
@@ -3,6 +3,12 @@ import fs from 'fs';
 import { diffString } from 'json-diff';
 import path from 'path';
 
+// coercion function to collect multiple values for a flag
+const collector = (value: string, arr: string[]) => {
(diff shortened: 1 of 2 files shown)


JakeFerrero (approved):

packages/integration-sdk-cli/src/commands/diff.ts
```diff
@@ -184,6 +258,16 @@ function diffRelationships(
 
   const newRelationships: DiffableRelationships = {};
   for (const edge of newEdges) {
+    if (options?.ignoreSystemMapper) {
```
VDubber, line 261: It appears that very similar logic is being repeated. Can we extract some of this out and include unit tests?

PR #767, opened 2022-08-04
INT-4596: Update parseTimePropertyValue 

VDubber:
…to correctly handle numeric time values in seconds.

Add tests against parseTimePropertyValue


packages/integration-sdk-core/src/data/converters.ts
```diff
@@ -179,11 +179,40 @@ export function getTime(
  * Date(time).getTime()`.
  *
  * @param time a time value
+ * @param precision - the precision of the numeric value being supplied. Used to determine how to convert to ms.
+ *          ms - milliseconds
+ *          sec - seconds
  */
 export function parseTimePropertyValue(
   time: Date | string | number | undefined | null,
+  precision?: 'ms' | 'sec',
```
zemberdotnet, line 188:
```suggestion
  sourcePrecision?: 'ms' | 'sec',
```

packages/integration-sdk-core/src/data/converters.ts
```diff
@@ -179,11 +179,40 @@ export function getTime(
  * Date(time).getTime()`.
  *
  * @param time a time value
+ * @param precision - the precision of the numeric value being supplied. Used to determine how to convert to ms.
+ *          ms - milliseconds
+ *          sec - seconds
  */
 export function parseTimePropertyValue(
```
zemberdotnet, line 186: We don't currently do this, but I think this may be a good place to use overloaded function signatures:
```ts
export function parseTimePropertyValue(time: Date): number | undefined;
export function parseTimePropertyValue(time: number, sourcePrecision: 'ms' | 'sec'): number | undefined
// other overloaded signature here
// source signature
export function parseTimePropertyValue(time: Date | string | number | undefined | null, sourcePrecision?: 'ms' | 'sec'): number | undefined;
```

Note the ability to require that `sourcePrecision` is required when `time` is `number`. This allows prohibit these errors in the future without changing requirements for many already working cases.

zemberdotnet (approved):

packages/integration-sdk-core/src/data/__tests__/converters.test.ts
```diff
@@ -150,3 +151,56 @@ describe('#truncateEntityPropertyValue', () => {
...
+    );
+    expect(parseTimePropertyValue(1601942400.99999, 'sec')).toBe(1601942400999);
+    expect(parseTimePropertyValue(1601942400.0083234234, 'sec')).toBe(
+      1601942400008,
+    );
+    // Milliseconds
+    expect(parseTimePropertyValue(1601942400000, 'ms')).toBe(1601942400000);
+    expect(() => parseTimePropertyValue(1601942400.001, 'ms')).toThrowError(
+      new Error(
+        'Argument time must not be a float when precision is set to ms.',
```
adam-in-ict, line 202: Strictly speaking I believe the Date() constructor will allow a floating point number to be passed to it, it'll just ignore everything after the decimal.  Feels more like a time to warn instead of throwing an error.

packages/integration-sdk-core/src/data/converters.ts
```diff
@@ -179,11 +179,40 @@ export function getTime(
...
+    // Require precision if time is a number
+    if (typeof time === 'number' && !isNaN(time)) {
+      if (!precision) {
+        throw new Error(
+          'Argument precision is required when parsing a number.',
+        );
+      } else if (precision === 'sec') {
+        time = Number(time) * 1000;
+      } else if (precision === 'ms' && !Number.isInteger(time)) {
+        throw new Error(
```
adam-in-ict, line 210: Strictly speaking I believe the Date() constructor will allow a floating point number to be passed to it, it'll just ignore everything after the decimal. Feels more like a time to warn instead of throwing an error.

VDubber: I agree.
I had a failing test case that I was dealing with but this isn't the solution. Correcting

packages/integration-sdk-core/src/data/converters.ts
```diff
@@ -169,7 +169,7 @@ export function convertNameValuePairs(
  * @see parseTimePropertyValue
  */
 export function getTime(
-  time: Date | string | number | undefined | null,
+  time: Date | string | undefined | null,
```
VDubber, line 172: Another breaking change. I feel great about this b/c:
1. This method is already deprecated
2. And instead of enhancing it we are removing functionality

adam-in-ict (approved):

zemberdotnet (approved):

PR #768, opened 2022-08-05
add optional option for instance config field

### LINKED ISSUE #504: Allow optional `instanceConfigField`
Proposal:

```typescript
export interface IntegrationInstanceConfigField {
  type?: 'string' | 'boolean';
  mask?: boolean;
  optional?: boolean;
}
```

Nick-NCSU:
closes #504 int-4941

Adds in `optional` option for instance config fields.

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/config.ts
+++ b/packages/integration-sdk-runtime/src/execution/config.ts
@@ -26,14 +26,20 @@ export function loadConfigFromEnvironmentVariables<
 
       const environmentVariableValue = process.env[environmentVariableName];
 
-      const value = getConfigValueForField({
+      if (environmentVariableValue === undefined) {
+        if (config.optional) {
+          return [field, undefined];
+        } else {
+          throw configFieldMissingError(field, environmentVariableName);
+        }
+      }
+      const convertedValue = convertEnvironmentVariableValueForField(
         field,
         config,
-        environmentVariableName,
         environmentVariableValue,
-      });
+      );
 
-      return [field, value];
+      return [field, convertedValue];
     })
     .reduce((acc: Record<string, string | boolean>, [field, value]) => {
       if (value !== undefined) {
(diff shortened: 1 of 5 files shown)


packages/integration-sdk-runtime/src/execution/config.ts
```diff
@@ -21,28 +21,53 @@ export function loadConfigFromEnvironmentVariables<
...
+  const { field, config, environmentVariableName, environmentVariableValue } =
+    params;
+  if (environmentVariableValue === undefined) {
+    if (config.defaultValue !== undefined) {
+      return config.defaultValue;
+    }
+    if (!config.optional) {
+      throw configFieldMissingError(field, environmentVariableName);
+    }
+    return undefined;
```
zemberdotnet, line 61:
```suggestion
    if (config.optional) {
       return config.defaultValue;
    } else {
      throw configFieldMissingError(field, environmentVariableName);
    }
```
The `defaultValue` is either a set value or `undefined`, correct? I think this frames the conditional in a positive reference and makes it slightly simpler with the same behavior. Let me know if I'm missing something.

Nick-NCSU: Yeah, the `defaultValue` is a set value or `undefined`. This looks like it has the same behavior and is definitely a lot simpler.

Nick-NCSU: Ah, but it relies on `optional` being set (which might or might not make sense)

Nick-NCSU: Thoughts on whether a `defaultValue` should require `optional` to be `true`?

Currently this would fail
```ts
exampleVariable: {
      type: 'string',
      defaultValue: 'otherString',
}
```

because it explicitly requires optional to be true
```ts
exampleVariable: {
      type: 'string',
      defaultValue: 'otherString',
      optional: true,
}
```

zemberdotnet: Hmm, yeah I see. Sorry I broke your tests 😬 
I'm not sure what the best behavior is. I think the use of `defaultValue` should be gated by `optional === true`.

A required field shouldn't need a `defaultValue` since it should always be defined.

zemberdotnet: I guess the thing is right now `optional` only exists as a way to get to `defaultValue`. In your earlier examples `defaultValue` implied that the field was `optional` even if `optional` was not set. I don't have enough experience to know if this is a good or bad idea, but here is an idea.

```diff
export interface IntegrationInstanceConfigField {
  type?: 'string' | 'boolean';
  /**
   * Defines a default value for the config.
   * Overrides the optional parameter if defined.
   */
-  defaultValue?: string | boolean;
+  defaultValue?: string | boolean | undefined;
  mask?: boolean;
-  optional?: boolean;
}
```

```ts
function getConfigValueForField(params: {
  field: string;
  config: IntegrationInstanceConfigField;
  environmentVariableName: string;
  environmentVariableValue?: string;
}): string | boolean | undefined {
  const { field, config, environmentVariableName, environmentVariableValue } =
    params;
  if (environmentVariableValue === undefined) {
    // explicit presence of `defaultValue` in the `IntegrationConfigField`
    if ('defaultValue' in config) {
     return config.defaultValue;
    } else {
      throw configFieldMissingError(field, environmentVariableName)
    }
  } else {
    return convertEnvironmentVariableValueForField(
      field,
      config,
      environmentVariableValue,
    );
  }
}
```
That being said I need to check over the logic for this again 😆

ndowmon (approved):
Looks good! Thanks!

PR #776, opened 2022-08-23
INT-5336 Add warn log statement for when our handleError function throws

JakeFerrero:
When we attempt to upload graph data and the upload fails, we retry using the
handleUploadDataChunkError function. However, if that function throws, we bail
immediately, which seems tobe happening in prod. As we investigate, add a warning
log if this happens.

### DIFF
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -338,11 +338,6 @@ function handleUploadDataChunkError({
   batch,
   uploadCorrelationId,
 }: HandleUploadDataChunkErrorParams): void {
-  if (err.code !== 'CredentialsError') {
-    // we don't want to log on credential errors
-    logger.warn({ error: err }, 'Error uploading graph data chunk', err);
-  }
-
   /**
    * The JupiterOne system will encapsulate error details in the response in
    * some situations. For example:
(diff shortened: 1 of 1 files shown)


ndowmon (approved):

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -421,13 +416,22 @@ export async function uploadDataChunk<
...
       factor: 1.05,
       handleError(err, attemptContext) {
-        handleUploadDataChunkError({
-          err,
-          attemptContext,
-          logger,
-          batch,
-          uploadCorrelationId,
-        });
+        try {
```
JakeFerrero, line 419: No functionality changes - just adding a warning log if our handleError function throws, which it seems to be doing rarely in prod.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -338,11 +338,6 @@ function handleUploadDataChunkError({
   batch,
   uploadCorrelationId,
 }: HandleUploadDataChunkErrorParams): void {
-  if (err.code !== 'CredentialsError') {
```
JakeFerrero, line 341: This old log statement I added yesterday is no longer needed.

PR #780, opened 2022-08-24
INT-5226 Fix the getPropSizeMapFromEntity function when shrinking _rawData

JakeFerrero:
When attempting to shrink _rawData, the helper function getPropSizeMapFromEntity
may be called. It goes through each key/value pair of a given entity and records
how large each field of the entity is. However, some values of an entity can be undefined,
which will cause Buffer.byteLength to throw. If there is an undefined value, set the size
to 0.


mdaum (approved):
good find

VDubber (approved):

PR #782, opened 2022-08-24
INT-5226 Update getPropSizeMapFromEntity and add new unit test

JakeFerrero:
Updated the getPropSizeMapFromEntity helper function so that it only sets
a size of 0 to undefined properties, not all falsey properties.
Additionally added a unit test that tests this new functionality.

### DIFF
--- a/packages/integration-sdk-runtime/src/synchronization/__tests__/shrinkBatchRawData.test.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/__tests__/shrinkBatchRawData.test.ts
@@ -304,7 +304,6 @@ describe('shrinkBatchRawData', () => {
       shrinkBatchRawData(data, logger);
       shrinkBatchRawDataSucceeded = true;
     } catch (err) {
-      expect(shrinkBatchRawDataSucceeded).toEqual(false);
       expect(err).toBeInstanceOf(IntegrationError);
       expect(logger.error).toBeCalledTimes(1);
       // should give details on largest entity in batch after finished shrinking, this should be item with _key=testKey3
@@ -333,5 +332,7 @@ describe('shrinkBatchRawData', () => {
       expect(logger.info).toBeCalledTimes(1);
       expect(logger.info).toHaveBeenCalledWith('Attempting to shrink rawData');
     }
+
+    expect(shrinkBatchRawDataSucceeded).toEqual(false);
   });
 });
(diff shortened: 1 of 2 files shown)


packages/integration-sdk-runtime/src/synchronization/__tests__/shrinkBatchRawData.test.ts
```diff
@@ -261,4 +261,74 @@ describe('shrinkBatchRawData', () => {
...
+            rawData: {
+              testRawData: 'test123',
+              testFinalData: 'test789',
+            },
+          },
+        ],
+      },
+    ];
+    try {
+      shrinkBatchRawData(data, logger);
```
austinkelleher, line 303: Minor: Can you add an assertion to ensure that this function actually throws. Otherwise, this test could skip all assertions in the below `catch`.

```ts
let shrinkBatchRawDataSucceeded = false;

try {
  shrinkBatchRawData(data, logger);
  shrinkBatchRawDataSucceeded = true;
} catch (err) {

}

expect(shrinkBatchRawDataSucceeded).toEqual(false);
```

packages/integration-sdk-runtime/src/synchronization/__tests__/shrinkBatchRawData.test.ts
```diff
@@ -261,4 +261,74 @@ describe('shrinkBatchRawData', () => {
...
+    } catch (err) {
+      expect(err).toBeInstanceOf(IntegrationError);
+      expect(logger.error).toBeCalledTimes(1);
```
austinkelleher, line 306: This is a separate issue, but I kind of wish that this function didn't have access to the logger at all. It would be much cleaner if relevant properties were attached to the thrown error object. The logging could happen by the caller instead of inside of this utility. Separation of concerns!

packages/integration-sdk-runtime/src/synchronization/shrinkBatchRawData.ts
```diff
@@ -220,7 +220,7 @@ function getPropSizeMapFromEntity(data: Entity): any {
   const propSizeMap = {};
 
   for (const [key, value] of Object.entries(data)) {
-    if (!value) {
+    if (value === undefined) {
```
austinkelleher, line 223: 👍

packages/integration-sdk-runtime/src/synchronization/__tests__/shrinkBatchRawData.test.ts
```diff
@@ -261,4 +261,74 @@ describe('shrinkBatchRawData', () => {
...
+            },
+          },
+        ],
+      },
+      {
+        _class: 'test',
+        _key: 'testKey2',
+        _type: 'testType',
+        largeProperty: largeData,
+        undefProp: undefined,
```
austinkelleher, line 289: 👍

packages/integration-sdk-runtime/src/synchronization/__tests__/shrinkBatchRawData.test.ts
JakeFerrero: @austinkelleher Done.

austinkelleher (approved):

PR #785, opened 2022-09-01
feat: decouple buffer size from file size

zemberdotnet:
# Description

This PR decouples the size of the graph object files from the size of the buffer.
The goals of this is to allow for the use of much larger buffers (for larger uploads), while 
not causing poor performance of `findEntity` due to larger file sizes.


packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -41,6 +43,11 @@ export interface FileSystemGraphObjectStoreParams {
    */
   graphObjectBufferThreshold?: number;
 
+  /**
+   * The maximum number of entities/relationships stored in each file.
+   */
+  graphObjectFileSize?: number;
```
zemberdotnet, line 49: Do we foresee any need for different file sizes for relationships and entities?

austinkelleher (approved):

PR #786, opened 2022-09-03
Adding option to specify a different database name for Neo4j commands

adam-in-ict:
This PR adds the option to specify a different database name other than the default `neo4j` for Neo4j commands. This will allow us to add this requested functionality to Starbase.

### DIFF
--- a/packages/integration-sdk-cli/src/neo4j/__tests__/neo4jGraphStore.test.ts
+++ b/packages/integration-sdk-cli/src/neo4j/__tests__/neo4jGraphStore.test.ts
@@ -86,6 +86,20 @@ describe('#neo4jGraphStore', () => {
     expect(spy).toHaveBeenCalledTimes(1);
   });
 
+  test('should generate call to create a driver connection to a database other than the default', () => {
+    const spy = jest.spyOn(neo4j, 'driver').mockReturnValue(mockDriverResp);
+
+    const emptyStore = new Neo4jGraphStore({
+      uri: '',
+      username: '',
+      password: '',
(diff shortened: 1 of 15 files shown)


packages/integration-sdk-cli/src/neo4j/__tests__/neo4jGraphStore.test.ts
```diff
@@ -86,6 +86,20 @@ describe('#neo4jGraphStore', () => {
...
+
+    const emptyStore = new Neo4jGraphStore({
+      uri: '',
+      username: '',
+      password: '',
+      integrationInstanceID: testInstanceID,
+      database: 'testDatabase',
+    });
+    expect(async () => await emptyStore.close()).toReturn;
+    expect(spy).toHaveBeenCalledTimes(1);
```
austinkelleher, line 100: Should we add an assertion that the spy was called with specific args? (e.g. the specific database that was supplied)

austinkelleher (approved):
Nice change! Left a small test suggestion.

packages/integration-sdk-cli/src/neo4j/__tests__/neo4jGraphStore.test.ts
adam-in-ict: Unfortunately, the neo-forgery class doesn't appear to have support for easily verifying what database is being specified when running (and we don't use the database name until the session is created and a command is run).  I'm going to add in a TODO for now to revisit this, because I agree it'd be a great test.

zemberdotnet (approved):

PR #792, opened 2022-09-15
Update docs/development with new options and clean up a bit

aiwilliams:


docs/integrations/development.md
```diff
@@ -36,25 +36,25 @@ required for authenticating with provider APIs and otherwise necessary to
...
 Example:
 
 ```typescript
 {
   clientId: {
-    type: 'string',
-    mask: false,
-  },
+    type: 'string';
+    mask: false;
```
zemberdotnet, line 48: This is `instanceConfigFields` not a type definition, so I think the `,` are correct.

austinkelleher (approved):

zemberdotnet (commented):
Great improvements!

docs/integrations/development.md
aiwilliams: Good catch! I'll take another look at this in the release PR. My editor was complaining about it and these changes made the complaints go away. 😬

PR #796, opened 2022-09-17
Upgrade Commander and improve CLI consistency and testing

aiwilliams:
Please see CHANGELOG.

### DIFF
--- a/packages/integration-sdk-cli/src/commands/options.ts
+++ b/packages/integration-sdk-cli/src/commands/options.ts
@@ -1,51 +1,93 @@
-import { getApiBaseUrl } from '@jupiterone/integration-sdk-runtime';
+import {
+  DEFAULT_UPLOAD_BATCH_SIZE,
+  getAccountFromEnvironment,
+  getApiKeyFromEnvironment,
+  JUPITERONE_DEV_API_BASE_URL,
+  JUPITERONE_PROD_API_BASE_URL,
+} from '@jupiterone/integration-sdk-runtime';
+import { Command, OptionValues } from 'commander';
+import path from 'path';
 
-/**
- * Determines the JupiterOne `apiBaseUrl` based on these mutually-exclusive `options`:
- * - --api-base-url - a specific URL to use
- * - --development - use the development API
- *
- * @throws Error if both --api-base-url and --development are specified
- */
-export function getApiBaseUrlOption(options: any): string {
-  let apiBaseUrl: string;
-  if (options.apiBaseUrl) {
-    if (options.development) {
-      throw new Error(
-        'Invalid configuration supplied.  Cannot specify both --api-base-url and --development(-d) flags.',
-      );
-    }
-    apiBaseUrl = options.apiBaseUrl;
-  } else {
-    apiBaseUrl = getApiBaseUrl({
-      dev: options.development,
-    });
-  }
-  return apiBaseUrl;
+export interface PathOptions {
+  projectPath: string;
 }
 
-interface SynchronizationJobSourceOptions {
+export function addPathOptionsToCommand(command: Command): Command {
+  return command.option(
+    '-p, --project-path <directory>',
+    'path to integration project directory',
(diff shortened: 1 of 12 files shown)


aiwilliams:
I just realized I need to do a bit more to make `--skip-finalize` work in the `sync` command...

aiwilliams:
Good to go!

packages/integration-sdk-cli/src/commands/options.ts
```diff
@@ -1,51 +1,93 @@
...
+          return value;
+        }
+      },
+      'integration-managed',
+    )
+    .option('--scope <anystring>', 'specify synchronization job scope value')
+    .option(
+      '-u, --upload-batch-size <number>',
+      'specify number of items per batch for upload',
+      (value, _previous: Number) => Number(value),
```
VDubber, line 71: is _previous still being used here?

VDubber: nit: items -> entities or relationships

packages/integration-sdk-cli/src/commands/options.ts
```diff
@@ -1,51 +1,93 @@
...
+      DEFAULT_UPLOAD_BATCH_SIZE,
+    )
+    .option(
+      '-ur, --upload-relationship-batch-size <number>',
+      'specify number of relationships per batch for upload',
+      (value, _previous: Number) => Number(value),
+      DEFAULT_UPLOAD_BATCH_SIZE,
+    )
+    .option('--skip-finalize', 'skip synchronization finalization', false);
```
VDubber, line 80: consider adding what the end result of this is. Something like: The Sync job will remain open to receive additional data.

packages/integration-sdk-cli/src/commands/options.ts
```diff
@@ -55,19 +97,90 @@ export function getSynchronizationJobSourceOptions(
...
+    uploadRelationshipBatchSize: options.uploadRelationshipBatchSize,
+    skipFinalize: options.skipFinalize,
+  };
+}
+
+export interface ApiClientOptions {
+  development: boolean;
+  apiBaseUrl: string;
+  apiKey?: string;
+  account?: string;
```
VDubber, line 132: Consider using `accountId`

packages/integration-sdk-cli/src/commands/options.ts
```diff
@@ -55,19 +97,90 @@ export function getSynchronizationJobSourceOptions(
...
+      'specify synchronization API base URL',
+      JUPITERONE_PROD_API_BASE_URL,
+    )
+    .option(
+      '-d, --development',
+      '"true" to target apps.dev.jupiterone.io (JUPITERONE_DEV environment variable)',
+      !!process.env.JUPITERONE_DEV,
+    )
+    .option(
+      '--account <account>',
```
VDubber, line 148: ditto: `accountId`

VDubber (dismissed):
Thanks for getting these upgrade and cleaned up.

packages/integration-sdk-cli/src/commands/options.ts
aiwilliams: I think we'll have to do this one in another PR that allows for using both and deprecating one, since as I understand it, this was an option already for some commands.

packages/integration-sdk-cli/src/commands/options.ts
aiwilliams: These reflect the names of the options and this would be updated alongside a move to `--account-id`.

packages/integration-sdk-cli/src/commands/options.ts
aiwilliams: > is _previous still being used here?

This argument is required on the function provided to `option()` to make the typings work correctly. Without the second argument defined, it defaults to only allowing a `string` or `boolean` (I think it was) to be returned.

VDubber (approved):

PR #797, opened 2022-09-20
remove multipleResolves event from onFailure

zemberdotnet:
The `multipleResolves` is not a strict failure. It is at times even expected behavior. As such, we should exclude it from calling `onFailure` which is a more serious error reporting mechanism than just logging the error at level `error`.

### DIFF
--- a/packages/integration-sdk-runtime/src/logger/registerEventHandlers.ts
+++ b/packages/integration-sdk-runtime/src/logger/registerEventHandlers.ts
@@ -108,7 +108,10 @@ function integrationLoggerEventHandlerCallback(
   return (err, event) => {
     const logger = getErrorLogger();
     logger.error({ err, event });
-    if (logger.onFailure) {
+    // multipleResolves is excluded from `onFailure` as it is
+    // not a strict failure and is at times even expected behavior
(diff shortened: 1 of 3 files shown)


packages/integration-sdk-runtime/src/logger/registerEventHandlers.ts
```diff
@@ -108,7 +108,10 @@ function integrationLoggerEventHandlerCallback(
   return (err, event) => {
     const logger = getErrorLogger();
     logger.error({ err, event });
-    if (logger.onFailure) {
+    // multipleResolves is excluded from `onFailure` as it is
+    // not a strict failure and is at times even expected behavior
+    // in node-fetch, Promise.all, and Promise.race
+    if (logger.onFailure && event !== 'multipleResolves') {
```
austinkelleher, line 114: It would be awesome if we could add a unit test for this!

austinkelleher: Should we major version the SDK after this?

austinkelleher (dismissed):

austinkelleher (approved):

PR #798, opened 2022-09-21
INT-5556 Add no-console lint rule

edreyer1:


packages/integration-sdk-dev-tools/config/eslint.json
```diff
@@ -47,6 +47,7 @@
   ],
   "rules": {
     "no-unused-vars": "off",
+    "no-console": "error",
```
ndowmon, line 50: Have you confirmed (perhaps by linking `@jupiterone/integration-sdk-dev-tools` in a local version of `graph-aws`) that the addition of this rule causes `yarn lint` to fail if you add a `console.log()` statement somewhere in `graph-aws`?

ndowmon (approved):

packages/integration-sdk-dev-tools/config/eslint.json
edreyer1: Screen Shot 2022-09-21 at 10 21 54 AM

edreyer1: Just checked  and yes it does seem to catch console.log

austinkelleher (approved):
This is awesome. Should we technically do a major version bump for this?

PR #803, opened 2022-09-28
Increase jest default timeout to 20 seconds

i5o:
As test suites grow bigger we are in the need of increasing the default timeout of Jest, this prevents setting `jest.setTimeout` on each tests that times out.

### DIFF
--- a/packages/integration-sdk-dev-tools/config/jest.js
+++ b/packages/integration-sdk-dev-tools/config/jest.js
@@ -18,4 +18,5 @@ module.exports = {
   setupFilesAfterEnv: [
     '<rootDir>/node_modules/@jupiterone/integration-sdk-dev-tools/config/setupJestTestFramework.js',
   ],
+  testTimeout: 20000,
 };


packages/integration-sdk-dev-tools/config/jest.js
```diff
@@ -18,4 +18,5 @@ module.exports = {
   setupFilesAfterEnv: [
     '<rootDir>/node_modules/@jupiterone/integration-sdk-dev-tools/config/setupJestTestFramework.js',
   ],
+  testTimeout: 20000,
```
ndowmon, line 21: If this is a limit we're bumping up against in the `graph-aws` project specifically, what if we just override this there? I would definitely prefer to keep the base config as light as possible.

Consider doing something like the following in the target project's `jest.config.js` file:

```diff
- module.exports = require('@jupiterone/integration-sdk-dev-tools/config/jest');
+ module.exports = {
+   ...require('@jupiterone/integration-sdk-dev-tools/config/jest'),
+   testTimeout: 20000,
+ };```

PR #804, opened 2022-10-05
Add no misused promises

Nick-NCSU:


packages/integration-sdk-dev-tools/config/eslint.json
```diff
@@ -30,7 +30,13 @@
         "@typescript-eslint/explicit-module-boundary-types": "off",
         "@typescript-eslint/await-thenable": "error",
         "@typescript-eslint/require-await": "error",
-        "@typescript-eslint/no-floating-promises": "error"
+        "@typescript-eslint/no-floating-promises": "error",
+        "@typescript-eslint/no-misused-promises": [
+          "error",
+          {
+            "checksVoidReturn": false
```
Nick-NCSU, line 37: Would like opinions on this rule, ideally we would use 
```suggestion
            "checksVoidReturn": {
              "arguments": false
            }
```
but this option does not seem to work for me.
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-misused-promises.md#checksvoidreturn

packages/integration-sdk-cli/src/neo4j/uploadToNeo4j.ts
```diff
@@ -27,7 +27,7 @@ export async function uploadToNeo4j({
       'ERROR: must provide login information in function call or include NEO4J_URI, NEO4J_USER, and NEO4J_PASSWORD files in your .env file!',
     );
   }
-  if (!isDirectoryPresent(pathToData)) {
+  if (!(await isDirectoryPresent(pathToData))) {
```
ndowmon, line 30: 😳

packages/integration-sdk-dev-tools/config/eslint.json
ndowmon: Can you give an example of why we would want to set `checksVoidReturn` and `arguments` to `false`? I'm not fully understanding the docs for this rule, and am trying to understand the difference between using these arguments instead of the base case:

```json
"@typescript-eslint/no-misused-promises": "error",
```

Nick-NCSU: Now that I am looking again at this, it may just be a bigger issue with how we use the NodeJS EventEmitter...

Example: https://github.com/JupiterOne/sdk/blob/cad4c3bd50fcac73fa90098578e27121bd5c79d0/packages/integration-sdk-cli/src/commands/run.ts#L76

Here we have an `EventEmitter.on()` which expects a function that returns `void` but since `eventPublishingQueue.enqueue(event)` returns `Promise<void>` the ESLint rule catches it. My original thinking was that allowing arguments to be Promises would be _ok_, but now that I look at it, this promise would still never be caught. 

Even if we change the line to
`.on('event', async (event) => await eventPublishingQueue.enqueue(event))`
we would still be returning a `Promise<void>` in this synchronous function.

PR #806, opened 2022-10-07
Updated ESLint No-misused-promises to error always

Nick-NCSU:


ndowmon (approved):

packages/integration-sdk-runtime/src/synchronization/events.ts
```diff
@@ -8,7 +8,7 @@ import {
 } from '../synchronization';
 
 type EventPublishingQueue = {
-  enqueue: (event: IntegrationEvent) => Promise<void>;
+  enqueue: (event: IntegrationEvent) => void;
```
austinkelleher, line 11: I'm a little confused about this change. It looks like the usage below is, in fact, async. Can you explain?

Nick-NCSU: I believe even though enqueue previously returned a Promise<void> we were not using async/await or promises at all. An example (and where TypeScript ESLint threw the error) can be found here: https://github.com/JupiterOne/sdk/blob/main/packages/integration-sdk-runtime/src/synchronization/index.ts#L87

NodeJS Event Emitters always run synchronously so passing in a promise will not be awaited at all. The way we should be using `PQueue.enqueue()` is not by awaiting it but instead adding the events and awaiting `PQueue.onIdle()`, which you can see we do [here](https://github.com/JupiterOne/sdk/blob/27dc50b6a98359db6f280da71d213aeadd9c5abc/packages/integration-sdk-runtime/src/synchronization/index.ts#L117)

austinkelleher: I’m still confused by this and I’m worried that it could cause bad usage. I’d rather see the eslint rule suppressed in places where we aren’t awaiting on the promise (e.g. event emitter handlers) instead of completely changing the TypeScript typing.

The usage of enqueue clearly returns a Promise here and now the TypeScript typing states that it does not.

PR #807, opened 2022-10-12
Release 8.26.0

zemberdotnet:


austinkelleher (approved):
Do you see this as a non-breaking change @zemberdotnet?

zemberdotnet:
Shouldn't really be breaking. `multipleResolves` isn't a strict failure the way `uncaughtExceptions`/`unhandledPromiseRejection` is.

PR #808, opened 2022-10-16
INT-5868 - Allow skipping collection of `disk-usage` metric

austinkelleher:
We have a theory that the `disk-usage` metric may be causing the Node.js process to hang in some situations. You can now disable this metric from emitted completely by specifying the `DISABLE_DISK_USAGE_METRIC` environment variable with any value.

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/executeIntegration.ts
+++ b/packages/integration-sdk-runtime/src/execution/executeIntegration.ts
@@ -164,6 +164,10 @@ export async function executeWithContext<
 ): Promise<ExecuteIntegrationResult> {
   const { logger } = context;
 
+  // NOTE (austinkelleher): There is a theory that there may be some issue
+  // with the disk publishing metric code causing the process to hang.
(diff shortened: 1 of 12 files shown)


packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -164,35 +164,26 @@ export async function executeWithContext<
...
-    }
-
-    const stepStartStates: StepStartStates =
-      (await config.getStepStartStates?.(context)) ??
-      getDefaultStepStartStates(config.integrationSteps);
-
-    validateStepStartStates(config.integrationSteps, stepStartStates);
-
-    diskUsagePublishInterval = setInterval(() => {
+  function createDiskUsagePublishInterval() {
```
austinkelleher, line 185: Turned this into a function

zemberdotnet (dismissed):

zemberdotnet (approved):

PR #811, opened 2022-10-18
INT-5717 - Publish custom graph data collection metrics

austinkelleher:
After graph data has been successfully added to the `JobState`, a custom metric is emitted that allows us to track the count of each graph object by `_type`.

The following new custom metrics have been added:

- `collected_entities`
- `collected_relationships`
- `collected_mapped_relationships`

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/executeIntegration.ts
+++ b/packages/integration-sdk-runtime/src/execution/executeIntegration.ts
@@ -1,4 +1,7 @@
 import {
+  AfterAddEntityHookFunction,
+  AfterAddRelationshipHookFunction,
+  Entity,
   ExecutionContext,
   ExecutionHistory,
   IntegrationInstance,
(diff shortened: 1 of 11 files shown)


packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -243,6 +252,8 @@ export async function executeWithContext<
       createStepGraphObjectDataUploader,
       beforeAddEntity: config.beforeAddEntity,
       beforeAddRelationship: config.beforeAddRelationship,
+      afterAddEntity: createAfterAddEntityInternalHook(logger),
```
mdaum, line 255: Do we think we may ever want to allow user to override these hooks? Could be added in future change, but we could export the default hook creation function and allow people to slap extra logic onto their hook beyond metric publishing.

mdaum (approved):
looks good! one "thinking aloud" question but everything looks good to me

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
austinkelleher: Right. My thinking is the same. We can easily add in the future, but I don't want to add a preemptively add an unused feature that may never be used.

PR #814, opened 2022-11-01
Remove asynchronous hasKey

ndowmon:
The `hasKey` function has no asynchronous implementation, and awaiting `jobState.hasKey` can unexpectedly cause duplicate key errors in some specific edge cases (see https://github.com/JupiterOne/sdk/pull/694/files).

I believe this is a major version change, because it changes a public interface. Adopting projects should remove `await` anywhere this method is called.


adam-in-ict (approved):
:100:  I've found having both to be confusing.  Appreciate the update!

PR #815, opened 2022-11-03
Allow duplicate types in document command

Nick-NCSU:
### Changed

- Added option to allow duplicate types in `getSortedJupiterOneTypes`
- Changed `j1-integration document` command to allow duplicate types

PR #816, opened 2022-11-04
Update executeStepWithDependencies to allow dependencyGraphId

Nick-NCSU:
I have had this change partially implemented for a while and was reminded of it when implementing #815 

Should keep current functionality and only changes how the function works if you define `dependencyStepIds`

### DIFF
--- a/packages/integration-sdk-testing/src/getStepExecutionOrder.ts
+++ b/packages/integration-sdk-testing/src/getStepExecutionOrder.ts
@@ -1,0 +1,40 @@
+import { Step } from '@jupiterone/integration-sdk-core';
+import { buildStepDependencyGraph } from '@jupiterone/integration-sdk-runtime';
+import {
+  DEFAULT_DEPENDENCY_GRAPH_IDENTIFIER,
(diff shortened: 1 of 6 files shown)


packages/integration-sdk-testing/src/getStepExecutionOrder.ts
```diff
@@ -0,0 +1,40 @@
...
+import {
+  DEFAULT_DEPENDENCY_GRAPH_IDENTIFIER,
+  seperateStepsByDependencyGraph,
+} from '@jupiterone/integration-sdk-runtime/src/execution/utils/seperateStepsByDependencyGraph';
+
+export function getStepExecutionOrder(params: {
+  integrationSteps: Step<any>[];
+  dependencyStepIds: string[];
+  dependencyGraphOrder: string[];
+}): Set<string> {
```
Nick-NCSU, line 12: ES6 Sets maintain insertion order: https://262.ecma-international.org/6.0/#sec-set.prototype.foreach

ndowmon (approved):

PR #827, opened 2022-11-30
NO-TICKET - Fix cache paths on Windows

austinkelleher:
Also remove an unused file.


packages/integration-sdk-runtime/src/execution/collection.ts
```diff
@@ -1,19 +0,0 @@
-export type GraphObjectTypeSummary = {
```
austinkelleher, line 1: File unused

zemberdotnet (approved):

zemberdotnet (approved):

PR #841, opened 2023-02-20
Add tags with value: '' to .tags property

ndowmon:
The AWS integration (and, I suspect, other integrations), upload allow tags to have empty values (`''`). In this case, JupiterOne's `assignTags` functionality basically skips these tags altogether, preventing any useful querying capabilities for these types of tags.

Screenshot 2023-02-20 at 1 39 06 PM

This PR checks for tags with empty string values (`''`) and assigns them to the `.tags` array. 

In the future, I think we should consider adding _every_ tag name to the `.tags` property, instead of selectively choosing just those that match `/^true$/i`.


austinkelleher (approved):
Thanks for the fix

PR #846, opened 2023-03-07
fix cli tests when user sets env vars

zemberdotnet:
# Description

This PR fixes failing tests when the user has set env vars like `JUPITERONE_API_KEY` and `JUPITERONE_ACCOUNT`.


packages/cli/src/__tests__/cli-import.test.ts
```diff
@@ -237,6 +237,9 @@ test('should exclude relationships when specified', async () => {
 });
 
 test('should throw error when missing api key', async () => {
+  // if a developer has these set it could interfere with the test
```
VDubber, line 240: nit: wrap in a method: cleanEnvironment()

VDubber (approved):

PR #852, opened 2023-03-22
INT-7550 - fix duplicate integration data model doc rows

austinkelleher:
This change updates the automatic integration documentation generator strategy to avoid creating duplicate records for entities, relationships, and mapped relationships. There was a change that was introduced recently that allowed for duplicate records. In some integrations, relationships and mapped relationships have the same `_type` value and a different value for one/more of the `_class`, `sourceType`, `targetType`, and `direction` properties. Originally, the de-duplication code in the `j1-integration document` command used the `_type` value for entities and relationships as the de-dup key. The `_type` is insufficient, but we still must ensure that other important properties are considered in the de-dup process instead of blindly allowing duplicates.

The new approach removes allowing duplicates for entities entirely and uses additional properties for de-duplicating relationship rows.

### DIFF
--- a/packages/integration-sdk-cli/src/utils/getSortedJupiterOneTypes.test.ts
+++ b/packages/integration-sdk-cli/src/utils/getSortedJupiterOneTypes.test.ts
@@ -7,7 +7,7 @@ import {
   StepRelationshipMetadata,
 } from '@jupiterone/integration-sdk-core';
 import { collectGraphObjectMetadataFromSteps } from './getSortedJupiterOneTypes';
-import { randomUUID } from 'crypto';
+import { v4 as uuid } from 'uuid';
 
 function createIntegrationStep({
   entities = [],
@@ -18,7 +18,7 @@ function createIntegrationStep({
   relationships?: StepRelationshipMetadata[];
   mappedRelationships?: StepMappedRelationshipMetadata[];
 }): Step<any> {
-  const stepId = randomUUID();
+  const stepId = uuid();
 
   return {
     id: `step-${stepId}`,
--- a/packages/integration-sdk-cli/src/utils/getSortedJupiterOneTypes.test.ts
+++ b/packages/integration-sdk-cli/src/utils/getSortedJupiterOneTypes.test.ts
@@ -2,197 +2,215 @@ import {
   RelationshipClass,
   RelationshipDirection,
   Step,
+  StepEntityMetadata,
+  StepMappedRelationshipMetadata,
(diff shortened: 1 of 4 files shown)


packages/integration-sdk-cli/src/utils/getSortedJupiterOneTypes.ts
```diff
@@ -108,7 +103,7 @@ export function collectGraphObjectMetadataFromSteps(
     >;
 
     for (const e of step.entities) {
-      if (!duplicateTypes && entityTypeSet.has(e._type)) {
```
austinkelleher, line 111: This was an invalid case. Entities should never be duplicated.

packages/integration-sdk-cli/src/utils/getSortedJupiterOneTypes.ts
```diff
@@ -141,3 +140,11 @@ export function collectGraphObjectMetadataFromSteps(
     mappedRelationships,
   };
 }
+
+function toRelationshipSetValue(r: StepRelationshipMetadata) {
```
austinkelleher, line 144: The main updates can be summarized in these two functions. Previously, `_type` was being used to de-dup doc rows.

packages/integration-sdk-cli/src/utils/getSortedJupiterOneTypes.test.ts
```diff
@@ -2,197 +2,215 @@ import {
...
+function createIntegrationStep({
+  entities = [],
+  relationships = [],
+  mappedRelationships = [],
+}: {
+  entities?: StepEntityMetadata[];
+  relationships?: StepRelationshipMetadata[];
+  mappedRelationships?: StepMappedRelationshipMetadata[];
+}): Step<any> {
+  const stepId = randomUUID();
```
zemberdotnet, line 21: FYI if we do this we need to add an engine field in the `package.json`. I believe this was added in 14.19. Correct me if I'm wrong.

austinkelleher: Good comms. Going to use `uuid` and we can make the breaking change in a separate PR.

ndowmon (approved):
Thanks for the change. Looks good. 

Did you get a chance to run this against `graph-aws` or `graph-azure` to verify the changes you expect?

austinkelleher:
@ndowmon Yes. Ran against both.

PR #854, opened 2023-03-24
INT-7594 - switch `uuid` module to native `crypto` `randomUUID`

austinkelleher:
The native `crypto` `randomUUID` module was introduced in Node.js 14.17.0. The `engines` field in each package has been updated to reflect the new Node.js version requirement. This is technically a breaking change. Additional changes are necessary before we can update the `engines` field to support new versions of Node.js.


yarn.lock
```diff
@@ -9007,16 +9007,6 @@ uuid@^3.3.2:
   resolved "https://registry.yarnpkg.com/uuid/-/uuid-3.4.0.tgz#b23e4358afa8a202fe7a100af1f5f883f02007ee"
   integrity sha512-HjSDRw6gZE5JMggctHBcjVak08+KEVhSIiDzFnT9S9aegmp85S/bReBVTb4QTFaRNptJ9kuYaNhnbNEOkbKb/A==
 
-uuid@^7.0.3:
```
austinkelleher, line 9010: Less is more.

zemberdotnet (approved):
☘️

PR #859, opened 2023-03-30
extend globby to exclude decleration files

zemberdotnet:
# Description

Exclude delceration files from the search for TypeScript files. Just because deleartion files exists does not mean the integration should run through ts-node.

The distributed version of graph projects is js with decleration files. Registering ts-node is slow, so when there isn't any benefit, we shouldn't do it.


packages/integration-sdk-cli/src/config.ts
```diff
@@ -61,7 +61,7 @@ async function isTypescriptPresent(
   // Refs:
   // - https://github.com/mrmlnc/fast-glob#pattern-syntax
   // - https://github.com/micromatch/micromatch#backslashes
-  const paths = await globby('**/*.ts', {
+  const paths = await globby(['**/*.ts', '!**/*.d.ts'], {
```
austinkelleher, line 64: Can you explain what motivated the change exactly?

austinkelleher (approved):

packages/integration-sdk-cli/src/config.ts
zemberdotnet: I sometimes want to run the debugger on the built version of graph projects. However, since the graph projects ship deceleration files, they always run with ts-node. It's interfered with the debugger. Additionally, using the CLI on built versions of projects shouldn't need to take the extra time and overhead of registering ts-node.

PR #865, opened 2023-04-03
[INT-7559] - add ingestionConfig to the invocationConfig and new utils to be used in graph repos

dobregons:
## Summary
- Add a new field `ingestionConfig` to the `invocationConfig` in  `@jupiterone/integration-sdk-core`. This will be
used to store information about data ingestion sources that can be enabled or disabled. More info in the `README.md` and the unit tests.
- ~~Add new helper function `createIntegrationIngestionConfig` in `@jupiterone/integration-sdk-core`. This function will be used in the graph repos to produce the `ingestionConfig`.~~


packages/integration-sdk-core/src/data/createIntegrationIngestionConfig.ts
```diff
@@ -0,0 +1,47 @@
...
+ * The childIngestionSources will be the list of stepIds that have any dependencies
+ * on the steps that match the ingestion sources specified.
+ *
+ * @export
+ * @template TStepExecutionContext
+ * @param {IntegrationIngestionConfigData} ingestionConfigData ingestionData without childIngestionSources
+ * @param {Step<TStepExecutionContext>[]} integrationSteps total list of integration steps
+ * @return {*}  {IntegrationIngestionConfigFieldMap} ingestionData with childIngestionSources
+ */
+export function createIntegrationIngestionConfig<
```
VDubber, line 20: This looks good. 
What are your thoughts on how this method is called? I'm hoping that the SDK will provide a single exported method that's referencing a wrapper method of this with its args already provided. Something like `getIntegrationIngestionConfig()`

docs/integrations/development.md
```diff
@@ -448,6 +449,62 @@ export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> =
...
+};
+
+export const ingestionConfig: IntegrationIngestionConfigFieldMap =
+  createIntegrationIngestionConfig(ingestionConfigData, integrationSteps);
+
+export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> =
+  {
+    instanceConfigFields: {},
+    integrationSteps,
+    ingestionConfig,
```
VDubber, line 504: ingestionConfig is not needed during invocation. What will be needed during invocation will be stored inside the existing config that is saved from the UI. That context.config is already being passed to the getStepStartStates(context) method.

```
export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> =
  {
    instanceConfigFields,
    getStepStartStates,
    integrationSteps,
  };

export {
  validateInvocation,
  actionExecutionHandler,
  getIntegrationIngestionConfig, // This is the additional export we need to have.
};
```

dobregons:
### Update
- I created `generateIntegrationDataSourcesCommand` that will be called when we bundle an integration in `integration-deployments` so it'll generate a file called `integrationDataSources.json` in the path `src/${integration}/deploy/terraform/data/integrationDataSources.json` in that repo. This file will have all the ingestion sources data based on the `ingestionConfig` provided by the graph in the `invocationConfig`.

packages/integration-sdk-cli/src/config.ts
```diff
@@ -51,6 +51,48 @@ export function loadInvocationConfig(
...
+function loadConfigFromSrc(projectPath: string) {
+  return loadConfig(path.join(projectPath, 'src'));
+}
+
+function loadConfigFromDist(projectPath: string) {
+  return loadConfig(path.join(projectPath, 'dist'));
+}
+
+/**
+ * The way that integration npm packages are distributed has changed over time.
```
VDubber, line 63: 😎

packages/integration-sdk-cli/src/commands/generate-integration-data-sources.test.ts
```diff
@@ -0,0 +1,182 @@
...
+            _type: 'github_issue',
+            _class: ['Issue'],
+          },
+        ],
+        relationships: [],
+        dependsOn: ['fetch-repos', 'fetch-users', 'fetch-collaborators'],
+        executionHandler: jest.fn(),
+      },
+      {
+        id: 'fetch-teams',
```
VDubber, line 109: Since `fetch-issues` and `fetch-teams` do not have `ingestionSourceId` they will not show up in the ingestionConfig? Is that correct?
I wonder if we want them to show up in the ingestionConfig so that they show up in the UI but as being "required". Looking at Michelle's mockups, that is the case.

docs/integrations/development.md
```diff
@@ -448,6 +449,56 @@ export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> =
...
+    executionHandler: /* Step Execution Handler */,
+  },
+];
+
+const ingestionConfig: IntegrationIngestionConfigFieldMap = {
+  [INGESTION_SOURCE_IDS.FINDING_ALERTS]: {
+    title: 'Finding Alerts',
+    description:
+      'Dependabot vulnerability alert ingestion and Code scanning alerts',
+    defaultsToDisabled: true,
```
VDubber, line 489: add cannotBeDisabled: true

VDubber (dismissed):
Sweet

VDubber (approved):

PR #866, opened 2023-04-03
Release 8.33.0

zemberdotnet:
Release 8.33.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,8 +33,7 @@
     "lint-staged": "^10.2.6",
     "prettier": "^2.2.1",
     "ts-jest": "^27.1.0",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -33,8 +33,7 @@
     "lint-staged": "^10.2.6",
     "prettier": "^2.2.1",
     "ts-jest": "^27.1.0",
-    "typescript": "^4.2.4",
-    "@jupiterone/integration-sdk-dev-tools": "^8.11.0"
```
zemberdotnet, line 37: I'm also removing this root dependency here. It's not needed afaik and will cause versioning issues.

PR #869, opened 2023-04-07
[NO-TICKET] - Converting schema validation errors to warnings

adam-in-ict:
It's been a persistent issue that schema validation checks can be more of a hindrance than a help in integration development.  The information is valuable in helping maintain compliance with the data model, but it sometimes forces developers into the unpleasant decision between forcing in hardcoded values or using the `--disable-schema-validation` flag in instances where some required data simply isn't available.

### DIFF
--- a/packages/integration-sdk-core/src/data/createIntegrationEntity.ts
+++ b/packages/integration-sdk-core/src/data/createIntegrationEntity.ts
@@ -103,7 +103,11 @@ export function createIntegrationEntity(
   validateRawData(generatedEntity);
 
   if (process.env.ENABLE_GRAPH_OBJECT_SCHEMA_VALIDATION) {
-    validateEntityWithSchema(generatedEntity);
+    try {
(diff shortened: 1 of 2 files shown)


VDubber (commented):
Any documentation that we should update? 
The schema validation seems to be tribal knowledge.

packages/integration-sdk-core/src/data/createIntegrationEntity.ts
```diff
@@ -103,7 +103,12 @@ export function createIntegrationEntity(
   validateRawData(generatedEntity);
 
   if (process.env.ENABLE_GRAPH_OBJECT_SCHEMA_VALIDATION) {
```
VDubber, line 105: I wonder if we should remove support for the env var.

VDubber: Meaning, during development we should always being logging the warnings but still not in production.

PR #871, opened 2023-04-07
Add suppport for string[] environment variables

ndowmon:
### Added

- Added support for `string[]` environment variables. Example:

`.env` file:

```
// .env
SEVERITIES='HIGH,CRITICAL'
```

Code: 
```ts
import { Client } from './client';
import { createVulnEntity } from './converter';

type IntegrationConfig = { apiKey: string, severities: string[] };
const invocationConfig: IntegrationInvocationConfig = {
  instanceConfigFields: [{
    apiKey: { type: 'string', mask: true },
    severities: { type: 'string[]' },
  }],
  integrationSteps: [{
    id: 'fetch-vulnerabilities',
    name: 'Fetch Vulnerabilities',
    entities: [{ resourceName: 'Vuln', _type: 'vuln', _class: 'Finding' }],
    relationships: [],
    executionHandler: ({ instance }) => {
      const { apiKey, severities } = instance.config;
      const client = new Client({ apiKey });

      await iterateVulnerabilitiesForSeverity(
        { severities },
        async (vuln) => {
          await jobState.addEntity(createVulnEntity(vuln));
        }
      )
    }
  }]
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,48 @@ and this project adheres to
 
 ## Unreleased
 
+### Added
+
+- Added support for `string[]` environment variables. Example:
+
+`.env` file:
+
+```
+// .env
+SEVERITIES='HIGH,CRITICAL'
+```
+
+Code: 
+```ts
+import { Client } from './client';
+import { createVulnEntity } from './converter';
+
+type IntegrationConfig = { apiKey: string, severities: string[] };
+const invocationConfig: IntegrationInvocationConfig = {
+  instanceConfigFields: [{
+    apiKey: { type: 'string', mask: true },
+    severities: { type: 'string[]' },
+  }],
+  integrationSteps: [{
+    id: 'fetch-vulnerabilities',
+    name: 'Fetch Vulnerabilities',
+    entities: [{ resourceName: 'Vuln', _type: 'vuln', _class: 'Finding' }],
+    relationships: [],
+    executionHandler: ({ instance }) => {
+      const { apiKey, severities } = instance.config;
+      const client = new Client({ apiKey });
+
+      await iterateVulnerabilitiesForSeverity(
+        { severities },
+        async (vuln) => {
+          await jobState.addEntity(createVulnEntity(vuln));
+        }
+      )
+    }
+  }]
+}
+```
+
 ## 8.33.1 - 2023-04-03
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


packages/integration-sdk-runtime/src/execution/config.ts
```diff
@@ -70,6 +70,13 @@ function convertEnvironmentVariableValueForField(
       }
       break;
     }
+    case 'string[]': {
+      convertedValue = environmentVariableValue
+        ?.split(',')
+        .map((v) => v.trim())
+        .filter((v) => v.length > 0) || []
```
ndowmon, line 77: I drew inspiration from the exact logic we use in the frontend: https://github.com/JupiterOne/web-integrations/blob/main/src/pages/InstanceManagement/Template/IntegrationInstanceForm/useIntegrationInstanceFormLogic.ts#L69

packages/integration-sdk-runtime/src/execution/__tests__/config.test.ts
```diff
@@ -13,32 +13,38 @@ jest.mock('fs');
...
+    Record<'stringVariable' | 'booleanVariable' | 'stringArrayVariable', IntegrationInstanceConfigField>
   > = {
     stringVariable: {
       type: 'string',
     },
     booleanVariable: {
       type: 'boolean',
     },
+    stringArrayVariable: {
+      type: 'string[]',
```
ndowmon, line 38: I also considered `array`, which is consistent with what's in `web-frontend`. But wonder if in the future we might want to load arrays of numbers, etc as well.

CHANGELOG.md
```diff
@@ -9,6 +9,48 @@ and this project adheres to
...
 
+### Added
+
+- Added support for `string[]` environment variables. Example:
+
+`.env` file:
+
+```
+// .env
+SEVERITIES='HIGH,CRITICAL'
```
zemberdotnet, line 20: I know we've done it for a while, but I think we should consider supporting a valid JSON array instead:
.env:
```
SEVERITIES=["HIGH", "CRITICAL"]
```
It would just involve calling JSON.parse on the loaded string and then validating they are all strings. Wdyt?

ndowmon: makes sense - we could expand this type to just be `json` and load whatever we want

packages/integration-sdk-runtime/src/execution/config.ts
```diff
@@ -70,6 +73,10 @@ function convertEnvironmentVariableValueForField(
       }
       break;
     }
+    case 'json': {
+      convertedValue = JSON.parse(environmentVariableValue);
```
zemberdotnet, line 77: If we fail to parse the value, it will probably be a good idea to throw an error saying that value `foo` is not valid JSON rather than just bubbling up a generic JSON.parse error.

zemberdotnet (approved):

austinkelleher (approved):

PR #876, opened 2023-04-17
Introduce new event names to consolidate to names provided by SDK

VDubber:
Introduce warn_incomplete_data, info event names. Removed the deprecated publishEvent method to restrict names to only those provided by the SDK.


packages/integration-sdk-core/src/types/logger.ts
zemberdotnet, line 1: Are accompanying changes in the runtime not needed?

ndowmon, line 1: One of the original motivations for `publishInfoEvent`/`publishWarnEvent`/`publishErrorEvent` was to actually stop us from doing string matching for **COMPETED_WITH_ERRORS**, and start adding a property called `level` to the event records in `integration-events` table. See https://github.com/JupiterOne/sdk/pull/558

Is that something we can/should be doing as part of this effort?

VDubber: After discussion with Nick, it seems we need to get by-in to remove publishEvent().
The other EventNames will be added for future use.

JakeFerrero (approved):

zemberdotnet (approved):

PR #877, opened 2023-04-18
Auto-generated Entity Types

maxheckel:
Creates auto-generation of types based on the json schema from the data model.  Introduces a new package `integration-sdk-entities` to contain these types so we don't run into any conflicting names for common entity names (i.e. `Entity`)

These will eventually be imported into the `integration-sdk-core` to allow `graph` libraries to use these instead of raw javascript objects when building integrations, hopefully significantly reducing the developer cost to know and validate that their code is working.


packages/integration-sdk-entities/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@jupiterone/integration-sdk-entities",
+  "version": "8.35.0",
+  "description": "The SDK for developing JupiterOne integrations",
```
zemberdotnet, line 4:
```suggestion
  "description": "Generated types for the JupiterOne data-model",
```

packages/integration-sdk-entities/README.md
```diff
@@ -0,0 +1,32 @@
+# @jupiterone/integration-sdk-entities
+
+This package contains all entities generated from J1's data model to be used
+throughout J1 systems.
```
zemberdotnet, line 4: I'm okay with it being entities specific, but the `data-model` technically defines the schema for all graph objects including relationships.

packages/integration-sdk-entities/tools/copy-schemas/_schemas/.gitkeep
zemberdotnet, line 1: I've actually never used `.gitkeep` before. Is this just here so the directory is created?

packages/integration-sdk-entities/tools/create-barrel/index.js
```diff
@@ -0,0 +1,44 @@
...
+    const filePath = path.join(dir, file);
+    const typeBlob = await fs.readFile(filePath);
+    const typeString = typeBlob.toString();
+    const empty = `/* eslint-disable */
+`;
+    if (typeString !== empty) {
+      fileNames.push(file);
+    }
+  }
+
```
zemberdotnet, line 42: Adding a comment on what this file does may be helpful. It seems pretty specific to the way we are processing things, but I can't quite make out what the purpose is.

zemberdotnet (approved):

packages/integration-sdk-entities/README.md
maxheckel: Made it more explicit

zemberdotnet (approved):

PR #879, opened 2023-04-19
Bump Node.js to v18

austinkelleher:
It is what it says on the box.


packages/integration-sdk-runtime/src/logger/registerEventHandlers.ts
```diff
@@ -71,13 +53,10 @@ export function registerEventHandlers(
 ): RegisteredEventListeners {
   const multipleResolveListener = createMultipleResolveListener(callback);
   process.on('multipleResolves', multipleResolveListener);
-  const unhandledRejectionListener = createUnhandledRejectionListener(callback);
-  process.on('unhandledRejection', unhandledRejectionListener);
```
austinkelleher, line 75: The behavior of `unhandledRejection` changed in Node.js 15. I talked with @ndowmon, who implemented this original event handler, and he said that we can remove it.

See: https://developer.ibm.com/blogs/nodejs-15-release-blog/

VDubber (approved):
LGTM
What are next steps for testing?

packages/integration-sdk-runtime/src/logger/registerEventHandlers.ts
zemberdotnet: Wdyt about removing multipleResolvesListener as well. It's deprecated.

austinkelleher: @ndowmon What do you think?

ndowmon: go for it.

zemberdotnet (approved):

PR #881, opened 2023-04-19
NO-TICKET - Improve performance of SDK build

austinkelleher:
Introduce yarn caching and leverage `--frozen-lockfile` when installing packages.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,17 +11,18 @@ jobs:
         os: [ubuntu-latest, windows-latest]
 
     steps:
+      - name: Check out code repository source code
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -11,17 +11,18 @@ jobs:
         os: [ubuntu-latest, windows-latest]
 
     steps:
+      - name: Check out code repository source code
+        uses: actions/checkout@v2
```
austinkelleher, line 15: We've found in the past that`actions/checkout` needs to be before `actions/setup-node` to leverage caching.

zemberdotnet (approved):

PR #882, opened 2023-04-20
INT-7803 Fallback to syncJob.integrationJobId in logger if needed

JakeFerrero:
When we create and return the synchronizationContext during initialization, in the logger, set the integrationJobId to the syncJob's integrationJobId if a job id was not passed in in the jobConfiguration.

### DIFF
--- a/packages/integration-sdk-runtime/src/synchronization/index.ts
+++ b/packages/integration-sdk-runtime/src/synchronization/index.ts
@@ -180,6 +180,8 @@ export async function initiateSynchronization(
     job,
     logger: logger.child({
       ...jobConfiguration,
+      integrationJobId:
+        jobConfiguration.integrationJobId ?? job.integrationJobId,
       synchronizationJobId: job.id,
     }),
     uploadBatchSize,


packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -180,6 +180,8 @@ export async function initiateSynchronization(
     job,
     logger: logger.child({
       ...jobConfiguration,
+      integrationJobId:
+        jobConfiguration.integrationJobId ?? job.integrationJobId,
```
ndowmon, line 184: Will `job.integrationJobId` ever be unset? Is it enough to simply do:

```suggestion
      integrationJobId: job.integrationJobId,
```

ndowmon (approved):

packages/integration-sdk-runtime/src/synchronization/index.ts
JakeFerrero: I'm not sure. In theory if jobConfiguration.integrationJobId is set, then the syncJob's job.integrationJobId _should_ be equal to whatever jobConfiguration.integrationJobId  is. But just in case, I went with this since it will ensure the functionality is the same as it was before, just with the included fallback.

PR #883, opened 2023-04-21
Int 7937 add support for string array of key value pairs for assign tags

chrichts:
Ticket: https://jupiterone.atlassian.net/browse/INT-7937


chrichts:
Feedback from @austinkelleher was that this will break the logic when customers have tags with `:` characters. I did a spot check in Callisto and this is a common pattern, so I will move this logic to the cloudflare integration.

PR #886, opened 2023-04-28
INT-7562 - add steps using the ingestion source id to childIngestionSources

dobregons:
## Summary
Before this PR, I was only adding the child sources to the array `childIngestionSources`, that means, I was not adding the steps that actually points to the ingestionSourceId in that array. With the previous implementation, there was no way to identify those steps in the UI. e.g.

image

Based on that image we were not able to obtain the `AWS Backup service` but `AWS Backup vault`. This PR adds the changes to obtain both in the command `generate-ingestion-sources-config`: 
1. The steps pointing to the `ingestionSourceId` 
2. The child steps or steps that depends on it.

### DIFF
--- a/packages/integration-sdk-cli/src/commands/generate-ingestion-sources-config.test.ts
+++ b/packages/integration-sdk-cli/src/commands/generate-ingestion-sources-config.test.ts
@@ -26,7 +26,7 @@ describe('#generateIngestionSourcesConfig', () => {
     },
   };
 
-  it('should return the ingestionConfig with empty childIngestionSources', () => {
+  it('should return the ingestionConfig with steps using the ingestion source id in the childIngestionSources.', () => {
     const integrationSteps: IntegrationStep<IntegrationInstanceConfig>[] = [
       {
         id: 'fetch-vulnerability-alerts',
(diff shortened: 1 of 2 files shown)


packages/integration-sdk-cli/src/commands/generate-ingestion-sources-config.test.ts
```diff
@@ -131,15 +131,27 @@ describe('#generateIngestionSourcesConfig', () => {
...
-    // New property added
+    // New property added with the right child ingestion sources
     expect(
       ingestionSourcesConfig[INGESTION_SOURCE_IDS.FETCH_REPOS]
         .childIngestionSources,
-    ).toEqual(['fetch-vulnerability-alerts', 'fetch-issues']);
-    // For FINDING_ALERTS the ingestionConfig keep exactly the same
+    ).toEqual(
+      expect.arrayContaining([
+        'fetch-repos',
```
VDubber, line 140: What does `FETCH_REPOS` have a child ingestion source of `fetch-repos`?

VDubber (approved):

PR #887, opened 2023-04-28
INT-7563 - generate step start states for disabled sources

RonaldEAM:


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ and this project adheres to
 - Remove the following Node.js event listeners due to Node.js core deprecation:
   - `unhandledRejection`
   - `multipleResolves`
+- BREAKING: `executeWithContext(context: TExecutionContext, ...)` now requires
+  that `TExecutionContext.instance` is provided.
```
ndowmon, line 20: Is there any way to avoid making this a breaking change? Is it possible to make `instance` an optional argument instead?

PR #894, opened 2023-05-02
INT-7563: change types for disabledSources changes

RonaldEAM:

PR #895, opened 2023-05-02
INT-7563: modify stepStartStates based on disabledSources

RonaldEAM:
Depends on https://github.com/JupiterOne/sdk/pull/894 to be released first

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to
 
 ## Unreleased
 
+- `executeWithContext(context: TExecutionContext, ...)` now accepts a
+  `context.instance` property
+
 ### 9.0.0
 
 ## Changed
--- a/packages/integration-sdk-runtime/src/execution/__tests__/jobState.test.ts
+++ b/packages/integration-sdk-runtime/src/execution/__tests__/jobState.test.ts
@@ -163,21 +163,21 @@ describe('#hasKey', () => {
     vol.reset();
   });
 
-  test('answers false when no entity or relationship added', async () => {
+  test('answers false when no entity or relationship added', () => {
     const jobState = createTestStepJobState();
-    expect(await jobState.hasKey('a')).toBeFalse();
+    expect(jobState.hasKey('a')).toBeFalse();
   });
 
   test('answers true when entity added', async () => {
     const jobState = createTestStepJobState();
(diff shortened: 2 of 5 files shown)


packages/integration-sdk-runtime/src/execution/__tests__/utils/getIngestionSourceStepStartStates.test.ts
```diff
@@ -0,0 +1,83 @@
...
+    }),
+  ];
+
+  const mockConfigStepStartStates: StepStartStates = {
+    'step-1': { disabled: true, disabledReason: DisabledStepReason.PERMISSION },
+    'step-2': { disabled: false },
+    'step-3': { disabled: false },
+  };
+
+  it('keeps a disabled step as disabled with the same reason', () => {
```
dobregons, line 45: This text is a bit confused, I think you are trying to say that it keeps the step configuration from the `getStepStartStates` from the integration, so what do you think of something like:
```
prioritizes the step state from the instance's getStepStartStates
```
Or something like that...

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to
 
 ## Unreleased
 
+- `executeWithContext(context: TExecutionContext, ...)` now accepts a
```
dobregons, line 12: What will be the version?

dobregons (dismissed):
LMK when the other PR is done and I can give you again the approve.

packages/integration-sdk-runtime/src/execution/__tests__/jobState.test.ts
```diff
@@ -163,21 +163,21 @@ describe('#hasKey', () => {
     vol.reset();
   });
 
-  test('answers false when no entity or relationship added', async () => {
+  test('answers false when no entity or relationship added', () => {
     const jobState = createTestStepJobState();
-    expect(await jobState.hasKey('a')).toBeFalse();
+    expect(jobState.hasKey('a')).toBeFalse();
```
VDubber, line 168: This was from another PR, yes?

RonaldEAM: Yes, this was changed by Austin and eslint started to complaint about it so I'm just adding this as additional

VDubber (approved):

PR #904, opened 2023-05-11
Disallow string option from TargetFilterKey

ndowmon:
For safety, we should nudge developers to always supply `targetFilterKeys: string[][]`.


packages/integration-sdk-core/src/types/relationship.ts
```diff
@@ -32,7 +32,7 @@ type RelationshipPropertyValue =
   | null
   | RelationshipMapping;
 
-export type TargetFilterKey = string | string[];
+export type TargetFilterKey = string[];
```
ndowmon, line 35: This is technically a major _type_s change, but not a major _functionality_ change, so I would prefer to release a minor version here rather than a major version.

ndowmon: @softwarewright can you confirm that the following two configurations are equivalent?

```ts
  targetFilterKeys: ['<ROOT>']

  targetFilterKeys: [['<ROOT>']]
```

philidem: @ndowmon `[['<ROOT_ENTITY>']]` is not equivalent to `['<ROOT>']`

VDubber (approved):

PR #912, opened 2023-06-19
[INT-8424] - modify generateIngestionSourcesConfigCommand to include child steps metadata in the childIngestionSources property 

dobregons:
## Summary
Before this PR, the `generateIngestionSourcesConfigCommand` function only returned the stepIds in the `childIngestionSources` property. This implementation was fine, but now we want to show additional data related to the `childIngestionSources` on the frontend. To meet this requirement, I have modified the function to also return the `StepMetadata` (excluding the `executionHandler`).

### DIFF
--- a/packages/integration-sdk-cli/src/commands/generate-ingestion-sources-config.ts
+++ b/packages/integration-sdk-cli/src/commands/generate-ingestion-sources-config.ts
@@ -4,6 +4,7 @@ import {
   IntegrationSourceId,
   Step,
   StepExecutionContext,
+  StepMetadata,
 } from '@jupiterone/integration-sdk-core';
 import { createCommand } from 'commander';
 import { loadConfigFromTarget } from '../config';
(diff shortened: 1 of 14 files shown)


packages/integration-sdk-cli/src/commands/generate-ingestion-sources-config.ts
```diff
@@ -91,18 +92,19 @@ export function generateIngestionSourcesConfig<
...
-            matchedIntegrationStepIds.includes(value),
-          ),
-        )
-        .map(({ id }) => id);
+      // Get the dependent steps for the given matchedIntegrationStepIds
+      const childIngestionSources = integrationSteps.filter((step) =>
+        step.dependsOn?.some((value) =>
+          matchedIntegrationStepIds.includes(value),
+        ),
+      );
```
dobregons, line 100: This is basically the main change.

PR #919, opened 2023-07-04
Release 9.8.0 - add new log entry for each disabled child step

dobregons:
## Summary
This PR introduces a new log entry for each disabled child step to indicate that they have been skipped. To achieve this, a new `DisabledReason` has been implemented exclusively for child steps that are disabled. This `DisabledReason` provides a message like this:
```
Skipped step "{childStepName}". Step was disabled because parent step "{parentStepName}" was disabled. In order to enable this step, please check logs for more information about the parent step
```
when a child step is disabled.

### DIFF
--- a/packages/integration-sdk-core/src/types/logger.ts
+++ b/packages/integration-sdk-core/src/types/logger.ts
@@ -10,10 +10,15 @@ interface ChildLogFunction {
   (options: object): IntegrationLogger;
 }
 
+export interface StepLogAdditionalContext {
+  parentStep?: StepMetadata;
+}
+
 type StepLogFunction = (step: StepMetadata) => void;
 type StepLogFunctionWithReason = (
   step: StepMetadata,
   reason: DisabledStepReason,
+  additionalContext?: StepLogAdditionalContext,
 ) => void;
 type StepLogFunctionWithError = (step: StepMetadata, err: Error) => void;
 type SynchronizationLogFunction = (job: SynchronizationJob) => void;
--- a/packages/integration-sdk-runtime/src/execution/__tests__/dependencyGraph.test.ts
+++ b/packages/integration-sdk-runtime/src/execution/__tests__/dependencyGraph.test.ts
@@ -5,6 +5,7 @@ import { randomUUID as uuid } from 'crypto';
 import waitForExpect from 'wait-for-expect';
 
 import {
+  DisabledStepReason,
   Entity,
   GraphObjectStore,
   IntegrationError,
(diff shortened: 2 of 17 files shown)


packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -281,6 +281,19 @@ export function executeStepDependencyGraph<
                     DisabledStepReason.NONE,
                 );
               skippedStepTracker.add(stepId);
+              // Log child steps disabled
+              const stepDependencies = inputGraph.dependantsOf(stepId);
```
VDubber, line 285: So does this handle the scenario where a different step graph is provided? AWS does not always use the default.

packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -341,6 +346,10 @@ export class IntegrationLogger
         description += `Step was disabled via configuration. Update instance config to enable.`;
         break;
       }
+      case DisabledStepReason.PARENT_DISABLED: {
+        description += `Step was disabled because parent step "${additionalContext?.parentStep?.name}" was disabled. In order to enable this step, please check logs for more information about the parent step.`;
```
VDubber, line 350: 🔥

packages/integration-sdk-core/src/types/logger.ts
```diff
@@ -10,10 +10,15 @@ interface ChildLogFunction {
...
 
+export interface StepLogAdditionalContext {
+  parentStep?: StepMetadata;
+}
+
 type StepLogFunction = (step: StepMetadata) => void;
 type StepLogFunctionWithReason = (
   step: StepMetadata,
   reason: DisabledStepReason,
+  additionalContext?: StepLogAdditionalContext,
```
VDubber, line 21: Instead of adding additionalContext as a new param, could we add an addition type to reason?

DisabledStepReasonWithContext

And that way PARENT_DISABLED could require the additional context.

This would prevent a scenario where the string built does not have the parent's name.

Not hard set on this, thoughts from you?

VDubber:
```
type DisabledParentStepReason {
 reason: DisabledStepReason.DISABLED_PARENT
 parentStepName: string
}

```

packages/integration-sdk-runtime/src/execution/__tests__/dependencyGraph.test.ts
```diff
@@ -1304,4 +1305,153 @@ describe('executeStepDependencyGraph', () => {
...
+    expect(spyA).toHaveBeenCalledBefore(spyC);
+    expect(spyC).toHaveBeenCalledBefore(spyD);
+    expect(spyE).toHaveBeenCalledBefore(spyD);
+
+    // Assert skipStep being called for parent and child steps
+    expect(stepSkipSpy).toHaveBeenCalledTimes(2);
+    // Child step "d" is just skipped once even though it depends on more steps
+    // As "b" is processed first, the "stepSkip" function is invoked for "d" with "b" acting as its parent.
+    // However, for the remaining steps, even though "d" is a child, it has already been logged.
+    expect(stepSkipSpy.mock.calls).toEqual([
```
VDubber, line 1450: This is a great test. thank you

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
dobregons: I confirm you this is taking into account the scenarios where the default graph is not used.  Explanation:
The `inputGraph` that comes as param is created [here](https://github.com/JupiterOne/sdk/blob/main/packages/integration-sdk-runtime/src/execution/step.ts#L86) and the steps that we send to the `buildStepDependencyGraph` function are taking into account the `dependencyGraphId` when it is specified in the integration. That can be seen in [these lines](
https://github.com/JupiterOne/sdk/blob/main/packages/integration-sdk-runtime/src/execution/step.ts#L65-L72)

packages/integration-sdk-core/src/types/logger.ts
dobregons: Since this is an exposed function and other repos are using it, I wanted to implement something to have the less possible impact and I think adding an additional optional param would not affect the existing repos using this function instead of modifying one of the existing params (it could potentially have any impact in the clients). What do you think? @VDubber

VDubber: I see your point. Maybe we should just validate the parent step name and throw if it isn't present.

packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -347,6 +347,10 @@ export class IntegrationLogger
         break;
       }
       case DisabledStepReason.PARENT_DISABLED: {
+        if (!additionalContext?.parentStep?.name) {
+          this.validationFailure(new Error('Parent step not provided'));
```
VDubber, line 351: It seems like the validationFailure function is more for configuration errors.

packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -347,6 +347,10 @@ export class IntegrationLogger
...
+          return;
+        }
         description += `Step was disabled because parent step "${additionalContext?.parentStep?.name}" was disabled. In order to enable this step, please check logs for more information about the parent step.`;
```
VDubber, line 354: let's skip the error and instead, if the parent step name is not included do:
`Step was disabled because parent step was disabled. In order to....
`

VDubber (approved):

VDubber (approved):

PR #920, opened 2023-07-05
Introduce `afterExecution` hook to integration invocation config

austinkelleher:
`afterExecution` is an optional hook function that can be provided. The function is called after an integration has executed, regardless of whether the execution was successful or not. An example of when you may decide to use this function is when you need to close out a globally configured client in an integration.

Example:

```typescript
import { IntegrationInvocationConfig } from '@jupiterone/integration-sdk-core';
import { IntegrationConfig } from './types';

export const invocationConfig: IntegrationInvocationConfig<IntegrationConfig> =
  {
    instanceConfigFields: {},
    integrationSteps: [],
    afterExecution(context) {
      context.logger.info('Integration execution completed...');
    },
  };
```

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/executeIntegration.ts
+++ b/packages/integration-sdk-runtime/src/execution/executeIntegration.ts
@@ -186,10 +186,6 @@ export async function executeWithContext<
     'Starting execution with config...',
   );
 
-  if (shouldPublishDiskUsageMetric) {
-    await tryPublishDiskUsageMetric(context);
-  }
-
   let diskUsagePublishInterval: NodeJS.Timeout | undefined;
   let executionComplete = false;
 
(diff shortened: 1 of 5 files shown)


packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -218,96 +214,116 @@ export async function executeWithContext<
...
+      }
 
-    if (shouldPublishDiskUsageMetric) {
-      await tryPublishDiskUsageMetric(context);
+      if (shouldPublishDiskUsageMetric) {
+        await tryPublishDiskUsageMetric(context);
+      }
+    }
+  } finally {
+    if (config?.afterExecution) {
```
austinkelleher, line 318: NOTE: You should review this with "Ignore whitespace" enabled. There is very little changed in this file. I mainly wrapped this function in a `try/finally`.

Screenshot 2023-07-05 at 2 00 17 PM
e.

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -218,96 +214,116 @@ export async function executeWithContext<
...
-    });
+      context.logger.info(
+        { collectionResult: summary },
+        'Integration data collection has completed.',
+      );
+
+      processDeclaredTypesDiff(summary, (step, undeclaredTypes) => {
+        if (
+          step.status === StepResultStatus.SUCCESS &&
+          undeclaredTypes.length
```
austinkelleher, line 295: Prettier

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -218,96 +214,116 @@ export async function executeWithContext<
   }
 
   try {
-    await removeStorageDirectory();
+    if (shouldPublishDiskUsageMetric) {
+      await tryPublishDiskUsageMetric(context);
+    }
```
zemberdotnet, line 219: Why does this exist when there is just the interval maker above?

zemberdotnet (approved):

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
austinkelleher: Spoke in DM: I think this code can be removed and was likely originally put here to establish a baseline. We discussed that the code can probably be re-organized a bit, but removing this entirely may be the best option.

PR #925, opened 2023-07-11
Batching persister calls by size

Gonzalo-Avalos-Ribas:
Currently we allow integrations to tune the size of the batches of entities/relationships uploaded to the persister api by setting the # of entities/relationships they’d like in each batch. Seeing as we’d like to keep our batched uploads sized  5MB  we should build our batches until they are adequately sized.

We will find that some integrations simply do not produce this much data, in which case we should upload everything for the job in single batch.

In the event that we cannot chunk by size, remove the raw data instead.

This will also prevent some UPLOAD_ERRORS or even UNEXPECTED_SYNCRONIZATION_ERROR since the batching will only allow chunks to be below 5MB.

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/uploader.ts
+++ b/packages/integration-sdk-runtime/src/execution/uploader.ts
@@ -134,6 +134,7 @@ export interface CreatePersisterApiStepGraphObjectDataUploaderParams {
   uploadConcurrency: number;
   uploadBatchSize?: number;
   uploadRelationshipsBatchSize?: number;
+  uploadChunkInMb?: number;
 }
 
 export function createPersisterApiStepGraphObjectDataUploader({
@@ -142,6 +143,7 @@ export function createPersisterApiStepGraphObjectDataUploader({
   uploadConcurrency,
   uploadBatchSize,
   uploadRelationshipsBatchSize,
+  uploadChunkInMb
 }: CreatePersisterApiStepGraphObjectDataUploaderParams) {
   return createQueuedStepGraphObjectDataUploader({
     stepId,
@@ -159,6 +161,7 @@ export function createPersisterApiStepGraphObjectDataUploader({
           graphObjectData,
           uploadBatchSize,
           uploadRelationshipsBatchSize,
+          uploadChunkInMb
         );
       } catch (err) {
         context.logger.error(
--- a/packages/integration-sdk-runtime/src/execution/uploader.ts
+++ b/packages/integration-sdk-runtime/src/execution/uploader.ts
@@ -4,7 +4,6 @@ import { FlushedGraphObjectData } from '../storage/types';
 import {
   uploadGraphObjectData,
   SynchronizationJobContext,
-  BYTES_IN_MB,
 } from '../synchronization';
 import { randomUUID as uuid } from 'crypto';
 import { MAX_BATCH_SIZE } from '../synchronization/shrinkBatchRawData';
@@ -136,7 +135,7 @@ export interface CreatePersisterApiStepGraphObjectDataUploaderParams {
   uploadConcurrency: number;
   uploadBatchSize?: number;
   uploadRelationshipsBatchSize?: number;
-  uploadChunkInMb?: number;
+  uploadChunkInBytes?: number;
 }
 
 export function createPersisterApiStepGraphObjectDataUploader({
@@ -145,10 +144,10 @@ export function createPersisterApiStepGraphObjectDataUploader({
   uploadConcurrency,
   uploadBatchSize,
   uploadRelationshipsBatchSize,
-  uploadChunkInMb,
+  uploadChunkInBytes,
 }: CreatePersisterApiStepGraphObjectDataUploaderParams) {
-  if (uploadChunkInMb && uploadChunkInMb * BYTES_IN_MB > MAX_BATCH_SIZE) {
-    uploadChunkInMb = MAX_BATCH_SIZE;
+  if (uploadChunkInBytes && uploadChunkInBytes > MAX_BATCH_SIZE) {
+    uploadChunkInBytes = MAX_BATCH_SIZE;
   }
   return createQueuedStepGraphObjectDataUploader({
     stepId,
@@ -166,7 +165,7 @@ export function createPersisterApiStepGraphObjectDataUploader({
           graphObjectData,
           uploadBatchSize,
           uploadRelationshipsBatchSize,
-          uploadChunkInMb,
+          uploadChunkInBytes,
         );
       } catch (err) {
         context.logger.error(
--- a/packages/integration-sdk-runtime/src/execution/uploader.ts
+++ b/packages/integration-sdk-runtime/src/execution/uploader.ts
@@ -143,7 +143,7 @@ export function createPersisterApiStepGraphObjectDataUploader({
   uploadConcurrency,
   uploadBatchSize,
   uploadRelationshipsBatchSize,
-  uploadChunkInMb
+  uploadChunkInMb,
 }: CreatePersisterApiStepGraphObjectDataUploaderParams) {
   return createQueuedStepGraphObjectDataUploader({
     stepId,
@@ -161,7 +161,7 @@ export function createPersisterApiStepGraphObjectDataUploader({
           graphObjectData,
           uploadBatchSize,
           uploadRelationshipsBatchSize,
-          uploadChunkInMb
+          uploadChunkInMb,
         );
       } catch (err) {
         context.logger.error(
--- a/packages/integration-sdk-runtime/src/execution/uploader.ts
+++ b/packages/integration-sdk-runtime/src/execution/uploader.ts
@@ -4,8 +4,10 @@ import { FlushedGraphObjectData } from '../storage/types';
 import {
   uploadGraphObjectData,
   SynchronizationJobContext,
+  BYTES_IN_MB,
 } from '../synchronization';
 import { randomUUID as uuid } from 'crypto';
+import { MAX_BATCH_SIZE } from '../synchronization/shrinkBatchRawData';
 
 export interface StepGraphObjectDataUploader {
   stepId: string;
@@ -145,6 +147,9 @@ export function createPersisterApiStepGraphObjectDataUploader({
   uploadRelationshipsBatchSize,
   uploadChunkInMb,
 }: CreatePersisterApiStepGraphObjectDataUploaderParams) {
+  if (uploadChunkInMb && uploadChunkInMb * BYTES_IN_MB > MAX_BATCH_SIZE) {
+    uploadChunkInMb = MAX_BATCH_SIZE;
+  }
   return createQueuedStepGraphObjectDataUploader({
     stepId,
     uploadConcurrency,
--- a/packages/integration-sdk-runtime/src/execution/uploader.ts
+++ b/packages/integration-sdk-runtime/src/execution/uploader.ts
@@ -135,7 +135,7 @@ export interface CreatePersisterApiStepGraphObjectDataUploaderParams {
   uploadConcurrency: number;
   uploadBatchSize?: number;
   uploadRelationshipsBatchSize?: number;
-  uploadChunkInBytes?: number;
+  uploadBatchSizeInBytes?: number;
 }
 
 export function createPersisterApiStepGraphObjectDataUploader({
@@ -144,10 +144,13 @@ export function createPersisterApiStepGraphObjectDataUploader({
   uploadConcurrency,
   uploadBatchSize,
   uploadRelationshipsBatchSize,
-  uploadChunkInBytes,
+  uploadBatchSizeInBytes,
 }: CreatePersisterApiStepGraphObjectDataUploaderParams) {
-  if (uploadChunkInBytes && uploadChunkInBytes > MAX_BATCH_SIZE_IN_BYTES) {
-    uploadChunkInBytes = MAX_BATCH_SIZE_IN_BYTES;
+  if (
+    uploadBatchSizeInBytes &&
+    uploadBatchSizeInBytes > MAX_BATCH_SIZE_IN_BYTES
+  ) {
+    uploadBatchSizeInBytes = MAX_BATCH_SIZE_IN_BYTES;
   }
   return createQueuedStepGraphObjectDataUploader({
     stepId,
@@ -165,7 +168,7 @@ export function createPersisterApiStepGraphObjectDataUploader({
           graphObjectData,
           uploadBatchSize,
           uploadRelationshipsBatchSize,
-          uploadChunkInBytes,
+          uploadBatchSizeInBytes,
         );
       } catch (err) {
         context.logger.error(
--- a/packages/integration-sdk-runtime/src/execution/uploader.ts
+++ b/packages/integration-sdk-runtime/src/execution/uploader.ts
@@ -6,7 +6,7 @@ import {
   SynchronizationJobContext,
 } from '../synchronization';
 import { randomUUID as uuid } from 'crypto';
-import { MAX_BATCH_SIZE } from '../synchronization/shrinkBatchRawData';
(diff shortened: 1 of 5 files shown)


mishelashala (approved):

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
```diff
@@ -0,0 +1,117 @@
...
+      bestIndex = mid;
+      if (size >= thSize) {
+        return mid; //if its **close enough** to the target size
+      }
+    }
+    if (targetSize < size) {
+      right = mid - 1;
+      if (left <= right) {
+        const newMid = Math.floor((left + right) / 2);
+        size = size - getSizeOfObject(data.slice(newMid, mid));
```
ndowmon, line 67: I think `JSON.stringify()` is probably the most expensive operation we're using here, so the most efficient algorithm would only calculate the size of each object once. It seems like the binary search algorithm here might be calculating the size of each given object more than once, in the event that `if (targetSize < size)` ever evaluates to `true`.

What about looping through each object, calculating its size just once, and then building up your chunks one-by-one?

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
```diff
@@ -0,0 +1,117 @@
...
+      'Entity/Relationship is too big to ingest. This usually happens when item has a lot of long attributes.',
+    );
+  }
+}
+
+export function getSizeOfObject<T extends UploadDataLookup, K extends keyof T>(
+  object: T[K][],
+): number {
+  try {
+    return JSON.stringify(object).length;
```
ndowmon, line 109: I'm not exactly sure, but I'm seeing some evidence from this project (https://github.com/shelfio/array-chunk-by-size/blob/master/src/index.ts#L62) that the size in bytes for an object may not exactly equal its string length:

```ts
function getObjectSize<T>(obj: T): number {
  try {
    const str = stringify(obj);

    return Buffer.byteLength(str, 'utf8');
  ...
}
```

Can you see if `.length` is equivalent to `Buffer.byteLength()`? Or if it's safer to use `Buffer.byteLength()`?

ndowmon: I also notice that the same package I linked uses `json-stringify-safe` package, which is "Like JSON.stringify, but doesn't throw on circular references."

I don't expect our entities/relationships to have circular references, but it is probably worth doing the same here for safety.

ndowmon (commented):
Thanks for pushing this work forward! I've made a few suggestions - please let me know if you'd like to set up some time to discuss.

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
ndowmon: Maybe something like this: 

```ts
export function batchGraphObjectsBySizeInBytes<T extends UploadDataLookup, K extends keyof T>(
  graphObjects: T[K][],
  maximumBatchSizeInBytes: number,
  logger: IntegrationLogger,
): T[K][][] {
  const batches: T[K][][] = [];

  let currentBatch: T[K][] = [];
  let currentBatchSize: number = 0;
  for (const graphObject of graphObjects) {
    let graphObjectSizeInBytes = getSizeOfObject(graphObject);

    if (graphObjectSizeInBytes > maximumBatchSizeInBytes) {
      const rawDataSizeInBytes = getSizeOfObject(graphObject._rawData);

      if (graphObjectSizeInBytes-rawDataSizeInBytes > maximumBatchSizeInBytes) {
        logger.warn({ 
          _key, _type, maximumBatchSizeInBytes, graphObjectSizeInBytes, rawDataSizeInBytes,
         }, 'Graph object is larger than batch size in bytes; cannot upload graph object.');
         continue;
      } else {
        logger.info({ 
          _key, _type, maximumBatchSizeInBytes, graphObjectSizeInBytes, rawDataSizeInBytes,
         }, 'Graph object is larger than batch size in bytes; removing raw data.');
         graphObjectSizeInBytes -= rawDataSizeInBytes;
         graphObject._rawData = undefined;
      }
    }

    if (graphObjectSizeInBytes + currentBatchSize >= maximumBatchSizeInBytes) {
      batches.push(currentBatch);
      currentBatch = [];
      currentBatchSize = 0;
    }
    currentBatch.push(graphObject);
    currentBatchSize += graphObjectSizeInBytes;
  }

  if (currentBatch.length > 0) {
    batches.push(currentBatch);
  }
  return batches;
}
```

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
Gonzalo-Avalos-Ribas: console.log("🌏".length)//2
console.log(Buffer.byteLength("🌏"))//4
console.log("a".length)//1
console.log(Buffer.byteLength("a"))//1

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
```diff
@@ -0,0 +1,89 @@
...
+export function getSizeOfObject<T extends UploadDataLookup, K extends keyof T>(
+  object: T[K][],
+): number {
+  try {
+    return Buffer.byteLength(JSON.stringify(object), 'utf8');
+  } catch (error) {
+    if (error instanceof RangeError) {
+      //If object is too large to size, stringify runs out of memory. We fallback to the max batchSize.
+      return MAX_BATCH_SIZE;
+    }
```
ndowmon, line 86: Is this true? If you get a range error, doesn't it seem likely that the the object is _larger_ than the max batch size (and it's probably unsafe to send it to the persister)?

packages/integration-sdk-runtime/src/synchronization/shrinkBatchRawData.ts
```diff
@@ -14,7 +14,7 @@ import { IntegrationLogger } from '../logger';
 // however, we do have to consider the size of whatever aws sdk wraps
 // our payload in. In practice we have seen batches as small as 5800000
 // fail. To be completely safe, we are using 5500000 bytes as default
-const MAX_BATCH_SIZE = 5500000;
+export const MAX_BATCH_SIZE = 5500000;
```
ndowmon, line 17: I definitely recommend renaming this to `MAX_BATCH_SIZE_IN_BYTES`, and looking at some of the multiplication by `BYTES_IN_MB` that we have to do in the code, I think defaulting to batch size in _bytes_ all around would be a good change. (e.g. change `uploadBatchSizeInMb` to `uploadBatchSizeInBytes` everywhere)

```suggestion
export const MAX_BATCH_SIZE_IN_BYTES = 5500000;
```

Also, tip: for readability of large numbers in javascript, you can add arbitrary underscore characters. I use these to often when delineating the thousands, millions, billions, etc. digits.

```suggestion
export const MAX_BATCH_SIZE_IN_BYTES = 5_500_000;
```

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
```diff
@@ -0,0 +1,89 @@
...
+const BATCH_THRESHOLD = 0.9; //will stop chunking if chunk size is between BATCH_THRESHOLD*sizeInBytes and sizeInBytes
+
+export function chunkBySize<T extends UploadDataLookup, K extends keyof T>(
+  data: T[K][],
+  sizeInBytes: number,
+  logger: IntegrationLogger,
+): T[K][][] {
+  return chunk(data, sizeInBytes, logger);
+}
+export function chunk<T extends UploadDataLookup, K extends keyof T>(
```
ndowmon, line 17: It's a little confusing that there is another function called `chunk` that's used in `synchronization/index.ts` - Also, the functions in this file `chunkBySize` and `chunk` are identical. I recommend just moving your logic into `chunkBySize()` and removing `chunk()` from this file, to avoid making `chunk()` an overloaded function name.

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+  chunkedData.push(data.slice(0, bestIndex));
+  if (bestIndex !== data.length)
+    chunkedData.push(
+      ...chunk(data.slice(bestIndex), sizeInBytes, logger), //chunk the rest of the data
+    );
+  return chunkedData;
+}
+//We are linearly searching for the best index
+function linearSearch<T extends UploadDataLookup, K extends keyof T>(
```
ndowmon, line 40: Formatting nit: make sure to leave a space between functions, and always prefer docstrings over comments when describing the intent of a function.

```suggestion

/** 
 * We are linearly searching for the best index
 */
function linearSearch<T extends UploadDataLookup, K extends keyof T>(
```

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
```diff
@@ -0,0 +1,89 @@
...
+): T[K][][] {
+  return chunk(data, sizeInBytes, logger);
+}
+export function chunk<T extends UploadDataLookup, K extends keyof T>(
+  data: T[K][],
+  sizeInBytes: number,
+  logger: IntegrationLogger,
+): T[K][][] {
+  const chunkedData: T[K][][] = [];
+  if (getSizeOfObject(data) < sizeInBytes) {
```
ndowmon, line 23: Still not sure this is quite right - we're now still calculating the size of the entire batch, plus the size of each individual object, so calculating size twice. 

Let's meet and talk through https://github.com/JupiterOne/sdk/pull/925#discussion_r1277783164

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
ndowmon: Can you try and determine around what object size we encounter `RangeError`? (e.g. 20mb)

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
```diff
@@ -0,0 +1,86 @@
...
+  sizeInBytes: number,
+  logger: IntegrationLogger,
+) {
+  const bigBatch = data.slice(0, 1) as Entity[] | Relationship[];
+  if (bigBatch[0] && bigBatch[0]._rawData) {
+    logger.info(
+      bigBatch[0].key,
+      'Entity/Relationship is too big to ingest. Will remove _rawData',
+    );
+    bigBatch[0]._rawData = [];
```
Gonzalo-Avalos-Ribas, line 61: should be undefined

packages/integration-sdk-runtime/src/execution/uploader.ts
```diff
@@ -134,6 +135,7 @@ export interface CreatePersisterApiStepGraphObjectDataUploaderParams {
   uploadConcurrency: number;
   uploadBatchSize?: number;
   uploadRelationshipsBatchSize?: number;
+  uploadChunkInBytes?: number;
```
ndowmon, line 138:
```suggestion
  uploadBatchSizeInBytes?: number;
```

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
```diff
@@ -0,0 +1,102 @@
...
+import { UploadDataLookup } from '.';
+import { MAX_BATCH_SIZE_IN_BYTES } from './shrinkBatchRawData';
+
+/**
+Converts an array of graphObjects into an array of groups of this objects.
+@param graphObjects  Entities or relationships to group.
+
+@param maximumBatchSizeInBytes The maximum size of the group of objects.
+
+@param logger
```
ndowmon, line 16:
```suggestion
 * Converts an array of graphObjects into an array of groups of this objects.
 * @param graphObjects  Entities or relationships to group.
 * 
 * @param maximumBatchSizeInBytes The maximum size of the group of objects.
 * 
 * @param logger 
```

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
```diff
@@ -0,0 +1,102 @@
...
+  logger: IntegrationLogger,
+): T[K][][] {
+  const batches: T[K][][] = [];
+
+  let currentBatch: T[K][] = [];
+  let currentBatchSize: number = 0;
+  for (const graphObject of graphObjects) {
+    const typedGraphObject = graphObject as Entity | Relationship;
+
+    let graphObjectSizeInBytes = getSizeOfObject(graphObject);
```
ndowmon, line 33: I think this needs to be try/caught - in the event that `getSizeOfObject()` throws, we also need to alert the user that we cannot safely upload this object.

```suggestion
    let graphObjectSizeInBytes = 0;
    try { 
      graphObjectSizeInBytes = getSizeOfObject(graphObject);
    } catch (err) {
      logger.warn({ err }, 'Failed to calculate size of object in bytes');
      logger.publishWarnEvent({ ... });
      continue;
    }
```

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
```diff
@@ -0,0 +1,102 @@
...
+  try {
+    return Buffer.byteLength(JSON.stringify(object), 'utf8');
+  } catch (error) {
+    if (error instanceof RangeError) {
+      //If object is too large to size, stringify runs out of memory. We fallback to the max batchSize.
+      //This is highly unlikely since it should only happen after ~500MB.
+      return MAX_BATCH_SIZE_IN_BYTES + 1;
+    }
+    throw error;
+  }
```
ndowmon, line 101: I"m a little concerned about a small edge case with this logic. Let's say for example, you have a graph object (`graphObject`) for which `Buffer.byteLength(JSON.stringify(object), 'utf8')` returns the following:

- `Buffer.byteLength(JSON.stringify(graphObject), 'utf8')`
   `  => RangeError`
- `Buffer.byteLength(JSON.stringify(graphObject._rawData), 'utf8')`
   `  => 1000`

Because we return `MAX_BATCH_SIZE_IN_BYTES + 1` for all `RangeError`, our logic inside of `batchGraphObjectsBySizeInBytes()` could have us unsafely uploading a resource that is too large.

E.g.

1. `getSizeOfObject(graphObject) `
    `  => MAX_BATCH_SIZE_IN_BYTES + 1`
2. `getSizeOfObject(graphObject._rawData) `
    `  => 1000`
3. `getSizeOfObject(graphObject) - getSizeOfObject(graphObject._rawData)`
     `  => MAX_BATCH_SIZE_IN_BYTES - 999` <= # this is below the `MAX_BATCH_SIZE_IN_BYTES`, and will get uploaded

I would recommend removing the `try/catch`ing inside of this function altogether, and letting the caller`catch` and handle any error from `getSizeOfObject()`.

```suggestion
  return Buffer.byteLength(JSON.stringify(object), 'utf8');
```

packages/integration-sdk-runtime/src/synchronization/batchBySize.ts
```diff
@@ -0,0 +1,102 @@
...
+            graphObjectSizeInBytes,
+            rawDataSizeInBytes,
+          },
+          'Graph object is larger than batch size in bytes; cannot upload graph object.',
+        );
+        logger.publishWarnEvent({
+          name: IntegrationWarnEventName.IngestionLimitEncountered,
+          description:
+            'Graph object is larger than batch size in bytes; cannot upload graph object. Will skip graph object.',
+        });
```
ndowmon, line 58: After this PR gets merged, let's talk with @VDubber about how best to handle these job log events and create a follow-on PR with whatever we come up with. For example:

1. What if many entities are too large? We probably want to avoid sending hundreds or thousands of these warn events to user job logs.
2. Should we include information about _what_ couldn't be uploaded? For example, the `_type`, or the `_key`? 
3. How is the user alerted to these `warn_ingestion_limit_encountered` events? Is the job marked as completed with errors, completed with some incomplete data? Is the user error messaging appropriate for these cases?

PR #934, opened 2023-07-27
[SRE-1128] Create generic wrapper for allowing spans to be wrapped

ImDevinC:


packages/integration-sdk-core/src/types/config.ts
```diff
@@ -86,6 +86,7 @@ export interface InvocationConfig<
    *
    */
   ingestionConfig?: IntegrationIngestionConfigFieldMap;
+  stepWrapper?: StepWrapperFunction<TStepExecutionContext>;
```
JakeFerrero, line 89: I'd probably add a doc string for the new parameter and what it could be used for.

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -75,6 +75,7 @@ export function buildStepDependencyGraph<
  * created more leaf nodes and executes them. This continues
  * until there are no more nodes to execute.
  */
+
```
JakeFerrero, line 78: nitpick: shouldn't be a space since this doc string is related to below function?

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -267,7 +270,7 @@ export function executeStepDependencyGraph<
                   stepId,
                   cached: hasCachePath(stepId).toString(),
                 },
-                operation: () => executeStep(step),
+                operation: () => executeStep(step, stepWrapper),
```
JakeFerrero, line 273: Just curious, I see you decided to pass stepWrapper into `executeStep` and then only wrap part of the function (where we call `step.executionHandler`. What's the difference if we instead just wrap all of `executeStep`? Is there any difference in results in terms of spans? Then we could do something like below.

```typescript
let operation: () => Promise<void>;
if (stepWrapper) operation = () => { return stepWrapper(step, executeStep(step)); };
else operation = () => executeStep(step);
```

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -306,7 +309,10 @@ export function executeStepDependencyGraph<
      * Errors from an execution handler are caught and used to
      * determine a status code for the step's result.
      */
-    async function executeStep(step: Step<TStepExecutionContext>) {
+    async function executeStep(
+      step: Step<TStepExecutionContext>,
+      wrapper?: StepWrapperFunction<TStepExecutionContext>,
```
ndowmon, line 314: Technically, you shouldn't need to add `wrapper` to the interface of `executeStep`, since you've already got access to `stepWrapper` in your current scope.

```diff
    async function executeStep(
      step: Step<TStepExecutionContext>,
-     wrapper?: StepWrapperFunction<TStepExecutionContext>,
    ) {
      const { id: stepId } = step;
      
      // ...

        if (status !== StepResultStatus.CACHED) {
-         if (wrapper) {
+         if (stepWrapper) {
-           await wrapper(
+           await stepWrapper(
              step,
              async () => await step.executionHandler(context),
            );
          } else {
            await step.executionHandler(context);
          }

File: packages/integration-sdk-runtime/src/execution/dependencyGraph.ts [hunk 5]
```diff
@@ -345,7 +351,14 @@ export function executeStepDependencyGraph<
         }
 
         if (status !== StepResultStatus.CACHED) {
-          await step.executionHandler(context);
+          if (wrapper) {
```
ndowmon, line 354: I would consider setting a default value for `stepWrapper` in order to avoid needing an `if/else` statement here:

```diff
export function executeStepDependencyGraph<
  TExecutionContext extends ExecutionContext,
  TStepExecutionContext extends StepExecutionContext,
>({
  executionContext,
  inputGraph,
  stepStartStates,
  duplicateKeyTracker,
  graphObjectStore,
  dataStore,
  createStepGraphObjectDataUploader,
  beforeAddEntity,
  beforeAddRelationship,
  afterAddEntity,
  afterAddRelationship,
- stepWrapper,
+ stepWrapper = (_, executionHandler) => executionHandler(),
}: {
  executionContext: TExecutionContext;
  inputGraph: DepGraph<Step<TStepExecutionContext>>;
  stepStartStates: StepStartStates;
  duplicateKeyTracker: DuplicateKeyTracker;
  graphObjectStore: GraphObjectStore;
  dataStore: MemoryDataStore;
  createStepGraphObjectDataUploader?: CreateStepGraphObjectDataUploaderFunction;
  beforeAddEntity?: BeforeAddEntityHookFunction<TExecutionContext>;
  beforeAddRelationship?: BeforeAddRelationshipHookFunction<TExecutionContext>;
  afterAddEntity?: AfterAddEntityHookFunction<TExecutionContext>;
  afterAddRelationship?: AfterAddRelationshipHookFunction<TExecutionContext>;
  stepWrapper?: StepWrapperFunction<TStepExecutionContext>;
}): Promise<IntegrationStepResult[]> {

  // ...

        if (status !== StepResultStatus.CACHED) {
+         await stepWrapper({ step }, () => step.executionHandler());
-         if (wrapper) {
-           await wrapper(
-             step,
-             async () => await step.executionHandler(context),
-           );
-         } else {
-           await step.executionHandler(context);
-         }

File: packages/integration-sdk-core/src/types/step.ts [hunk 6]
```diff
@@ -240,3 +240,10 @@ export type StepMetadata = StepGraphObjectMetadataProperties & {
    */
   ingestionSourceId?: string;
 };
+
+export type StepWrapperFunction<
```
ndowmon, line 244: Just thinking out loud - this is meant to wrap the execution handler (or step execution handler) - just proposing a few alternative interface names, in case it improves clarity:

- `StepWrapperFunction`
- `ExecutionHandlerWrapperFunction`
- `StepExecutionHandlerWrapperFunction`

packages/integration-sdk-core/src/types/step.ts
```diff
@@ -240,3 +240,10 @@ export type StepMetadata = StepGraphObjectMetadataProperties & {
...
+  TStepExecutionContext extends StepExecutionContext,
+> = (
+  step: Step<TStepExecutionContext>,
```
ndowmon, line 247: Do you foresee ever needing anything beyond `step` in the future? Sometimes making the first argument an object allows simple extension in the future, in case we need to add something (e.g., a logger) in the future.

```suggestion
  context: { 
    step: Step<TStepExecutionContext>,  
    // logger: IntegrationLogger,
    // context: TStepExecutionContext,
  }
```

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
ImDevinC: Because the error is caught on line 386, we never get the error shown in the trace

packages/integration-sdk-core/src/types/step.ts
ImDevinC: It can't hurt, that's for sure. Will update accordingly

packages/integration-sdk-core/src/types/step.ts
ImDevinC: Agreed

packages/integration-sdk-core/src/types/config.ts
```diff
@@ -86,6 +90,14 @@ export interface InvocationConfig<
    *
    */
   ingestionConfig?: IntegrationIngestionConfigFieldMap;
+  /**
+   * Wraps the executionHandler for each step in an operation to allow for adding
+   * context before and after the executionHandler completes. Can be used for adding
+   * logic like tracing or logging.
+   *
+   * If not provided, this defaults to a noop wrapper.
```
JakeFerrero, line 98: Biggest nitpick of the century: If I understand correctly, looking at line 358 in dependancyGraph.ts, if no wrapper is provided we just run the handler as is; not run a noop wrapper. So these docs _could_ be updated but it's really not a big deal.

JakeFerrero (dismissed):

packages/integration-sdk-core/src/types/config.ts
ImDevinC: It's worth fixing so it doesn't cause confusion later, will update accordingly

JakeFerrero (dismissed):

yarn.lock
i5o, line 1: Any reason on to why you switched to [npmjs.org](http://npmjs.org/) instead of yarn pkg?
Did you ran npm install instead of yarn install?

i5o (dismissed):
LGTM. 👍 🚀

i5o (approved):

PR #943, opened 2023-08-16
Add eventCode to event input

VDubber:


packages/integration-sdk-runtime/src/logger/index.ts
```diff
@@ -399,20 +405,27 @@ export class IntegrationLogger
 
   validationFailure(err: Error) {
     const eventName = 'validation_failure';
-    const { errorId, description } = createErrorEventDescription(
+    const { errorId, errorCode, description } = createErrorEventDescription(
```
VDubber, line 408: All we get from the graph projects is an error during the validationFailure phase. We could enhance this but I don't know if it is necessary.

packages/integration-sdk-core/src/types/logger.ts
```diff
@@ -40,6 +40,10 @@ export type IntegrationEvent = {
 export type PublishEventInput = {
   name: string;
   description: string;
+  /**
+   * Allows the event to be identified via a code.
```
ndowmon, line 44: Can you add examples to the docstrings?

packages/integration-sdk-core/src/types/logger.ts
```diff
@@ -40,6 +40,10 @@ export type IntegrationEvent = {
...
+   */
+  eventCode?: string;
```
ndowmon, line 46: thought? But i think `eventCode` is chill as well.
```suggestion
  eventMetadata?: { [k: string]: string };
```

ndowmon: This might cause issues with our consumer...

ndowmon (dismissed):

ndowmon (dismissed):

PR #944, opened 2023-08-22
Initial commit; get more data to uploadData function

Gonzalo-Avalos-Ribas:


packages/integration-sdk-runtime/src/storage/memory.ts
```diff
@@ -203,6 +217,7 @@ export class InMemoryGraphObjectStore implements GraphObjectStore {
 
       entityTypeKeysMap.delete(entity._key);
     }
+    this.entitySizeInBytes[stepId] = 0;
```
ndowmon, line 220: Not a huge deal, but it might be better to delete `this.entitySizeInBytes[stepId]` rather than set to `0` - AWS, for example, has 399 steps, so by the end of the integration, the `getTotalEntitySizeInBytes()` call will be summing 399 counts, most of which will be `0`. I think this operation occurs on every `jobState.addEntity()` call.

Maybe you can look into the performance of `getTotalEntitySizeInBytes()` when its an object with 1 property compared to an object of 400 properties.

packages/integration-sdk-runtime/src/storage/memory.test.ts
```diff
@@ -201,9 +201,9 @@ describe('#InMemoryGraphObjectStore', () => {
 
     const e1 = createTestEntity();
     const e2 = createTestEntity();
-
-    await store.addEntities(uuid(), [e1, e2]);
-    store.flushEntities([e1]);
+    const stepId = uuid();
+    await store.addEntities(stepId, [e1, e2]);
+    store.flushEntities([e1], stepId);
```
ndowmon, line 206: Hm, a little awkward here - it looks like the value of `store.getTotalEntitySizeInBytes()` would be zero (even though there is actually one entity in the store, and thus the size in bytes is >0).

You made a previous suggestion while we talked in-person about actually decrementing the size in the `flushEntities()` function - and I recommended this optimization.

However, I now wonder if your initial proposal, decrementing the size inside of `flushEntities()` itself, is worth implementing. 

Let's discuss. At the very least, we should add a `TODO` note in `getTotalEntitySizeInBytes()` and `getTotalRelationshipSizeInBytes()` about the out-of-sync issue between `entityCount` and `entitySizeInBytes`, or maybe we just should decrement inside of `flushEntities()`.

Gonzalo-Avalos-Ribas: Could be an improvement in terms of consistency but its slower. I don't see that we are using `flushEntities` or `flushRelationships` anywhere else.
If the method flushes all entities for a step, always, then i def makes sense to have it how it is now.
We could also create a new method called `flushEntitiesOfStep` that includes the stepId and it removes all the entities using the stepId.  And we modify the original to measure the size and decrement.

Gonzalo-Avalos-Ribas:
I added a log and tried it on aws dev. Here are all the uploads that it makes with the changes.
It would be an improvement to not flush after each step since thats causing a lot of uploads with only one entity.
We also managed to do 1 upload with almost 900 entities proving that the changes have impacted correctly~ 
[Logs.csv](https://github.com/JupiterOne/sdk/files/12440117/logs-insights-results.1.csv)

ndowmon (approved):
WIll you add CHANGELOG in this PR or when you version?

PR #946, opened 2023-09-01
Allow Lambda to be Passed to Alpha

zemberdotnet:
- fix: allow Lambda class to be pased to Alpha
- update CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to
 
 ## Unreleased
 
+## 10.2.0 - 2023-09-01
+
+- Allow user-defined `Lambda` class to be passed to Alpha in `createApiClient`
+
 ## 10.1.0 - 2023-08-28
 
 - 'InMemoryGraphObjectStore' no longer stores a set amount of entities or
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to
 
 ## 10.2.0 - 2023-09-01
 
-- Allow user-defined `Lambda` class to be passed to Alpha in `createApiClient`
+- Allow user to override `AlphaOptions` in `createApiClient`
 
 ## 10.1.0 - 2023-08-28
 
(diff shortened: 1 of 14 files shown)


packages/integration-sdk-runtime/src/api/index.ts
```diff
@@ -18,6 +19,7 @@ interface CreateApiClientInput {
   account: string;
   accessToken?: string;
   retryOptions?: RetryOptions;
+  lambda?: typeof Lambda;
```
ndowmon, line 22: Confusingly, the alpha client options include both `lambda` (a callback function) and `Lambda` (the class). I recommend using `Lambda` as the argument in case we want to extend and use `lambda` at some point in the future.

packages/integration-sdk-runtime/src/api/index.ts
```diff
@@ -61,6 +64,10 @@ export function createApiClient({
     opts.signAwsV4 = { credentials: defaultProvider({ maxRetries: 3 }) };
   }
 
+  if (lambda) {
+    opts.Lambda = lambda;
+  }
```
ndowmon, line 69: Is the intention to have the managed runtime create and pass this in? I guess I expected the SDK to create the `Lambda` client

zemberdotnet: That was the intention. I was hoping to keep most of the `aws-sdk` things out of the SDK, as for the open-source graph projects, they aren't really needed. But what's your thoughts on why you expected it here?

ndowmon: Good point. Maybe we should actually just allow `alphaOptions` to be passed in to this function. Then we could remove `isLambdaUrl()` and all the stuff around `signAwsV4` from this project, which would be idea.

And I guess the `signAwsV4` stuff isn't even necessary anymore anyway

ndowmon: This way you wouldn't even need to explicitly import `@aws-sdk/client-lambda`

ndowmon (dismissed):

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to
 
 ## Unreleased
 
+## 10.2.0 - 2023-09-01
+
+- Allow user-defined `Lambda` class to be passed to Alpha in `createApiClient`
```
ndowmon, line 14:
```suggestion
- Allow user to override `AlphaOptions` in `createApiClient`
```

ndowmon (approved):

PR #949, opened 2023-09-07
Batch by size added by default

Gonzalo-Avalos-Ribas:


packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -37,7 +37,7 @@ export interface FileSystemGraphObjectStoreParams {
   integrationSteps?: IntegrationStep[];
   /**
    * The maximum size in bytes of entities/relationships stored in memory at one time.
-   * default: 25_000_000
```
Gonzalo-Avalos-Ribas, line 40: Message was wrong

gastonyelmini (approved):

PR #951, opened 2023-09-08
Logging Improvement / Error Updates

zemberdotnet:
- feat: use IntegrationLogger for all logs in run command
- update CHANGELOG
- [fix: update err serializer to use node inspect](https://github.com/JupiterOne/sdk/pull/951/commits/7f32fa8126f1e20ddd1047bc30e8caeac50d52b9)
[7f32fa8](https://github.com/JupiterOne/sdk/pull/951/commits/7f32fa8126f1e20ddd1047bc30e8caeac50d52b9)
- [update tsconfig to produce es2022](https://github.com/JupiterOne/sdk/pull/951/commits/8dcc52ce90a812d140ec81793f27b9344f19cad2)


packages/integration-sdk-cli/src/__tests__/cli-run.test.ts
```diff
@@ -80,92 +74,19 @@ test('disables graph object schema validation', async () => {
   expect(process.env.ENABLE_GRAPH_OBJECT_SCHEMA_VALIDATION).toBeUndefined();
 });
 
-test('step should fail if enableSchemaValidation = true', async () => {
```
zemberdotnet, line 83: I removed these two tests. It would be somewhat difficult to test them now that the run command doesn't rely on the `log` package. It's just a boolean flag and the actual implementation is tested elsewhere.

packages/integration-sdk-cli/src/__tests__/cli-run.test.ts
```diff
@@ -256,7 +139,9 @@ test('does not publish events for source "api" since there is no integrationJobI
 
   let eventsPublished = false;
   polly.server
-    .post(`https://example.com/persister/synchronization/jobs/${job.id}/events`)
+    .post(
+      `https://api.us.jupiterone.io/persister/synchronization/jobs/${job.id}/events`,
```
zemberdotnet, line 143: This test was passing incorrectly. 😱

packages/integration-sdk-cli/src/__tests__/util/synchronization.ts
```diff
@@ -97,8 +97,14 @@ export function generateSynchronizationJob(
...
-    integrationJobId: options?.integrationJobId || 'test-job-id',
-    integrationInstanceId: options?.integrationInstanceId || 'test-instance-id',
+    integrationJobId:
+      options?.source === 'api'
+        ? undefined
+        : options?.integrationJobId || 'test-job-id',
+    integrationInstanceId:
+      options?.source === 'api'
+        ? undefined
+        : options?.integrationInstanceId || 'test-instance-id',
```
zemberdotnet, line 107: Fixes the `api` doesn't publish events test.

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -261,11 +261,6 @@ export async function finalizeSynchronization({
     );
   }
 
-  logger.info(
-    { synchronizationJob: finalizedJob },
-    'Synchronization finalization result.',
-  );
-
```
zemberdotnet, line 268: I debated keeping this. The other synchronization job functions `synchronizationStatus` and `synchronizationAbort` don't log their own results. If this goes through I'll likely add it back in inside the managed integration.

I do end up logging the job but in `run.ts` instead.

packages/integration-sdk-cli/src/commands/run.ts
```diff
@@ -52,15 +51,19 @@ export function run(): Command {
...
-      log.debug(
-        `Configured JupiterOne API client. (apiBaseUrl: '${clientApiOptions.apiBaseUrl}', account: ${clientApiOptions.account})`,
-      );
 
       let logger = createIntegrationLogger({
         name: 'local',
         pretty: !options.noPretty,
       });
 
+      logger.info(
```
ImDevinC, line 60: Any reason we're changing from debug to info?

zemberdotnet: Previously this was using the `log.ts` instead of the actual logger. That code looks like this:
https://github.com/JupiterOne/sdk/blob/447bba4c679b24316a16cfb86a7813d5efd19f74/packages/integration-sdk-cli/src/log.ts#L14-L29

Despite it being set at debug it would always be logged since the log levels in that file are non-functioning. I decided to just keep the same behavior by setting this one to info.

tsconfig.base.json
```diff
@@ -2,8 +2,8 @@
   "exludes": ["jest.config.base.js"],
   "compilerOptions": {
     "composite": true,
-    "target": "ES2021",
-    "lib": ["ES2021", "DOM"],
+    "target": "ES2022",
+    "lib": ["ES2022", "DOM"],
```
zemberdotnet, line 6: I've been experiencing a lot of problems downstream with typescript complaining of the following error:
```
node_modules/@jupiterone/integration-sdk-core/dist/src/errors.d.ts:76:5 - error TS2425: Class 'Error' defines instance member property 'cause', but extended class 'IntegrationError' defines it as instance member function.

76     cause(): Error | undefined;
```
This is because we define `cause` as a method (i.e. `cause(): Error | undefined`) while the parent class in node 18's types defines it as `cause: Error | undefined`.

I've remedied this in the `errors.ts` file. I've also updated the error serializer in `integration-sdk-runtime` using Node's native `inspect`. See their recommendation here:
https://nodejs.org/docs/latest-v18.x/api/errors.html#new-errormessage-options

While it is technically a breaking change **cause**, I don't anticipate major versioning since the only reference I could find to it was in a test in a private package.

packages/integration-sdk-core/src/errors.ts
```diff
@@ -79,21 +80,14 @@ export class IntegrationError extends Error {
   readonly fatal: boolean | undefined;
 
   constructor(options: IntegrationErrorOptions) {
-    super(options.message);
+    super(options.message, { cause: options.cause });
```
zemberdotnet, line 83: See comment in `tsconfig.base.json`.

tsconfig.base.json
zemberdotnet: The switch to ES2022 shouldn't be breaking either as we already required node 18 and it supports all es2022 features.

ndowmon (dismissed):

PR #952, opened 2023-09-10
Revert "Initial commit; get more data to uploadData function"

ndowmon:
Reverts JupiterOne/sdk#944. We are seeing some instability in production with this change, so I recommend reverting and publishing a new minor version of this change.

Later, once stable, we can un-revert this change.


packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -18,37 +18,31 @@ import {
...
 const BINARY_SEMAPHORE_CONCURRENCY = 1;
 
 export interface FileSystemGraphObjectStoreParams {
   integrationSteps?: IntegrationStep[];
-  /**
-   * The maximum size in bytes of entities/relationships stored in memory at one time.
-   * default: 25_000_000
-   */
-  graphObjectBufferThresholdInBytes?: number;
+
```
ndowmon, line 36: This is a public interface, so we should not remove any arguments. We'll just leave this property unused for now.

```suggestion
  /**
   * The maximum size in bytes of entities/relationships stored in memory at one time.
   * default: 25_000_000
   */
  graphObjectBufferThresholdInBytes?: number;
```

Gonzalo-Avalos-Ribas (approved):

PR #957, opened 2023-09-13
Improve integration generator

geovannimp:
- Fix github flow
- Improve types
- Update dependencies
- Update README

### DIFF
--- a/packages/integration-sdk-cli/src/generator/template/.github/workflows/build.yml.hbs
+++ b/packages/integration-sdk-cli/src/generator/template/.github/workflows/build.yml.hbs
@@ -40,7 +40,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-          token: ${{ secrets.AUTO_GITHUB_PAT_TOKEN }}
+          token: $\{{ secrets.AUTO_GITHUB_PAT_TOKEN }}
       - name: Setup Node
         uses: actions/setup-node@v3
         with:
@@ -49,5 +49,5 @@ jobs:
       - name: Build and Release
         uses: jupiterone/action-npm-build-release@v1
         with:
-          npm_auth_token: ${{ secrets.NPM_AUTH_TOKEN }}
-          gh_token: ${{ secrets.AUTO_GITHUB_PAT_TOKEN }}
+          npm_auth_token: $\{{ secrets.NPM_AUTH_TOKEN }}
+          gh_token: $\{{ secrets.AUTO_GITHUB_PAT_TOKEN }}
--- a/packages/integration-sdk-cli/src/generator/template/src/config.ts.hbs
+++ b/packages/integration-sdk-cli/src/generator/template/src/config.ts.hbs
@@ -2,20 +2,19 @@ import {
   IntegrationExecutionContext,
   {{#if configFields}}IntegrationValidationError,{{/if}}
   IntegrationInstanceConfigFieldMap,
-  IntegrationInstanceConfig,
 } from '@jupiterone/integration-sdk-core';
 import { createAPIClient } from './client';
 
-export const instanceConfigFields: IntegrationInstanceConfigFieldMap = {
+export const instanceConfigFields = {
   {{#each configFields}}
   {{camelCase field}}: {
     type: '{{type}}',
     mask: {{mask}}
   },
   {{/each}}
-};
+} satisfies IntegrationInstanceConfigFieldMap;
 
-export interface IntegrationConfig extends IntegrationInstanceConfig {
+export interface IntegrationConfig {
   {{#each configFields}}
   {{camelCase field}}: {{ type }};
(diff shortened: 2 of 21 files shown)


packages/integration-sdk-cli/src/generator/template/src/config.ts.hbs
```diff
@@ -2,20 +2,19 @@ import {
...
   {{camelCase field}}: {
     type: '{{type}}',
     mask: {{mask}}
   },
   {{/each}}
-};
+} satisfies IntegrationInstanceConfigFieldMap;
 
-export interface IntegrationConfig extends IntegrationInstanceConfig {
+export interface IntegrationConfig {
```
geovannimp, line 17: `IntegrationInstanceConfig` have type `Record<string, any>`, so any place using `instanceConfigFields` or `IntegrationConfig` would accept any property as key.

packages/integration-sdk-cli/src/generator/template/.github/workflows/build.yml.hbs
```diff
@@ -49,5 +49,5 @@ jobs:
       - name: Build and Release
         uses: jupiterone/action-npm-build-release@v1
         with:
-          npm_auth_token: ${{ secrets.NPM_AUTH_TOKEN }}
-          gh_token: ${{ secrets.AUTO_GITHUB_PAT_TOKEN }}
+          npm_auth_token: $\{{ secrets.NPM_AUTH_TOKEN }}
```
geovannimp, line 52: The generator was generating these properties empty as `{{}}` is a handlebars expression

packages/integration-sdk-cli/src/generator/template/src/steps/constants.ts.hbs
```diff
@@ -4,25 +4,23 @@ import {
...
+export const Steps = {
   {{#each steps}}
   {{constantCase name}}: '{{kebabCase name}}',
   {{/each}}
-};
+} satisfies Record<string, string>;
 
-export const Entities: Record<{{#unless entities}}string{{/unless}}{{#each entities}}'{{constantCase resourceName}}'{{#unless @last}}|{{/unless}}{{/each}}
-, StepEntityMetadata > = {
+export const Entities = {
```
geovannimp, line 13: Replace with satisfies so we have stronger type and don't need to repeat code.

packages/integration-sdk-cli/src/generator/template/.github/workflows/build.yml.hbs
zemberdotnet: Nice catch!

packages/integration-sdk-cli/src/generator/template/src/steps/constants.ts.hbs
```diff
@@ -4,25 +4,23 @@ import {
   StepRelationshipMetadata,
 } from '@jupiterone/integration-sdk-core';
 
-export const Steps: Record<{{#unless steps}}string{{/unless}}{{#each steps}}'{{constantCase name}}'{{#unless @last}}|{{/unless}}{{/each}}, string> = {
+export const Steps = {
   {{#each steps}}
   {{constantCase name}}: '{{kebabCase name}}',
   {{/each}}
-};
+} satisfies Record<string, string>;
```
zemberdotnet, line 11: If we switch to `Record<string, string>` from the above will we lose any type information. I assumed the type hinting and suggestions came from the `Record<string,  'STEP_1' | 'STEP_2'>`

geovannimp: Actually `satisfies` fix this problem, it will validate if the returned type from the object matches the Record type, but return the original type. We have been using this in new integrations.
image

zemberdotnet (dismissed):

geovannimp:
@zemberdotnet Do you know if I need to update all packages version to release this?

zemberdotnet:
@geovannimp Yes, you'll need to `yarn lerna version` to bump the package versions and release this. We'll also need to update the CHANGELOG.

geovannimp:
@zemberdotnet  Version and Changelog updated, can you check again?

zemberdotnet:
@geovannimp Sorry, I failed to mention this earlier, but the version commit needs to be the last in the PR. You'll need to revert the changes from https://github.com/JupiterOne/sdk/pull/957/commits/e962986164828972dd7cbd18019cb9e493fa0975.

You'll also need to delete the tags before `yarn lerna version`'ing again.
```sh
# Delete tag locally
git tag -d v10.5.2

# Delete tag remotely
git push --delete origin v10.5.2
```

geovannimp:
@zemberdotnet Thanks, this make sense. Edited the commits/tag, and now it should be fine.

zemberdotnet (approved):
🚀

PR #958, opened 2023-09-13
Log info of requets that are longer than 10 seconds.

Gonzalo-Avalos-Ribas:


RonaldEAM (approved):

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -508,7 +508,18 @@ export async function uploadDataChunk<
...
+          {
+            uploadCorrelationId,
+            uploadType: type,
+            attemptNum: ctx.attemptNum,
+            batchSize: batch.length,
+            batchSizeInBytes: getSizeOfObject(batch),
+          },
+          'Finished uploading big batch',
+        );
+      }
```
ndowmon, line 522: This is the relevant change.

PR #965, opened 2023-09-27
Set entities and relationships to be optional in step definition

i5o:
Some steps don't generate entities or relationships, we shouldn't enforce to declare them.
We already do this with mappedRelationships.

### DIFF
--- a/packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
+++ b/packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
@@ -91,7 +91,7 @@ function integrationStepsToGraphObjectIndexMetadataMap(
       relationships: new Map(),
     };
 
-    for (const entityMetadata of step.entities) {
+    for (const entityMetadata of step.entities || []) {
       if (entityMetadata.indexMetadata) {
         metadataMap.entities.set(
           entityMetadata._type,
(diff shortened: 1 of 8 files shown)


packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -91,7 +91,7 @@ function integrationStepsToGraphObjectIndexMetadataMap(
       relationships: new Map(),
     };
 
-    for (const entityMetadata of step.entities) {
+    for (const entityMetadata of step.entities || []) {
```
Nick-NCSU, line 94: Isn't this typed in such a way that we always require at least an empty array?

i5o: It is. We shouldn't

i5o: Pushed an additional commit for this

zemberdotnet: Why shouldn't we?
It's easy to search for
```
entities: []
```
But very difficult to search for it's absence.

i5o: Once you're inside an executionHandler it's easy to just see what doesn't produce entities; it's kinda the same as you do but the opposite, search for entities and if not found then no entities, this just came up as I was trying to make a dummy step and having to add entities/relationships seems like an extra step that adds no value.

With that being said, moving to draft

i5o: btw -- same logic you commented would apply for mappedRelationships

PR #966, opened 2023-09-28
Remove unused batch size (count) parameters

xdumaine:
The `uploadBatchSize` and `uploadRelationshipsBatchSize` parameters are entirely unused, because they're piped down all the way to `uploadData` where the parameter is unused and instead it uses `uploadBatchSizeInBytes` instead (the last two lines in this PR).

This will be released in a major version update due to it being a breaking change to the interfaces/CLI commands.

### DIFF
--- a/packages/integration-sdk-cli/src/commands/options.ts
+++ b/packages/integration-sdk-cli/src/commands/options.ts
@@ -1,5 +1,4 @@
 import {
-  DEFAULT_UPLOAD_BATCH_SIZE,
   getAccountFromEnvironment,
   getApiKeyFromEnvironment,
   JUPITERONE_DEV_API_BASE_URL,
@@ -39,8 +38,6 @@ export interface SyncOptions {
   source: 'integration-managed' | 'integration-external' | 'api';
   scope?: string | undefined;
   integrationInstanceId?: string | undefined;
-  uploadBatchSize: number;
-  uploadRelationshipBatchSize: number;
   skipFinalize: boolean;
 }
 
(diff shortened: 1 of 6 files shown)


packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -558,7 +535,6 @@ export async function uploadData<T extends UploadDataLookup, K extends keyof T>(
   { job, apiClient, logger }: SynchronizationJobContext,
   type: K,
   data: T[K][],
-  uploadBatchSize?: number,
   uploadBatchSizeInBytes?: number,
```
xdumaine, line 538: the `uploadBatchSize` argument here was unused, which means all of the above is unused.

xdumaine: A follow up to this PR will allow overriding `DEFAULT_UPLOAD_BATCH_SIZE_IN_BYTES = 5_000_000;` to be higher.

xdumaine: https://github.com/JupiterOne/sdk/pull/967

packages/integration-sdk-cli/src/commands/options.ts
```diff
@@ -69,18 +66,6 @@ export function addSyncOptionsToCommand(command: Command): Command {
...
-      'specify number of entities and relationships per upload batch',
-      (value, _previous: Number) => Number(value),
-      DEFAULT_UPLOAD_BATCH_SIZE,
-    )
-    .option(
-      '-ur, --upload-relationship-batch-size <number>',
-      'specify number of relationships per upload batch, overrides --upload-batch-size',
-      (value, _previous: Number) => Number(value),
-      DEFAULT_UPLOAD_BATCH_SIZE,
-    )
```
zemberdotnet, line 83: If someone was previously using the `upload-batch-size` parameter and we remove it, will it break their script?
I understand it isn't actually being used, but I want to know if this will create a hard failure in their script.

xdumaine: Yes, this is a breaking change. I assume we have a mechanism to mark it as such - PR label? I'll add an update to the changelog.

packages/integration-sdk-runtime/src/execution/uploader.ts
```diff
@@ -133,17 +133,13 @@ export interface CreatePersisterApiStepGraphObjectDataUploaderParams {
...
-  uploadBatchSize?: number;
-  uploadRelationshipsBatchSize?: number;
   uploadBatchSizeInBytes?: number;
 }
 
 export function createPersisterApiStepGraphObjectDataUploader({
   stepId,
   synchronizationJobContext,
   uploadConcurrency,
-  uploadBatchSize,
```
Gonzalo-Avalos-Ribas, line 145: I know this two are used in https://github.com/JupiterOne/jupiter-managed-integration/blob/1391f095c258cfe7f55e68d459c025623eba3351/packages/runtime/src/util/uploader.ts#L28 but shouldn't break anything, maybe we also want to cleanup there

Gonzalo-Avalos-Ribas (approved):
Looks good! 🔥

PR #967, opened 2023-09-28
[CORE-3497] remove hard cap on upload batch size in bytes

xdumaine:
With this, we can send through larger batch sizes from jupiter-managed-integration. First one targeted will be AWS via numeric feature toggle - https://app.launchdarkly.com/integrations/production/features/integration-aws-upload-batch-size-bytes/targeting


Gonzalo-Avalos-Ribas (commented):
Can we make sure to also change DEFAULT_GRAPH_OBJECT_BUFFER_THRESHOLD_IN_BYTES to 5_500_000 since even if we change DEFAULT_UPLOAD_BATCH_SIZE_IN_BYTES to 5_500_000 if we only hold 5MB in memory we will never upload more than that at once.

Gonzalo-Avalos-Ribas (approved):

PR #975, opened 2023-10-04
Move Vis to devDependencies

zemberdotnet:


packages/integration-sdk-cli/package.json
```diff
@@ -40,10 +40,10 @@
     "plop": "^3.1.2",
     "runtypes": "5.1.0",
     "upath": "^1.2.0",
-    "url-exists": "^1.0.3",
-    "vis": "^4.21.0-EOL"
```
VDubber, line 44: What is 'vis' currently being used for? The visualization of the graph? 
Moving it to a devDep will result in the package not being installed when?

VDubber (approved):
Approved with comment

packages/integration-sdk-cli/package.json
zemberdotnet: It's types are being used in a few functions. When a user runs the visualization the actual vis library gets imported via a <script> tag in generated html. This should stop bundling it when only it's types are being used.

PR #977, opened 2023-10-05
Don't flush after step completes

Gonzalo-Avalos-Ribas:
The motivation for this PR is to avoid flushing every time we complete a step. There should only be two times we flush to disk/upload: 

1. When we reach the graphObjectBufferThresholdInBytes of data in memory
2. All steps are done.

Also, I don't think is optimal to divide by step the data we update, that would generate way more unnecessary uploads.

Tried it on dev, on instance `jupiterone-integration-dev`: We went from 461 uploads in a single job, to 5.
The average batch size: Went up from 35.6 Kb to 3.28Mb and can be improved further by fine tuning the .env variables.

We are also introducing UPLOAD_ERROR that fails multiple steps at the same time. This was a bug with the previous sdk, where if a stepA generates an invalid entity that gets flushed with entities of stepB and stepC, the only step to fail would be stepA, but stepB and stepC also loose data


packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -460,6 +447,43 @@ export function executeStepDependencyGraph<
 
     void promiseQueue
       .onIdle()
+      .then(async () => {
+        /** Instead of flushing after each step, flush only when we finish all steps OR when we reach the threshold limit
+         * Because the 'createStepGraphObjectDataUploader' needs a step I'm using the last step as it
+         * I think we should decouple as much as possible upload from step success.
+         */
+        let uploader: StepGraphObjectDataUploader | undefined;
```
Gonzalo-Avalos-Ribas, line 455: There might be a smarter way of doing this

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -460,6 +447,43 @@ export function executeStepDependencyGraph<
...
+              ? uploader?.enqueue({
+                  entities: [],
+                  relationships,
+                })
+              : undefined,
+        );
+        try {
+          await uploader?.waitUntilUploadsComplete();
+        } catch (error) {
+          executionContext.logger.publishErrorEvent({
```
Gonzalo-Avalos-Ribas, line 480: How should we handle the error - this usually fails the step.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -245,101 +244,102 @@ export class FileSystemGraphObjectStore implements GraphObjectStore {
...
-
-            if (typeof indexMetadata === 'undefined') {
-              return true;
-            }
-
-            return indexMetadata.enabled === true;
+    await this.lockOperation(async () => {
+      const entitiesByStep = this.localGraphObjectStore.collectEntitiesByStep();
+      let entitiesToUpload: Entity[] = [];
+      for (const [stepId, entities] of entitiesByStep) {
```
zemberdotnet, line 250: Why did we move this from pMap to a plain function?

Gonzalo-Avalos-Ribas: Plain and simply there is no reason to divide the uploads we do. **There might be a point of doing it for writes to disk.**
But if a flushed batch of 6MB of entities gets into the function, do we really want to generate 100 uploads if the entities come from 100 different steps?

zemberdotnet: That makes sense. In that case, we may want to alter the way this whole flushing process is done in the future. We could just remove the concept of steps in this. That can be future work though.

zemberdotnet (dismissed):

zemberdotnet:
Looks good to me. Let's make an alpha version.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -233,113 +241,151 @@ export class FileSystemGraphObjectStore implements GraphObjectStore {
...
-              return true;
-            }
-
-            return indexMetadata.enabled === true;
+    await this.lockOperation(async () => {
+      if (!force) {
+        if (
+          this.localGraphObjectStore.getTotalEntitySizeInBytes() <
+          this.graphObjectBufferThresholdInBytes
+        ) {
```
dobregons, line 271:
```suggestion
// Do not flush entities to disk if we haven't reached the `graphObjectBufferThresholdInBytes` of data in memory or if we don't force this process (when all the steps are completed)
if (!force && this.localGraphObjectStore.getTotalEntitySizeInBytes() < this.graphObjectBufferThresholdInBytes) {
```

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -233,113 +241,151 @@ export class FileSystemGraphObjectStore implements GraphObjectStore {
...
-            return indexMetadata.enabled === true;
+    await this.lockOperation(async () => {
+      if (!force) {
+        if (
+          this.localGraphObjectStore.getTotalRelationshipSizeInBytes() <
+          this.graphObjectBufferThresholdInBytes
+        ) {
+          return;
+        }
+      }
```
dobregons, line 346: Same comment here.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -153,11 +152,17 @@ export class FileSystemGraphObjectStore implements GraphObjectStore {
         integrationStepsToGraphObjectIndexMetadataMap(params.integrationSteps);
     }
   }
+  getStepsStored() {
+    return this.localGraphObjectStore.getStepsStored();
+  }
```
dobregons, line 157: Can you point me where is this used and what is this used for? I don't see any code referencing this new method.

Gonzalo-Avalos-Ribas: Thanks! It was used in a previous version of this change and is no longer needed. Is now removed!

zemberdotnet:
@Gonzalo-Avalos-Ribas I think a lot of our changes in this were made to figure out which _types we should mark as partial if an upload fails. What if we just iterated over the entities/relationships after an upload fails and mark those as partial instead of trying to infer it from the step. I think it might simplify some of the other code where we could get rid of things like `stepsInvolvedInUploads`

Gonzalo-Avalos-Ribas:
> @Gonzalo-Avalos-Ribas I think a lot of our changes in this were made to figure out which _types we should mark as partial if an upload fails. What if we just iterated over the entities/relationships after an upload fails and mark those as partial instead of trying to infer it from the step. I think it might simplify some of the other code where we could get rid of things like `stepsInvolvedInUploads`

@zemberdotnet Actually the changes are for failing the steps that have graphObjects in the uploads that fail - But marking them as partials could also work. How do the steps show in the event logs if they are marked as partial? How can we mark an entity as partial during the execution?

packages/integration-sdk-core/src/types/step.ts
```diff
@@ -113,6 +113,11 @@ export type IntegrationStepResult = Omit<
    * Duration of the step in milliseconds, calculated as endTime - startTime.
    */
   duration?: number;
+  /**
+   * TODO: Decouple steps from uploads. This shouldn't be here.
+   * This are types that are involved in a failed upload. We should mark them as partial.
+   */
+  partialTypesToForce?: string[];
```
Gonzalo-Avalos-Ribas, line 120: I def don't like to have this information in the step result, but might be the only way.

packages/integration-sdk-runtime/src/execution/step.ts
```diff
@@ -142,6 +142,9 @@ export function determinePartialDatasetsFromStepExecutionResults(
         stepPartialDatasets.types.push(...stepResult.declaredTypes);
       }
 
+      if (stepResult.partialTypesToForce) {
+        stepPartialDatasets.types.push(...stepResult.partialTypesToForce);
+      }
```
Gonzalo-Avalos-Ribas, line 147: We add the types to the partial datasets.

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -171,7 +171,7 @@ export function executeStepDependencyGraph<
         startTime,
         endTime,
         duration,
-        partialTypesToForce: partialTypesToForce,
+        partialTypes: existingResult.partialTypes.concat(partialTypes ?? []),
```
zemberdotnet, line 174: When we assign the partial types, we should check they aren't already in the list so we don't duplicate them.

zemberdotnet (approved):

PR #978, opened 2023-10-09
Add Node 20 Support

zemberdotnet:
- node 20 update


packages/cli/src/export/__tests__/exportJsonAssetsToCsv.test.ts
```diff
@@ -203,7 +203,7 @@ test('should log error when export fails', async () => {
       storageDirectory: TEST_DIRECTORY,
       apiKey: TEST_API_KEY,
     }),
-  ).rejects.toThrow(/Unexpected token/g);
+  ).rejects.toThrow();
```
zemberdotnet, line 206: JSON parsing error message changes in later verison.

VDubber (dismissed):
I like keeping this moving forward. Could you message the team and maybe also update documentation with a preferred version to use? It's easier on everyone if a recommended version is known.

packages/cli/src/export/__tests__/exportJsonAssetsToCsv.test.ts
VDubber: I'm interested to see if it actually improved!

VDubber (approved):

PR #983, opened 2023-10-18
Delete data from upload errors

Nick-NCSU:


packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -518,11 +518,7 @@ export async function uploadDataChunk<
             uploadCorrelationId,
           });
         } catch (error) {
-          logger.warn(
-            { error },
-            'handleUploadDataChunkError function threw',
-            error,
-          );
+          logger.warn({ error }, 'handleUploadDataChunkError function threw');
```
VDubber, line 521: Nice clean up!

packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -420,7 +420,7 @@ function handleUploadDataChunkError({
    * }
    */
   const systemErrorResponseData = getSystemErrorResponseData(err);
-
+  delete err?.config?.data;
```
VDubber, line 423: Should this data continued to be logged but at a different level? Or is it never helpful for debugging?

VDubber (approved):

PR #988, opened 2023-11-16
Add step concurrency to dependency graph

Nick-NCSU:
This will give us the ability to limit the number of concurrent steps. Currently the value is the default (`POSITIVE_INFINITY`)

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
+++ b/packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
@@ -82,6 +82,7 @@ export function executeStepDependencyGraph<
   executionContext,
   inputGraph,
   stepStartStates,
+  stepConcurrency,
   duplicateKeyTracker,
   graphObjectStore,
   dataStore,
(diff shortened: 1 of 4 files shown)


packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -114,7 +116,11 @@ export function executeStepDependencyGraph<
   const workingGraph = inputGraph.clone();
 
   // create a queue for managing promises to be executed
-  const promiseQueue = new PromiseQueue();
+  const promiseQueueOptions: { concurrency?: number } = {};
+  if (stepConcurrency !== undefined) {
+    promiseQueueOptions.concurrency = stepConcurrency;
+  }
+  const promiseQueue = new PromiseQueue(promiseQueueOptions);
```
Nick-NCSU, line 123: PromiseQueue does not allow you to send `{ concurrency: undefined }` so we only put it on the options if we have a value.

RonaldEAM (approved):

JakeFerrero (approved):

packages/integration-sdk-runtime/src/execution/dependencyGraph.ts
```diff
@@ -114,7 +116,11 @@ export function executeStepDependencyGraph<
   const workingGraph = inputGraph.clone();
 
   // create a queue for managing promises to be executed
-  const promiseQueue = new PromiseQueue();
+  const promiseQueueOptions: { concurrency?: number } = {};
+  if (stepConcurrency !== undefined) {
+    promiseQueueOptions.concurrency = stepConcurrency;
+  }
```
VDubber, line 122: Consider this approach for readability:

```
const promiseQueueOptions = {
   ...(stepConcurrency && { concurrency: stepConcurrency })
}

```

PR #996, opened 2023-12-07
Remove skipped config logs from job logs

Nick-NCSU:


i5o (approved):

Nick-NCSU:
This will not fully work, we need to also hide messages of steps that only depend on these steps as well.

JakeFerrero:
We have a different approach we are going to do - we've decided that skipped step logs make sense to be warnings for other integrations (specifically GCP). But since we have received so many complaints about the yellow warning text for skipped jobs in AWS, we are going to just set the skipped step logs to an INFO level in the UI _just for the aws integration_. This is a temporary solution, as the UI will be receiving a facelift that will include removing job logs entirely. This solution should hopefully alleviate customer concerns until the facelift is done.

PR #999, opened 2023-12-14
upload data improvements

zemberdotnet:
Improves logging for uploadData


packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -590,3 +598,13 @@ export async function abortSynchronization({
 
   return abortedJob;
 }
+
+function cleanAxiosError(err: AxiosError) {
```
VDubber, line 602: I have seen logged Axios errors that are huge.  Are we able to reduce logging to the valuable parts?

VDubber (approved):

PR #1001, opened 2023-12-15
INT-9047 Make Summary.json available (to TAM or users)

eyadmba:
The overall feature is better explained here: https://github.com/JupiterOne/callisto/pull/418

The change done here allows a function to be called to do something with the results of an integration job.
That's needed for this feature so that the managed runtime can use that callback function to store the job's output in S3 (https://github.com/JupiterOne/jupiter-managed-integration/pull/353).

---

Jira Ticket: [INT-9047 Make Summary.json available (to TAM or users)
](https://jupiterone.atlassian.net/browse/INT-9047)

### DIFF
--- a/packages/integration-sdk-runtime/src/execution/executeIntegration.ts
+++ b/packages/integration-sdk-runtime/src/execution/executeIntegration.ts
@@ -62,13 +62,14 @@ export interface ExecuteIntegrationOptions {
   enableSchemaValidation?: boolean;
   graphObjectStore?: GraphObjectStore;
   createStepGraphObjectDataUploader?: CreateStepGraphObjectDataUploaderFunction;
+  resultsCallback?: (results: IntegrationStepResult[]) => Promise<void>;
   pretty?: boolean;
 }
 
-export interface ExecuteWithContextOptions {
-  graphObjectStore?: GraphObjectStore;
-  createStepGraphObjectDataUploader?: CreateStepGraphObjectDataUploaderFunction;
-}
+type ExecuteWithContextOptions = Pick<
+  ExecuteIntegrationOptions,
+  'graphObjectStore' | 'resultsCallback' | 'createStepGraphObjectDataUploader'
+>;
 
 const THIRTY_SECONDS_STORAGE_INTERVAL_MS = 60000 / 2;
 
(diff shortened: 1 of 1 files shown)


packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -62,13 +62,14 @@ export interface ExecuteIntegrationOptions {
...
 }
 
-export interface ExecuteWithContextOptions {
-  graphObjectStore?: GraphObjectStore;
-  createStepGraphObjectDataUploader?: CreateStepGraphObjectDataUploaderFunction;
-}
+type ExecuteWithContextOptions = Pick<
+  ExecuteIntegrationOptions,
+  'graphObjectStore' | 'resultsCallback' | 'createStepGraphObjectDataUploader'
+>;
```
eyadmba, line 72: Seems less error-prone than duplicating the interface.

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -288,6 +290,17 @@ export async function executeWithContext<
         data: summary,
       });
 
+      if (resultsCallback != null) {
+        try {
+          await resultsCallback(integrationStepResults);
+        } catch (err) {
+          context.logger.warn(
```
eyadmba, line 297: Do we have a convention for appropriate log levels?

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -288,6 +290,17 @@ export async function executeWithContext<
         data: summary,
       });
 
+      if (resultsCallback != null) {
+        try {
+          await resultsCallback(integrationStepResults);
```
dobregons, line 295: The summary.json that we store locally is something like this:
```typescript
{
  integrationStepResults,
  metadata: {
    partialDatasets,
  },
}
```
I wonder if we should add the `metadata` object, same as the existing version?

eyadmba: Maybe I was over-engineering, but I thought the `resultsCallback` should provide the raw results, and then the function is responsible for summarizing it or whatever.

But if we don't think that's useful I can just pass in the `summary` instead.
Or I can also have the function that stores in S3 in the managed runtime run the summarization function before it sends it to S3.

dobregons: I think the current version is fine, if we need to change it we can do it later, but IMO it's fine!

dobregons (approved):

PR #1002, opened 2023-12-19
Add validateValueType function to ensure data type integrity

i5o:
Added the validateValueType function, significantly enhancing the robustness of our data handling in the createIntegrationEntity process.

The function is designed to enforce strict type checking, ensuring that all data conforms to our predefined SUPPORTED_TYPES.

This only applies for createIntegrationEntity, we already have types for `assign` but IMO it's never too late to validate anyway

### DIFF
--- a/packages/integration-sdk-core/src/data/createIntegrationEntity.ts
+++ b/packages/integration-sdk-core/src/data/createIntegrationEntity.ts
@@ -10,7 +10,7 @@ import { parseTimePropertyValue } from './converters';
 import { validateRawData } from './rawData';
 import { assignTags, ResourceTagList, ResourceTagMap } from './tagging';
 
-const SUPPORTED_TYPES = ['string', 'number', 'boolean'];
+const SUPPORTED_TYPES = ['string', 'number', 'boolean', 'array', 'undefined'];
 
 /**
  * Properties to be assigned to a generated entity which are declared in code
@@ -180,6 +180,40 @@ function generateEntity({
 }
 
 /**
+ * Validates that the provided value conforms to the supported types.
+ * If the value is an array, this function is called recursively on each element
+ * to ensure nested arrays are also validated. This function throws an error
+ * if it encounters a value type that is not supported.
+ *
+ * @param value The value to be validated. It can be a single value or an array.
+ *              For arrays, each element is validated recursively.
+ * @param path The path to the current property being validated, used for error messaging.
+ *             This is updated with each recursive call to reflect the current context.
+ */
+export function validateValueType(value: any, path: string): void {
+  // Explicitly allow null values
+  if (value === null) {
+    return;
+  }
+
+  if (Array.isArray(value)) {
+    // If the value is an array, validate each element
+    value.forEach((item, index) => {
+      validateValueType(item, `${path}[${index}]`);
+    });
+  } else {
+    // For non-array values, check if the type is supported
+    const valueType = typeof value;
+    if (!SUPPORTED_TYPES.includes(valueType)) {
+      throw new IntegrationError({
+        code: 'UNSUPPORTED_TYPE',
+        message: `Unsupported type found at "${path}": ${valueType}`,
+      });
+    }
+  }
+}
+
+/**
  * Answers a form of the provider data with only the properties supported by the
  * data model schema.
  *
@@ -192,12 +226,14 @@ function whitelistedProviderData(
 ): Omit<ProviderSourceData, 'tags'> {
   const whitelistedProviderData: ProviderSourceData = {};
   const schemaProperties = schemaWhitelistedPropertyNames(_class);
+
   for (const [key, value] of Object.entries(source)) {
+    // Ensure the property is part of the schema and not null
     if (value != null && schemaProperties.includes(key)) {
-      const valueType = Array.isArray(value) ? typeof value[0] : typeof value;
-      if (SUPPORTED_TYPES.includes(valueType)) {
-        whitelistedProviderData[key] = value;
-      }
+      if (key != 'tags') validateValueType(value, key);
+
+      // If validation passes, assign the value to the whitelisted data
+      whitelistedProviderData[key] = value;
     }
   }
   return whitelistedProviderData;
--- a/packages/integration-sdk-core/src/data/__tests__/createIntegrationEntity.test.ts
+++ b/packages/integration-sdk-core/src/data/__tests__/createIntegrationEntity.test.ts
@@ -5,6 +5,7 @@ import {
   IntegrationEntityData,
   schemaWhitelistedPropertyNames,
   schemaWhitelists,
+  validateValueType,
 } from '../createIntegrationEntity';
 
 const networkSourceData = {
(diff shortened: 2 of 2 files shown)


xdumaine:
It might be worth adding a couple scale tests to this to ensure that running on tens or hundreds of thousands of values doesn't significantly impact performance. You don't have to include those on every ci run, but they're good to have to validate. We did that with normalization in persister and caught a couple of important bugs that would have caused significant slow downs.

i5o:
@xdumaine I've added tests for 500k of items, they run properly on my M1 Pro laptop.

```
 PASS  packages/integration-sdk-core/src/data/__tests__/validateValueType.test.ts
  ● Console

    console.log
      Execution time for string dataset: 62.04570806026459 ms

      at Object.log (src/data/__tests__/validateValueType.test.ts:80:15)

    console.log
      Execution time for number dataset: 66.66854095458984 ms

      at Object.log (src/data/__tests__/validateValueType.test.ts:80:15)

    console.log
      Execution time for boolean dataset: 66.96124994754791 ms

      at Object.log (src/data/__tests__/validateValueType.test.ts:80:15)

    console.log
      Execution time for array dataset: 278.7055000066757 ms

      at Object.log (src/data/__tests__/validateValueType.test.ts:80:15)

    console.log
      Execution time for undefined dataset: 68.97012507915497 ms

      at Object.log (src/data/__tests__/validateValueType.test.ts:80:15)

    console.log
      Execution time until error: 0.13349997997283936 ms

      at Object.log (src/data/__tests__/validateValueType.test.ts:118:15)
      ````

**Comment by i5o** [comment]
Added `[DO NOT MERGE]` to review after holidays

File: packages/integration-sdk-core/src/data/createIntegrationEntity.ts [hunk 1]
```diff
@@ -10,7 +10,7 @@ import { parseTimePropertyValue } from './converters';
 import { validateRawData } from './rawData';
 import { assignTags, ResourceTagList, ResourceTagMap } from './tagging';
 
-const SUPPORTED_TYPES = ['string', 'number', 'boolean'];
+const SUPPORTED_TYPES = ['string', 'number', 'boolean', 'array', 'undefined'];
```
Nick-NCSU, line 13: Wouldn't this allow Arrays of arrays? I don't think we need to allow array here since we have a check for that already.
```suggestion
const SUPPORTED_TYPES = ['string', 'number', 'boolean',  'undefined'];
```

Nick-NCSU: Also isn't typeof [] === 'object' ?

zemberdotnet:
I think this approach is both valid and a good incremental step forward. It solves a problem we have. I do think that it may be worth considering if using a JSON schema validator on the entities and properties would not provide a more battle-tested form of data validation. We already make use of some JSON Schema validation during development and with better integration I think it could be a powerful safeguard in our platform.

Additionally, JSON schemas would help with search and entity property documentation, a current and longstanding problem.

packages/integration-sdk-core/src/data/createIntegrationEntity.ts
i5o: > Also isn't typeof [] === 'object' ?

That's why we have Array.isArray

i5o: https://github.com/JupiterOne/sdk/pull/1002/files#diff-4d79fa9d80062cd821c28d4014d267e94d76e24d49610be304dc8513b06981b8R199

Array are also allowed.
You are right about array inside arrays.

packages/integration-sdk-core/src/data/createIntegrationEntity.ts
```diff
@@ -179,6 +179,52 @@ function generateEntity({
...
+ *
+ * @param value The value to be validated. It can be a single value or an array.
+ *              For arrays, each element is validated recursively.
+ * @param path The path to the current property being validated, used for error messaging.
+ *             This is updated with each recursive call to reflect the current context.
+ */
+export function validateValueType(
+  value: any,
+  path: string,
+  depth: number = 1,
```
VDubber, line 196: nit: `depth` is missing from the jsdoc. Consider a rename to `currentValueDepth`

packages/integration-sdk-core/src/data/createIntegrationEntity.ts
```diff
@@ -192,12 +238,14 @@ function whitelistedProviderData(
 ): Omit<ProviderSourceData, 'tags'> {
   const whitelistedProviderData: ProviderSourceData = {};
```
VDubber, line 239: nit: avoid white/black lists. allowedProviderData

packages/integration-sdk-core/src/data/createIntegrationEntity.ts
```diff
@@ -179,6 +179,52 @@ function generateEntity({
...
+
+    // If the value is an array, validate each element
+    value.forEach((item, index) => {
+      validateValueType(item, `${path}[${index}]`, ++depth);
+    });
+  } else {
+    // For non-array values, check if the type is supported
+    const valueType = typeof value;
+    if (!SUPPORTED_TYPES.includes(valueType)) {
+      throw new IntegrationError({
```
VDubber, line 220: How can we keep steps from failing on unsupported data? This is a step towards strictness that'll have any overall impact on job statuses and the ingestion team.
How else could we improve the overall health of jobs and data?

Nick-NCSU:
Can we simplify this and avoid recursive calls with something like

validate(entity: Entity): void {
  for(const [key, value] of Object.entries(entity)) {
    const valid = Array.isArray(property) ? property.every(p => SUPPORTED_TYPES.includes(typeof p)) : SUPPORTED_TYPES.includes(typeof property);
    valid || throw new IntegrationError({
        code: 'UNSUPPORTED_TYPE',
        message: `Unsupported type found at "${path}": Nested arrays are not supported.`,
      });
  }
}

i5o:
@Nick-NCSU tried adding your code into my existing codebase but couldn't succeed, feel free to push to this branch!

packages/integration-sdk-core/src/data/createIntegrationEntity.ts
i5o: We should open a bigger discussion for this but this should be a MAJOR, we have some ideas in the AWS Pod as to how to proceed

packages/integration-sdk-core/src/data/createIntegrationEntity.ts
i5o: Renamed properties included function name allowedProviderData

packages/integration-sdk-core/src/data/createIntegrationEntity.ts
```diff
@@ -125,7 +125,7 @@ function generateEntity({
   const _class = Array.isArray(assign._class) ? assign._class : [assign._class];
 
   const entity: GeneratedEntity = {
-    ...(source ? whitelistedProviderData(source, _class) : {}),
+    ...(source ? allowedProviderData(source, _class), : {}),
```
i5o, line 128:
```suggestion
    ...(source ? allowedProviderData(source, _class) : {}),
```

PR #1011, opened 2024-01-09
Allow deletion of required properties from schema

Nick-NCSU:
To avoid the use of disableClassMatch I have added in a way to remove required properties from the schema and will be applied at the end of the processing of the schema to ensure it is not re-added.

We already have the functionality to override the data-model type properties but not the required array.

### DIFF
--- a/packages/integration-sdk-testing/src/__tests__/jest.test.ts
+++ b/packages/integration-sdk-testing/src/__tests__/jest.test.ts
@@ -358,7 +358,7 @@ describe('#toMatchGraphObjectSchema', () => {
             items: { type: 'object' },
           },
           displayName: { type: 'string' },
-          version: { delete: true },
+          version: { exclude: true },
         },
       },
     });
--- a/packages/integration-sdk-testing/src/__tests__/jest.test.ts
+++ b/packages/integration-sdk-testing/src/__tests__/jest.test.ts
@@ -296,6 +296,80 @@ describe('#toMatchGraphObjectSchema', () => {
     });
   });
 
+  test('should remove properties marked for deletion from schema', () => {
+    const standardEntity = {
+      _key: 'arn:aws:securityhub:us-east-1::standards/aws-foundational-security-best-practices/v/1.0.0',
+      _type: 'aws_securityhub_standard',
+      _class: ['Standard'],
+      name: 'AWS Foundational Security Best Practices v1.0.0',
+      displayName: 'AWS Foundational Security Best Practices v1.0.0',
+      enabledByDefault: true,
+      managingCompany: 'AWS',
+      managingProduct: 'Security Hub',
+      region: 'us-east-1',
+      _rawData: [
+        {
+          name: 'default',
+          rawData: {
+            EnabledByDefault: true,
+            Name: 'AWS Foundational Security Best Practices v1.0.0',
+            StandardsArn:
+              'arn:aws:securityhub:us-east-1::standards/aws-foundational-security-best-practices/v/1.0.0',
+            StandardsManagedBy: {
+              Company: 'AWS',
+              Product: 'Security Hub',
+            },
+          },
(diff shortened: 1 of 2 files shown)


packages/integration-sdk-testing/src/__tests__/jest.test.ts
```diff
@@ -296,6 +296,80 @@ describe('#toMatchGraphObjectSchema', () => {
...
+              Company: 'AWS',
+              Product: 'Security Hub',
+            },
+          },
+        },
+      ],
+    };
+
+    const failedResult = toMatchGraphObjectSchema(standardEntity, {
+      _class: ['Standard'],
```
Nick-NCSU, line 328: The 'Standard' class has `Version` as a required field.

packages/integration-sdk-testing/src/jest.ts
```diff
@@ -286,6 +286,45 @@ function dedupSchemaEnumValues(schema: GraphObjectSchema): GraphObjectSchema {
   };
 }
 
+/**
+ * Step to remove specified properties marked for deletion
+ * in the schemas. This step will run after all deduplication
+ * processes to ensure that the properties will be removed from
+ * the final schema.
+ *
+ * Example: parameter: { deleted: true }
```
i5o, line 295:
```suggestion
 * Example: parameter: { delete: true }
```

i5o (changes requested):

packages/integration-sdk-testing/src/__tests__/jest.test.ts
```diff
@@ -296,6 +296,80 @@ describe('#toMatchGraphObjectSchema', () => {
...
+      schema: {
+        additionalProperties: true,
+        properties: {
+          _type: { const: 'aws_securityhub_standard' },
+          _rawData: {
+            type: 'array',
+            items: { type: 'object' },
+          },
+          displayName: { type: 'string' },
+          version: { delete: true },
```
i5o, line 361:
```suggestion
          version: { delete: true },
```

```suggestion
          version: { exclude: true },
```

packages/integration-sdk-testing/src/jest.ts
i5o: I suggest we use "exclude" or "emptyAllowed: true"

packages/integration-sdk-testing/src/__tests__/jest.test.ts
Nick-NCSU: I like this better

JakeFerrero: "This" as in delete or exclude lol? I actually think I like exclude.

Nick-NCSU: > "This" as in delete or exclude lol? I actually think I like exclude.
Exclude :)

packages/integration-sdk-testing/src/jest.ts
```diff
@@ -286,6 +286,45 @@ function dedupSchemaEnumValues(schema: GraphObjectSchema): GraphObjectSchema {
...
+  schema: GraphObjectSchema,
+): GraphObjectSchema {
+  if (!schema.properties) {
+    return schema;
+  }
+
+  const newProperties: Record<string, any> = {};
+
+  for (const propertyName in schema.properties) {
+    const property = schema.properties[propertyName];
```
JakeFerrero, line 309: couldn't you do `for (const [propertyName, property] of Object.entries(schema.properties)` above so you don't need this line?

Nick-NCSU: Yes but I copied the implementation from the other schema modifying functions and wanted to be consistent.

JakeFerrero (approved):

PR #1013, opened 2024-01-09
Add option to require spec for step

Nick-NCSU:


packages/integration-sdk-testing/src/jest.ts
```diff
@@ -648,19 +648,72 @@ export function toImplementSpec<
...
   spec: IntegrationSpecConfig<TConfig>,
+  options?: {
+    requireSpec?: boolean;
+  },
 ) {
   const unimplementedSteps: string[] = [];
 
+  // Normalize the spec and integration objects
+  // as the order does not matter for the test
+  const normalizedSpec: IntegrationSpecConfig<TConfig> = {
```
Nick-NCSU, line 659: Bonus change: The order of `entities` `relationships` `mappedRelationships` and `dependsOn` arrays does not matter so we will sort them before comparing to ease development.

i5o (approved):

zemberdotnet (approved):

PR #1016, opened 2024-01-11
Automatically include _rawData in schema for entities

Nick-NCSU:
Currently when validating graph object's schema with `additionalProperties: false` you are required to add _rawData as a property ([As seen here](https://github.com/JupiterOne/graph-google-cloud/blob/597eb7bd798f4406d39f054fcdbff38040a040da/src/steps/service-usage/constants.ts#L28-L31))

This will remove automatically add the property to the schema to prevent this need.

### DIFF
--- a/packages/integration-sdk-testing/src/jest.ts
+++ b/packages/integration-sdk-testing/src/jest.ts
@@ -608,7 +608,11 @@ function toMatchSchema<T extends Entity | ExplicitRelationship>(
   received = Array.isArray(received) ? received : [received];
 
   for (let i = 0; i < received.length; i++) {
-    const data = received[i];
+    // Removes raw data before schema validation
+    const data = {
+      ...received[i],
+    };
+    delete data._rawData;
 
     if (dataModelIntegrationSchema.validate(schema, data)) {
       continue;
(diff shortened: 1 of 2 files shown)


i5o (approved):

packages/integration-sdk-testing/src/jest.ts
```diff
@@ -608,7 +608,11 @@ function toMatchSchema<T extends Entity | ExplicitRelationship>(
   received = Array.isArray(received) ? received : [received];
 
   for (let i = 0; i < received.length; i++) {
-    const data = received[i];
+    // Removes raw data before schema validation
+    const data = {
+      ...received[i],
+    };
+    delete data._rawData;
```
zemberdotnet, line 615: const {
   _rawData,
   ...data
} = received[i];
Let's not mutate the object.

zemberdotnet: Oh I guess it is a copy of the object. Should be fine then.

Nick-NCSU: Yeah it would have only removed it from data since it was a shallow copy

PR #1020, opened 2024-02-01
Add ability to return encountered keys in job output

Nick-NCSU:
Screenshot 2024-02-01 at 10 56 49 AM

### DIFF
--- a/packages/integration-sdk-core/src/types/config.ts
+++ b/packages/integration-sdk-core/src/types/config.ts
@@ -83,10 +83,18 @@ export interface InvocationConfig<
   /**
    * Optionally define the number of steps to execute in parallel.
    *
-   * If this value is not provided the concurrency defaults to
+   * If this value is not provided, the concurrency defaults to
    * Number.POSITIVE_INFINITY
    */
   stepConcurrency?: number;
+  /**
+   * Optionally collect the encountered keys and return them in
+   * the integration results. This can result in a large quantity
+   * of keys being returned.
+   *
+   * If this value is not provided, the default is false.
+   */
+  collectEncounteredKeys?: boolean;
   /**
    * This configuration element is used to store information about data
    * ingestion sources that can be enabled or disabled. When this element
--- a/packages/integration-sdk-runtime/src/execution/executeIntegration.ts
+++ b/packages/integration-sdk-runtime/src/execution/executeIntegration.ts
@@ -53,6 +53,7 @@ import { getIngestionSourceStepStartStates } from './utils/getIngestionSourceSte
 
 export interface ExecuteIntegrationResult {
   integrationStepResults: IntegrationStepResult[];
+  encounteredKeys?: string[][];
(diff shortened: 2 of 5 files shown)


i5o (approved):

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -292,7 +298,7 @@ export async function executeWithContext<
 
       if (resultsCallback != null) {
         try {
-          await resultsCallback(integrationStepResults);
+          await resultsCallback(summary);
```
zemberdotnet, line 301: Any tests we can add/update for this feature?

packages/integration-sdk-core/src/types/config.ts
```diff
@@ -83,10 +83,18 @@ export interface InvocationConfig<
...
    */
   stepConcurrency?: number;
+  /**
+   * Optionally collect the encountered keys and return them in
+   * the integration results. This can result in a large quantity
+   * of keys being returned.
+   *
+   * If this value is not provided, the default is false.
+   */
+  collectEncounteredKeys?: boolean;
```
eyadmba, line 97: How is this set? Do we have to re-deploy the integration when we want to enable this?

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
```diff
@@ -62,7 +63,7 @@ export interface ExecuteIntegrationOptions {
   enableSchemaValidation?: boolean;
   graphObjectStore?: GraphObjectStore;
   createStepGraphObjectDataUploader?: CreateStepGraphObjectDataUploaderFunction;
-  resultsCallback?: (results: IntegrationStepResult[]) => Promise<void>;
+  resultsCallback?: (results: ExecuteIntegrationResult) => Promise<void>;
```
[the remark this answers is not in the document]
eyadmba: I should've passed the `ExecuteIntegrationResult` anyway from the start tbh. I don't usually use lists as the parameter for a public interface. My bad.

packages/integration-sdk-core/src/types/config.ts
Nick-NCSU: This would be set in the invocationConfig which can be loaded dynamically with the loadInvocationConfig hook. We can toggle this at runtime this way.

packages/integration-sdk-runtime/src/execution/executeIntegration.ts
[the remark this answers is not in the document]
Nick-NCSU: I am updating the runtime to allow any to be sent to storeJobOutputJsonInS3 since it should be able to take anything serializable. Now that you mention it that would require the 2 to be updated together. (or at least upgrading the sdk would require the managed sdk to be updated first)

JakeFerrero (approved):

PR #1025, opened 2024-02-12
Add iterate API function

gastonyelmini:


packages/integration-sdk-http-client/src/index.ts
```diff
@@ -244,10 +244,84 @@ export abstract class BaseAPIClient {
...
+      const response = await this.request(
+        nextRequestQuery
+          ? `${initialRequest.endpoint}${this.buildQueryParams(
+              nextRequestQuery
+            )}`
+          : initialRequest.endpoint,
+        nextRequestOptions ? nextRequestOptions : initialRequest.options
+      );
+
+      const cbOptions = cb(response as T);
```
RonaldEAM, line 314: You don't need a generic in the function at all if you are returning the `Response` object (which we should, so the package clients have more control), so instead of `cb: (iterationResources: T) => {` it should be `cb: (response: Response) => {`
Unfortunately the Response type doesn't accept generics

PR #1044, opened 2024-03-14
http-client: add method to override default response parse in paginate

RonaldEAM:


packages/integration-sdk-http-client/src/client.ts
```diff
@@ -398,6 +398,10 @@ export abstract class BaseAPIClient {
     } while (nextUrl);
   }
 
+  protected async parseResponseInPaginate(response: Response): Promise<any> {
+    return await response.json();
```
gastonyelmini, line 402: nit: we could avoid having two awaits (it ends up being the same anyways) and just retuning `return response.json();` and the function could not be `async` and only just doing await in line 382

gastonyelmini (approved):

gastonyelmini (approved):

PR #1054, opened 2024-04-02
Delete unused files on bocchi

i5o:


Nick-NCSU (approved):

JakeFerrero (changes requested):
Bocchi can be used for both public and private graph repos. As such, I’d keep the files and then as an enhancement, we can have a “public” option or something in bocchi that will omit those files when creating everything.

So I think the only file here that really makes sense to go is pull_request_template. But like I said, I think this PR has given me a good idea for an enhancement to bocchi.

packages/integration-sdk-cli/src/bocchi/templates/top-level/.github/pull_request_template.md.hbs
```diff
@@ -1,17 +0,0 @@
-# Description
```
JakeFerrero, line 1: This can definitely go

packages/integration-sdk-cli/src/bocchi/templates/top-level/cortex.yaml.hbs
```diff
@@ -0,0 +1,14 @@
...
+  title: '@jupiterone/{{packageName}}'
+  description: A graph conversion tool for the {{packageName}}
+  x-cortex-git:
+    github:
+      repository: JupiterOne/integration-deployments
+      basepath: graph-private/{{packageName}}
+  x-cortex-owners:
+    - type: group
+      name: JupiterOne/integrations
+  x-cortex-tag: '@jupiterone/{{packageName}}
```
i5o, line 12:
```suggestion
  x-cortex-tag: '@jupiterone/{{packageName}}'
```

PR #1059, opened 2024-04-10
[APP-15115] - Add integration-sdk-entity-validator package

geovannimp:


packages/integration-sdk-entity-validator/src/__tests__/validator.test.ts
```diff
@@ -0,0 +1,121 @@
...
+  },
+  required: ['_key', '_class', '_type'],
+};
+
+describe('validator', () => {
+  let entityValidator: EntityValidator;
+  let classSchemas: AnySchema[] = [];
+
+  beforeAll(async () => {
+    classSchemas = await fetch(RESOLVED_SCHEMAS_URL)
```
geovannimp, line 53: The idea is to export a function in the dataModel that does this and returns the latest data model version.

packages/integration-sdk-entity-validator/src/__tests__/validator.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      errors: null,
+      validationType: 'type',
+    });
+  });
+
+  test('Should validate with type if both entity type and class schemas are loaded', async () => {
+    const graphObjectSchema = classSchemas.find(
+      (s) => typeof s === 'object' && s.$id === '#GraphObject',
+    );
+    entityValidator.addSchemas(graphObjectSchema!);
```
geovannimp, line 107: Just loading a single schema cause we have some duplicated required keys that ajv is complaining. As well is kinda a nice idea loading just the schemas that will be used, cause make it faster.

PR to fix duplicated required keys: https://github.com/JupiterOne/data-model/pull/200/files

packages/integration-sdk-entity-validator/src/validator.ts
```diff
@@ -0,0 +1,66 @@
+import Ajv, { AnySchema, ErrorObject } from 'ajv';
+import addFormats from 'ajv-formats';
+import { addJ1Formats } from './j1Formats';
+
+export class EntityValidator {
+  private ajvInstance: Ajv;
+
+  constructor({ schemas }: { schemas?: AnySchema[] }) {
+    this.ajvInstance = addJ1Formats(
```
eyadmba, line 9: I have a proposal for a different interface of this package, where this package provides the following functions:
- Fetch the _class schemas, e.g. `async fetchClassSchemas(): JsonSchemas[]`
- Fetch _type schemas of a certain integration, e.g. `async fetchTypeSchemas(integrationName: string): JsonSchemas[]`
- Run validation on an object (not `Entity`), with a passed in Ajv instance e.g.
```typescript
validateEntity(ajv: Ajv, obj: object): ValidationResult
```

This package **would not** have its own Ajv instance. That would instead be handled by the consumer, which allows them to configure the instance, and also handle its lifecycle (re-use it, or create a new one if they want).

What I'm proposing here does leak the detail of using Ajv to the consumer heavily, so you could say we're actually making an "Ajv utils" library not a generic validation library, but I think that's fine.

The reason for this drastic change of the API is that I don't think the validation logic can be abstracted away cleanly and promise the consumer that they don't need to know about how we actually run the validation.
Fetching the schemas is something the consumer will need to know about, especially fetching _type schemas, they might want to cache that or manage it in whatever way makes sense for them.
In addition, the results of the Ajv validator are not pleasant to work with. If we want to expose the Ajv validation result to the consumer, then we haven't really done anything useful to the consumer. They might as well just use Ajv by themselves without the need for this library.

ryan-willis (approved):

packages/integration-sdk-entity-validator/src/validator.ts
```diff
@@ -0,0 +1,69 @@
...
+      }
+    }
+
+    return {
+      isValid: errors.length === 0,
+      errors: errors.length
+        ? errors.map(ajvErrorToEntityValidationError)
+        : null,
+      type: typeValidator ? 'type' : 'class',
+    };
```
eyadmba, line 67: I like this. It's a good summary of what the function did.

packages/integration-sdk-entity-validator/src/validator.ts
```diff
@@ -0,0 +1,69 @@
...
+      if (!isValid) {
+        errors = [...errors, ...(typeValidator.errors ?? [])];
+      }
+    } else {
+      for (const className of classArray) {
+        const validator = this.ajvInstance.getSchema(`#${className}`);
+
+        if (!validator) {
+          throw new Error(`Schema not found for class "${className}"`);
+        }
```
eyadmba, line 51: ...which is why I don't like this. Throwing errors from inside a for loop is too "aggressive".
Can't we also have this information returned?
If we don't find a _class it makes sense to me that it would just be skipped, instead of halting the entire validation.

packages/integration-sdk-entity-validator/src/validator.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  addSchemas(schema: AnySchema | AnySchema[]) {
+    this.ajvInstance.addSchema(schema);
+    return this;
+  }
+
+  validateEntity<Entity extends { _class: string[] | string; _type: string }>(
+    entity: Entity,
+  ) {
```
eyadmba, line 31: Do we want to change the parameter type to not expect an `Entity` type?

packages/integration-sdk-entity-validator/src/entityValidationError.ts
```diff
@@ -0,0 +1,38 @@
...
+  message: string;
+  validation: string;
+};
+
+const getPropertyNameFromAjvError = (
+  error: ErrorObject,
+): EntityValidationError['property'] => {
+  // instancePath is either '/{propertyName}', or if it's a list: '/{propertyName}/index'
+  // instancePath will always start with a slash
+  return error.instancePath.split('/')[1] || unknownPropertySymbol;
```
eyadmba, line 16: "instancePath will always start with a slash" does that also apply even if the error is an object-level error?

I think it does, but I can't be certain.

Because if there's ever an error where `instancePath == ""`, grabbing the index 1 of the split() will throw an error.

packages/integration-sdk-entity-validator/src/validator.ts
eyadmba: `ajvErrorToEntityValidationError()` is much cleaner than what I had. Nice work.

packages/integration-sdk-entity-validator/src/entityValidationError.ts
geovannimp: JS don't throw exceptions to "invalid" index because all arrays have "Infinite" size.
image

packages/integration-sdk-entity-validator/src/validator.ts
geovannimp: That a really good point. I'm going to return a list of validated and not validated classes and let the consumer handle if it should throw or not the exception.

packages/integration-sdk-entity-validator/src/validator.ts
geovannimp: I forgot about that. Let me fix it.

packages/integration-sdk-entity-validator/src/validator.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+  }
+
+  addSchemas(schema: AnySchema | AnySchema[]) {
+    this.ajvInstance.addSchema(schema);
+    return this;
+  }
+
+  validateEntity(
+    entity: object,
```
eyadmba, line 76: How about we keep all the runtime checking you have, but change the type of the `entity` param to be `{_class: string; _type: string}` to have typescript help the consumer do the right thing?

packages/integration-sdk-entity-validator/src/validator.ts
```diff
@@ -0,0 +1,158 @@
...
+        }
+      } else {
+        skippedSchemas.push({
+          schemaId: typeSchemaId,
+          reason: 'not-found',
+          type: 'type',
+        });
+      }
+
+      if (!typeValidator || forceClassValidationWithValidatedType) {
```
eyadmba, line 112: what's the use-case for `forceClassValidationWithValidatedType`?

packages/integration-sdk-entity-validator/src/validator.ts
```diff
@@ -0,0 +1,158 @@
...
+    entity: object,
+    {
+      forceClassValidationWithValidatedType = false,
+    }: ValidateEntityOptions = {},
+  ): ValidateEntityResult {
+    const errors: EntityValidationError[] = [];
+    const skippedSchemas: SkippedSchema[] = [];
+
+    try {
+      assertEntity(entity);
```
eyadmba, line 85: I don't know if it's possible, but I'll say it anyway and leave it to you to decide. The nesting is making the code hard to follow, so if possible, it would be better to flatten out this function a little bit.

packages/integration-sdk-entity-validator/src/validator.ts
```diff
@@ -0,0 +1,158 @@
...
+        );
+      }
+    } catch (error) {
+      errors.push(convertUnknownErrorToEntityValidationError(error));
+    }
+
+    return {
+      isValid: errors.length === 0,
+      errors: errors.length ? errors : null,
+      skippedSchemas: skippedSchemas.length ? skippedSchemas : null,
```
eyadmba, line 155: We lost the information of if we used the `_type` vs `_class` for the validation. Is that deliberate?

eyadmba (approved):
I left some comments and questions, but overall it looks good to me.

packages/integration-sdk-entity-validator/src/validator.ts
geovannimp: I'm not sure, that because it's false by default. But you may want to validate the latest class schema as well cause the type schema has a outdated class schema merged in it.

packages/integration-sdk-entity-validator/src/validator.ts
geovannimp: With the current implementation you could have both validations. You can get this information now from the skippedSchemas.

 Here is an example: 
image

packages/integration-sdk-entity-validator/src/validator.ts
geovannimp: I'm not sure the type is useful. It can make people force the type to use the functions, like:

```typescript
validator.validateEntity(myEntityObject as unknown as {_class: string; _type: string})
```

packages/integration-sdk-entity-validator/src/validator.ts
eyadmba: I was under the impression that when a class schema is updated, the type schemas that used it would also be updated (and published). But I also don't really know how that would work tbh.

packages/integration-sdk-entity-validator/src/validator.ts
```diff
@@ -0,0 +1,158 @@
...
+        skippedSchemas.push(
+          ...classArray.map(
+            (className) =>
+              ({
+                schemaId: `#${className}`,
+                reason: 'type-already-validated',
+                type: 'class',
+              }) satisfies SkippedSchema,
+          ),
+        );
```
eyadmba, line 146: What are we doing here?

geovannimp: As we validated the type, we are adding all class schemas as skipped.

packages/integration-sdk-entity-validator/src/validator.ts
geovannimp: Refactored. Everything is inside a single for loop right now making the code clear.

packages/integration-sdk-entity-validator/src/validator.ts
geovannimp: That's the idea, but I'm not sure when we are getting there. So I think having this flag available might be helpful for now. In the best scenario it won't be used and we can remove it.

packages/integration-sdk-entity-validator/src/validator.ts
eyadmba: It's beautiful!

packages/integration-sdk-entity-validator/src/validator.ts
```diff
@@ -0,0 +1,153 @@
...
+            ...(validator.errors?.map((error) =>
+              ajvErrorToEntityValidationError(schemaId, error),
+            ) ?? []),
+          );
+        }
+
+        if (type === 'type' && !forceClassValidationWithValidatedType) {
+          skippedSchemas.push(
+            // Skip each other schema as the type should already have class properties injected
+            ...schemasToValidate.slice(Number(i) + 1).map(
```
eyadmba, line 131: This type of thing scares me. It's easy to miss when making changes, and it's tricky to test reliably.

Is there a way we can ensure that the schemasToValidate only contain the actual schemas we want to run. Like if we move that `forceClassValidationWithValidatedType` check up to when we build the `schemasToValidate` list?

packages/integration-sdk-entity-validator/src/validator.ts
eyadmba: Doesn't that still indicate to the consumer that they're doing the wrong thing though? The type is doing its job still by guiding the consumer to not mess up unintentionally.

If they do want to over-rule the type error, they deserve all the evil things that will happen to them. In this life and the next.

eyadmba (approved):

packages/integration-sdk-entity-validator/src/j1Formats.ts
```diff
@@ -0,0 +1,22 @@
+import Ajv, { Format as AjvFormat } from 'ajv';
+import addFormats from 'ajv-formats';
+
+const ipv4 = addFormats.get('ipv4') as RegExp;
+const ipv6 = addFormats.get('ipv6') as RegExp;
+
+const ipv4CidrRegex =
+  /^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\/(3[0-2]|[12]?[0-9])$/;
+const ipv6CidrRegex =
+  /^([0-9a-fA-F]{1,4}:){7,7}[0-9a-fA-F]{1,4}\/(12[0-8]|1[01][0-9]|[1-9]?[0-9])$|^([0-9a-fA-F]{1,4}:){1,7}:\/(12[0-8]|1[01][0-9]|[1-9]?[0-9])$|^::\/(12[0-8]|1[01][0-9]|[1-9]?[0-9])$|^([0-9a-fA-F]{1,4}:){1,7}[0-9a-fA-F]{1,4}$/;
```
zemberdotnet, line 10: Were these not exported from the data model package?

geovannimp: Yep, but we'll be removing this (and ajv validation) from the data-model and using this package instead.

PR #1068, opened 2024-04-30
add support for using hasKey from graphObjectStore

zemberdotnet:
# Description
This PR adds support for using a `hasKey` implementation from the `graphObjectStore` rather than the `DuplicateKeyTracker` if it is available.

### DIFF
--- a/packages/integration-sdk-private-test-utils/src/graphObjectStore.ts
+++ b/packages/integration-sdk-private-test-utils/src/graphObjectStore.ts
@@ -66,3 +66,68 @@ export class InMemoryGraphObjectStore implements GraphObjectStore {
     return Promise.resolve();
   }
 }
+
+export class InMemoryGraphObjectStoreWithHasKeyImpl
+  implements GraphObjectStore
+{
+  private readonly entityMap = new Map<string, Entity>();
+  private readonly relationshipMap = new Map<string, Relationship>();
+
+  hasKey(_key: string | undefined) {
+    if (!_key) {
+      return false;
+    }
+    return this.entityMap.has(_key) || this.relationshipMap.has(_key);
+  }
+
+  async addEntities(stepId: string, newEntities: Entity[]): Promise<void> {
+    for (const entity of newEntities) {
+      this.entityMap.set(entity._key, entity);
+    }
+
+    return Promise.resolve();
+  }
+
+  async addRelationships(
+    stepId: string,
+    newRelationships: Relationship[],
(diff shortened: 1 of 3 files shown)


packages/integration-sdk-private-test-utils/src/graphObjectStore.ts
```diff
@@ -66,3 +66,68 @@ export class InMemoryGraphObjectStore implements GraphObjectStore {
     return Promise.resolve();
   }
 }
+
+export class InMemoryGraphObjectStoreWithHasKeyImpl
```
VDubber, line 70: Why not add `hasKey` to the existing ObjectStore above?

packages/integration-sdk-runtime/src/execution/jobState.ts
```diff
@@ -149,14 +150,36 @@ export function createStepJobState({
...
+
     for (const [index, entity] of entities.entries()) {
       try {
-        duplicateKeyTracker.registerKey(entity._key);
+        if (hasHasKeyFn) {
+          if (graphObjectStore.hasKey!(entity._key)) {
+            throw new IntegrationDuplicateKeyError(
+              `Duplicate _key detected (_key=${entity._key})`,
+            );
+            // if this add entities contains multiple entities we need to verify it doesn't contain any duplicates within
```
VDubber, line 169: The logic around this new set is messy. It introduces complexity where the duplicateKeyTracker was very straightforward. Can we move this complexity to a new implementation of the duplicateKeyTracker? 
It could first check the graphObjectStore and then have a temp store that would be exist while in this loop that takes the place of this new set.
It's not a blocker but seems like we could refine the approach?

packages/integration-sdk-runtime/src/execution/jobState.ts
```diff
@@ -275,7 +313,11 @@ export function createStepJobState({
 
     hasKey: (_key: string | undefined) => {
       if (!_key) return false;
-      return duplicateKeyTracker.hasKey(_key);
+      if (hasHasKeyFn) {
+        return graphObjectStore.hasKey!(_key);
+      } else {
+        return duplicateKeyTracker.hasKey(_key);
```
VDubber, line 319: This PR does not modify the _adding_ of a key to the `duplicateKeyTracker` only the use of the hasKey. Follow up PR to no longer add to the `duplicateKeyTracker` is coming?
I ask because we use the `duplicateKeyTracker` in the above findEntity method.

PR #1069, opened 2024-05-01
[APP-15119] - Add createIntegrationHelpers

geovannimp:


packages/integration-sdk-core/tools/generateTypeboxSchemas.ts
```diff
@@ -0,0 +1,46 @@
+import { TObject } from '@sinclair/typebox';
```
geovannimp, line 1: # Just ignore from here to the end. This must be moved to data-model repo.

jzolo22 (approved):

packages/integration-sdk-core/package.json
```diff
@@ -23,8 +23,9 @@
     "prepack": "yarn build:dist"
   },
   "dependencies": {
-    "@jupiterone/data-model": "^0.57.0",
+    "@jupiterone/data-model": "^0.61.3",
     "@jupiterone/integration-sdk-entity-validator": "^12.7.0",
+    "@sinclair/typebox": "^0.32.30",
```
eyadmba, line 28: Can typebox be a dev dependency?

tsconfig.base.json
```diff
@@ -6,6 +6,7 @@
     "lib": ["ES2022", "DOM"],
     "module": "commonjs",
     "moduleResolution": "node",
+    "resolveJsonModule": true,
```
eyadmba, line 9: What did we need this for?

eyadmba (dismissed):

packages/integration-sdk-core/package.json
geovannimp: No because we need it in runtime to generate the schema.

tsconfig.base.json
geovannimp: For some reason something changed in the data model and it cannot import the json schemas without this anymore :cry:

jzolo22 (dismissed):

jzolo22 (approved):

eyadmba (approved):
🔥

PR #1077, opened 2024-05-10
Bocchi updates

JakeFerrero:
Add documentation
Use new standard integrations' http client


packages/integration-sdk-cli/src/bocchi/templates/top-level/src/client.ts.hbs
```diff
@@ -1,53 +1,74 @@
-import { HttpServiceClient } from './httpServiceClient';
```
Nick-NCSU, line 1: Can we get rid of ./httpServiceClient?

packages/integration-sdk-cli/src/bocchi/templates/top-level/src/client.ts.hbs
```diff
@@ -1,53 +1,74 @@
...
-    })).{{response.dataPath}} as {{pascalCase entity.name}};
+      {
+        headers: { ...this.authHeaders },
+        method: '{{sanitizeHttpMethod request.method}}',
+      {{#if request.params}}
+        body: { {{#each (sanitizeHttpBody request.params)}}'{{@key}}': `{{this}}`,{{/each}} }
+      {{/if}}
+      }
+    );
+    return (await response.json() as any).{{response.dataPath}} as {{pascalCase entity.name}};
```
Nick-NCSU, line 71:
```suggestion
    import get from 'lodash/get';
    ...
    return get(await response.json() as any, response.dataPath);
```

packages/integration-sdk-cli/src/bocchi/templates/top-level/src/client.ts.hbs
```diff
@@ -1,53 +1,74 @@
...
-      {{/if}}
-    });
-    this.authHeaders = { {{#each (sanitizeAuthObject template.authentication.authHeaders)}}'{{@key}}': `{{this}}`,{{/each}} }
+    await this.getAuthorizationHeaders();
     {{/if}}
   }
 {{#each template.steps}}
 
   {{#if (isSingletonRequest response.responseType)}}
   public async get{{pascalCase entity.name}}({{#if parentAssociation}}parentEntity: Entity{{/if}}): Promise<any> {
```
Nick-NCSU, line 59:
```suggestion
  public async get{{pascalCase entity.name}}({{#if parentAssociation}}parentEntity: Entity{{/if}}): Promise<{{pascalCase entity.name}}> {
```

packages/integration-sdk-cli/src/bocchi/templates/top-level/src/client.ts.hbs
Nick-NCSU: Your docs seem to indicate that you can pass a `.` delimited path:

```json
{
    "response": {
        "data": [
            { "..." },
            { "..." },
            { "..." },
            "..."
        ]
    }
}
```

<i>Template:</i>

```json
{
  "dataPath": "response.data"
}
```

packages/integration-sdk-cli/src/bocchi/utils/types.ts
```diff
@@ -86,7 +92,7 @@ export interface Step {
   mappedRelationships?: {
     _class: RelationshipClass;
     direction: 'FORWARD' | 'REVERSE';
-    fieldMappings:
+    fieldMappings?:
```
Nick-NCSU, line 95: Is there a reason why a mapped relationship wouldn't be expected to have mappings?

PR #1087, opened 2024-06-10
integration helpers 

ryan-willis:


packages/integration-sdk-core/src/data/__tests__/createIntegrationHelpers.test.ts
```diff
@@ -117,30 +117,41 @@ describe('createIntegrationHelpers', () => {
...
-                maxItems: 1,
-                minItems: 1,
-              },
-              _type: { const: 'entity', type: 'string' },
-              someNewProperty: {
-                type: 'string',
+                  _type: { const: 'entity', type: 'string' },
+                },
+                type: 'object',
+                required: ['_class', '_type'],
```
jzolo22, line 138: If this is the only way we can do it, then we can just live with it and I can handle that in the docs but it is weird to have 2 separate `required` arrays 🤔

PR #1109, opened 2024-08-13
add-concurrency-to-iterate

VDubber:
Add concurrency parameter to the iterateEntities and iterateRelationship methods.

### DIFF
--- a/packages/integration-sdk-core/src/types/jobState.ts
+++ b/packages/integration-sdk-core/src/types/jobState.ts
@@ -5,6 +5,9 @@ export interface GraphObjectFilter {
 }
 
 export type GraphObjectIteratee<T> = (obj: T) => void | Promise<void>;
+export type GraphObjectIterateeOptions = {
+  concurrency?: number;
+};
 
 /**
  * The `JobState` is used to store and retrieve entities and relationships
@@ -103,7 +106,7 @@ export interface JobState {
    *
    * If concurrency is specified (defaults to 1), the iteratee will be executed
    * concurrently for graph objects that are found in memory and for
-   * graph objects found in single graph file.
+   * graph objects found in a given graph file. No specific ordering is guaranteed.
    * Consideration should be taken for provider rate limits when API requests
    * are being made within the iteratee and also while increasing concurrency
    * beyond the default.
@@ -120,7 +123,7 @@ export interface JobState {
   iterateEntities: <T extends Entity = Entity>(
     filter: GraphObjectFilter,
     iteratee: GraphObjectIteratee<T>,
-    concurrency?: number,
+    options?: GraphObjectIterateeOptions,
   ) => Promise<void>;
 
   /**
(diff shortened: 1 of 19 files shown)


packages/integration-sdk-core/src/types/jobState.ts
```diff
@@ -100,10 +100,27 @@ export interface JobState {
    * steps it depends on. Other steps outside the dependency ancestry may not
    * have run and therefore entities collected by those other steps should not
    * be expected to exist.
+   *
+   * If concurrency is specified (defaults to 1), the iteratee will be executed
+   * concurrently for graph objects that are found in memory and for
+   * graph objects found in single graph file.
```
Gonzalo-Avalos-Ribas, line 106: nit: ```suggestion
   * graph objects found in graph files.
```
We usually store in groups of 250 graphObjects per file, there is not a single file per entity type

File: packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts [hunk 2]
```diff
@@ -514,7 +514,7 @@ describe('iterateEntities', () => {
     const nonMatchingEntities = times(25, () =>
       createTestEntity({ _type: uuid() }),
     );
-    const matchingEntities = times(25, () =>
+    const matchingEntities = times(25000, () =>
```
Gonzalo-Avalos-Ribas, line 517: Is there a reason to have this be so big? 25K entities

packages/integration-sdk-core/src/types/jobState.ts
```diff
@@ -100,10 +100,27 @@ export interface JobState {
...
+   *   async (image) => {
+   *     ...
+   *   },
+   *   5,
+   * );
    */
   iterateEntities: <T extends Entity = Entity>(
     filter: GraphObjectFilter,
     iteratee: GraphObjectIteratee<T>,
+    concurrency?: number,
```
zemberdotnet, line 123: An options object would be more backwards compatible if we add addtl options in the future.

packages/integration-sdk-core/src/types/jobState.ts
VDubber: Thanks, this wasn't clear. We don't concurrently execute across multiple files at a time. Only one file at a time is what I meant.

packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/__tests__/FileSystemGraphObjectStore.test.ts
VDubber: Nope, I wanted to exercise the concurrency to see if there was any improvement. I can bring it back down

zemberdotnet (approved):

zemberdotnet (approved):

PR #1110, opened 2024-08-19
http-client: add body form urlencoded option

RonaldEAM:


packages/integration-sdk-http-client/src/client.ts
```diff
@@ -243,6 +243,17 @@ export abstract class BaseAPIClient {
         });
       } else if (bodyType === 'text' && typeof body === 'string') {
         fmtBody = body;
+      } else if (bodyType === 'urlencoded') {
+        fmtBody = new URLSearchParams();
+        Object.entries(body).forEach(([key, value]) => {
+          if (typeof value !== 'string') {
+            throw new IntegrationError({
```
VDubber, line 250: should we provide a softer interface that allows other types but we handle the conversion? Thoughts?

RonaldEAM: there's not too many options left, only binary, and I don't think we will ever upload files ever,  it's ok to keep it like this for now

packages/integration-sdk-http-client/src/client.ts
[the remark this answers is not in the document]
RonaldEAM: added!

packages/integration-sdk-http-client/src/__tests__/index.test.ts
```diff
@@ -179,6 +181,155 @@ describe('APIClient', () => {
         body: `{"test":"test"}`,
       });
     });
+
+    describe('fmtBody', () => {
```
VDubber, line 185: ❤️

VDubber (approved):

PR #1125, opened 2024-10-10
INT-11895: Add Retry logic

VDubber:
to finalizeSynchronization and abortSynchronization methods. Added tests. Added steps_finished Event Name


packages/integration-sdk-core/src/types/logger.ts
```diff
@@ -51,6 +51,7 @@ export enum IntegrationInfoEventName {
   Info = 'info',
   Stats = 'stats',
   Results = 'results',
+  StepsFinished = 'steps_finished',
```
VDubber, line 54: bonus add

dobregons (dismissed):

dobregons (approved):

PR #1126, opened 2024-10-14
Release 13.7.2

gastonyelmini:
Added argent option to the http client in order to be able to customize that in the integrations itself


packages/integration-sdk-http-client/src/client.ts
```diff
@@ -52,6 +53,7 @@ export abstract class BaseAPIClient {
   protected tokenBucketInitialConfig?: TokenBucketOptions | undefined;
   protected endpointTokenBuckets: Record<string, HierarchicalTokenBucket> = {};
   protected refreshAuth?: RefreshAuthOptions;
+  protected agent?: Agent;
```
VDubber, line 56: We should make these `readonly` where possible.. Follow up PR?

VDubber (approved):

PR #1129, opened 2024-10-23
Update createIntegrationHelpers.md

kyorkJ1:


packages/integration-sdk-core/docs/createIntegrationHelpers.md
```diff
@@ -111,7 +111,7 @@ export const { createEntityType, createEntityMetadata } =
 ```
 
 3. Replace INTEGRATION_NAME with the name of your integration
-4. Create a file named entities.ts in the src folder of your integration and add
+4. Create a file named entity-definition.ts in the src folder of your integration and add
```
jzolo22, line 114: Just wondering why the proposed change? I think the consensus was that as long as it's consistent the naming isn't that important and at this point I know most if not all of the ones we converted to use this have used the `entities.ts` convention

PR #1140, opened 2024-12-04
add object diff to log with duplicate entity report

RonaldEAM:


packages/integration-sdk-runtime/src/execution/duplicateKeyTracker.ts
```diff
@@ -132,8 +132,98 @@ export type DuplicateEntityReport = {
...
+    } else if (valDest === undefined) {
+      addDiff(fullPath, 'missing_in_dest');
+    } else if (typeof valSrc !== typeof valDest) {
+      addDiff(fullPath, 'type_mismatch', {
+        src: typeof valSrc,
+        dest: typeof valDest,
+      });
+    } else if (Array.isArray(valSrc) && Array.isArray(valDest)) {
+      if (JSON.stringify(valSrc) !== JSON.stringify(valDest)) {
+        addDiff(fullPath, 'value_mismatch');
```
VDubber, line 213:
```suggestion
        addDiff(fullPath, 'array_values_mismatch');
```

packages/integration-sdk-runtime/src/execution/duplicateKeyTracker.ts
```diff
@@ -132,8 +132,98 @@ export type DuplicateEntityReport = {
...
+      }
+    } else if (isObject(valSrc) && isObject(valDest)) {
+      // Recursive comparison for nested objects
+      const nestedDiff = diffObjects(valSrc, valDest, fullPath);
```
VDubber, line 217: How can we prevent infinite recursion if a property in an object references itself?

packages/integration-sdk-runtime/src/execution/duplicateKeyTracker.ts
```diff
@@ -144,12 +234,38 @@ export type DuplicateEntityReport = {
...
+      // ignore
+    }
+  }
+
+  let propertiesDiff: ObjectDiff | undefined;
+  if (!propertiesMatch) {
+    try {
+      propertiesDiff = diffObjects(aClone, bClone);
+    } catch (e) {
+      // ignore
```
VDubber, line 260: It'd be nice to get a log that the diffObject operation failed.

packages/integration-sdk-runtime/src/execution/duplicateKeyTracker.ts
```diff
@@ -144,12 +234,38 @@ export type DuplicateEntityReport = {
...
+    }
+  }
 
   return {
     _key: a._key,
-    rawDataMatch: isDeepStrictEqual(a._rawData, b._rawData),
-    propertiesMatch: isDeepStrictEqual(aClone, bClone),
+    rawDataMatch,
+    propertiesMatch,
```
VDubber, line 267: if easy, can we rename `propertiesMatch` to `entityPropertiesMatch`?

packages/integration-sdk-runtime/src/execution/duplicateKeyTracker.ts
```diff
@@ -132,8 +132,98 @@ export type DuplicateEntityReport = {
...
+ * console.log(result);
+ * // Output:
+ * // {
+ * //   "b.c": { type: "value_mismatch" },
+ * //   "d": { type: "missing_in_src" }
+ * // }
+ */
+export function diffObjects(
+  src: unknown,
+  dest: unknown,
```
VDubber, line 174: nit: Source and destination are a little odd for variable names in this context.
rename options: firstEncounteredObject, secondEncounteredObject?
or a, b
or object1, object2

packages/integration-sdk-runtime/src/execution/duplicateKeyTracker.ts
```diff
@@ -132,8 +132,98 @@ export type DuplicateEntityReport = {
...
 
+type DiffType =
+  | 'missing_in_src'
+  | 'missing_in_dest'
+  | 'type_mismatch'
+  | 'value_mismatch';
+
+interface ObjectDiff {
+  [key: string]: {
+    type: DiffType;
```
VDubber, line 147:
```suggestion
    diffType: DiffType;
```

packages/integration-sdk-runtime/src/execution/duplicateKeyTracker.ts
RonaldEAM: Hmm, I don't think that's possible, in raw data certainly not because it's parsed from a string (the response) so each nested object is a unique object. And in the entity properties we shouldn't have that issue either because we only accept string, number and boolean (and array of those).
We could prevent it by keeping track of every object we encounter but that adds a layer of complexity because we'll be using more memory

packages/integration-sdk-runtime/src/execution/duplicateKeyTracker.ts
RonaldEAM: This is hard to do, as we don't pass a logger to the job state code, I could add a property for the error message to the return object of this function, this ends up in a log so it could work, but it's weird. Let me know what you think!

VDubber (approved):

PR #1152, opened 2025-02-11
Release 15.2.3

jzolo22:
Add deprecation console warning to data-model properties and bump data-model package


VDubber:
Can we please do a major version bump?

PR #1155, opened 2025-02-24
INT-12497 - add support for system-internal

Sivli-Embir:


packages/integration-sdk-runtime/src/synchronization/index.ts
```diff
@@ -136,9 +140,23 @@ function buildJobConfiguration({
...
+    return {
+      source,
+      scope,
+    };
+  }
+
+  if (source === 'system-internal') {
+    return {
+      source,
+      scope,
```
Sivli-Embir, line 153: We need to pass scope through for this source

VDubber: Currently the SDK is public facing. We don't want external users incorrectly setting the source. 🤔

Sivli-Embir: They would have to intentionally set their source as `system-internal` which would be a case of self sabotage. I agree that ideally they would not be able to use this but I am not sure we have an alternative as internal integrations use the same SDK.

VDubber (approved):

PR #1160, opened 2025-03-05
http-client: create ResponseBodyError when the body is empty

RonaldEAM:


packages/integration-sdk-http-client/src/errors.ts
```diff
@@ -79,7 +79,10 @@ export async function retryableRequestError({
 
   if (response.status === 429) {
     return new RateLimitError({
-      cause: errorBody ? new ResponseBodyError(errorBody, headers) : undefined,
+      cause:
+        errorBody !== undefined
```
Gonzalo-Avalos-Ribas, line 83: null !== undefined is true, is that correct?

Gonzalo-Avalos-Ribas (approved):

PR #1178, opened 2025-09-10
Adds proxy support using the Alpha client

nickjupiter1:
For https://jupiterone.atlassian.net/browse/INT-12936

### DIFF
--- a/packages/integration-sdk-runtime/src/api/index.ts
+++ b/packages/integration-sdk-runtime/src/api/index.ts
@@ -1,4 +1,5 @@
 import { Alpha, AlphaInterceptor, AlphaOptions } from '@lifeomic/alpha';
+import { AxiosProxyConfig } from 'axios';
 import { IntegrationError } from '@jupiterone/integration-sdk-core';
 import dotenv from 'dotenv';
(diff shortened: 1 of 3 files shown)


packages/integration-sdk-runtime/src/api/index.ts
```diff
@@ -161,3 +169,32 @@ export const getApiKeyFromEnvironment = () =>
...
+      proxy.auth = {
+        username: decodeURIComponent(url.username),
+        password: decodeURIComponent(url.password),
+      };
+    }
+
+    return proxy;
+  } catch (error) {
+    const parsedError = error instanceof TypeError ? error : new TypeError(String(error));
+    console.warn('Failed to parse proxy URL:', proxyUrl, parsedError);
```
Gonzalo-Avalos-Ribas, line 192: this code only executes if proxyUrlString is populated, so if there is a proxy config, and we cannot parse it, its not safe to continue without the proxy

nickjupiter1: Agreed! I modified it to not catch the error.

Gonzalo-Avalos-Ribas (approved):

PR #1180, opened 2025-09-24
[TD-7352] - Add better logging for flushing and uploading failures

dobregons:


packages/integration-sdk-runtime/src/storage/FileSystemGraphObjectStore/FileSystemGraphObjectStore.ts
```diff
@@ -325,36 +333,83 @@ export class FileSystemGraphObjectStore implements GraphObjectStore {
...
+          }
         }
 
         this.localGraphObjectStore.flushEntities(entities, stepId);
         entitiesToUpload = entitiesToUpload.concat(entities);
       }
 
-      if (onEntitiesFlushed) {
-        await onEntitiesFlushed(entitiesToUpload);
+      if (onEntitiesFlushed && entitiesToUpload.length > 0) {
```
zemberdotnet, line 400: Why are we changing this condition?

Gonzalo-Avalos-Ribas (approved):

PR #1200, opened 2026-04-09
feat(cli): aggregate step authorization into top-level config field 

gastonyelmini:
Adds a top-level authorization field to the generated ingestionSourcesConfig.json that aggregates all authorization requirements from every integration step — regardless of whether the step has an ingestionSourceId or not.

Previously, authorization metadata was only available inside childIngestionSources, which excluded root steps (steps matched directly by ingestionSourceId) and steps without an ingestionSourceId (like fetch-account). This meant downstream consumers like docs-v2 could never generate complete authorization documentation.

The new field collects permissions, roles, oauthScopes, apis, endpoints, licenses, and documentationLinks from all steps, deduplicates them, and sorts them alphabetically. It is only present when at least one step declares authorization.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -4,5 +4,5 @@
     "packages/integration-sdk-*",
     "packages/cli"
   ],
-  "version": "17.3.0"
+  "version": "17.4.0"
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -20700,11 +20700,11 @@
     },
     "packages/cli": {
       "name": "@jupiterone/cli",
-      "version": "17.3.0",
+      "version": "17.4.0",
       "license": "MPL-2.0",
       "dependencies": {
-        "@jupiterone/integration-sdk-core": "^17.3.0",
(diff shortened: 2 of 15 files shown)


tom-foley-j1 (approved):

VDubber (approved):
## Review Summary

Clean, well-tested feature addition. The code correctly deduplicates, sorts, and conditionally omits the `authorization` field. Version bumps are consistent across all packages. No security, data volume, or performance concerns.

| Dimension | Score (1-5) | Rationale |
|-----------|-------------|-----------|
| **Security risk** | 1 | No auth, credential, or data exposure changes. CLI-time metadata only. |
| **Data volume impact** | 1 | No effect on runtime data pipelines. Adds a small field to a static config JSON. |
| **Performance risk** | 1 | Build-time CLI command processing a handful of steps. No runtime impact. |

**Verdict: Approve with suggestions** — see inline comments.

